PR #12, opened 2021-04-27
remove reaper sidecar integration

jsanda:
This replaces https://github.com/datastax/cass-operator/pull/335


burmanm:
You need to run ``mage operator:generateClient``  and perhaps ``mage operator:sdkGenerate`` ..

PR #17, opened 2021-04-30
Add support for getting configuration from a secret

### LINKED ISSUE #13: Be able to source CONFIG_FILE_DATA from a secret (datastax/cass-operator#413)
Issue to track datastax/cass-operator#413

jsanda:
Fixes #13

This PR adds support for specifying Cassandra configuration with a secret. Here is an example to illustrate the usage. 

cassdc.yaml:

```yaml
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: test
  configSecret: cassdc-config
  size: 1
  storageConfig:
    cassandraDataVolumeClaimSpec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
      storageClassName: standard
```

config-secret.yaml:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: cassdc-config
type: Opaque
stringData:
  config: |-
    {
      "cassandra-yaml": {
        "roles_update_interval_in_ms": 65000,
        "read_request_timeout_in_ms": 5000
      },
      "jvm-options": {
        "initial_heap_size": "512M",
        "max_heap_size": "512M"
      }
    }
```

Here is a summary of how things work:

* If the secret does not exist reconciliation fails
* Fetch the secret specified by `configSecret`
* Add an annotation with the name of the datacenter to the secret
    * This annotation is used to set up a watch
* Get the config from the secret and apply additional properties 
    * Note that this is existing functionality but needs to be pointed out
* Get or create the _datacenter config secret_
    * This secret is created by cass-operator. It contains the combined config.
* Check to see if the config has changed
* If the config has changed
    * Add annotation with a hash of the datacenter config secret to the datacenter object
    * Patch the datacenter
    * Persist the datacenter config secret
* When the statefulset is constructed check to see if `configSecret` is set. It takes precedence over the inline config
    * Add the `CONFIG_FILE_DATA env var to the server config init container and the value comes from the datacenter config secret
    * Add a `CONFIG_HASH` env that contains the config has
        * This is to facilitate statefulset updates when the config secret is updated

### DIFF
--- a/.github/workflows/kindIntegTest.yml
+++ b/.github/workflows/kindIntegTest.yml
@@ -88,6 +88,8 @@ jobs:
         - additional_seeds
         - cluster_wide_install
         - config_change
+        - internode_encryption
+        - config_secret
         - multi_cluster_management
         - oss_test_all_the_things
         - scale_down
--- a/.github/workflows/kindIntegTest.yml
+++ b/.github/workflows/kindIntegTest.yml
@@ -88,7 +88,6 @@ jobs:
         - additional_seeds
         - cluster_wide_install
         - config_change
-        - internode_encryption
         - config_secret
         - multi_cluster_management
         - oss_test_all_the_things
--- a/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -34,6 +34,12 @@ const (
 	// PromMetricsLabel is a service label that can be selected for prometheus metrics scraping
 	PromMetricsLabel = "cassandra.datastax.com/prom-metrics"
 
+	// DatacenterAnnotation is the operator's annotation for the datacenter name
+	DatacenterAnnotation = DatacenterLabel
+
+	// ConfigHashAnnotation is the operator's annotation for the hash of the ConfigSecret
+	ConfigHashAnnotation = "cassandra.datastax.com/config-hash"
+
 	// CassNodeState
 	CassNodeState = "cassandra.datastax.com/node-state"
 
@@ -84,6 +90,25 @@ type CassandraDatacenterSpec struct {
 	// Config for the server, in YAML format
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Config json.RawMessage `json:"config,omitempty"`
+
+	// ConfigSecret is the name of a secret that contains configuration for Cassandra. The
+	// secret is expected to have a property named config whose value should be a JSON
+	// formatted string that should look like this:
+	//
+	//    config: |-
+	//      {
+	//        "cassandra-yaml": {
+	//          "read_request_timeout_in_ms": 10000
+	//        },
+	//        "jmv-options": {
+	//          "max_heap_size": 1024M
+    //        }
+	//      }
+	//
+	// ConfigSecret is mutually exclusive with Config. ConfigSecret takes precedence and
+	// will be used exclusively if both properties are set. The operator sets a watch such
+	// that an update to the secret will trigger an update of the StatefulSets.
+	ConfigSecret string `json:"configSecret,omitempty"`
 
 	// Config for the Management API certificates
 	ManagementApiAuth ManagementApiAuthConfig `json:"managementApiAuth,omitempty"`
@@ -537,7 +562,7 @@ func (dc *CassandraDatacenter) GetSuperuserSecretNamespacedName() types.Namespac
 }
 
 // GetConfigAsJSON gets a JSON-encoded string suitable for passing to configBuilder
-func (dc *CassandraDatacenter) GetConfigAsJSON() (string, error) {
+func (dc *CassandraDatacenter) GetConfigAsJSON(config []byte) (string, error) {
 
 	// We use the cluster seed-service name here for the seed list as it will
 	// resolve to the seed nodes. This obviates the need to update the
@@ -600,8 +625,8 @@ func (dc *CassandraDatacenter) GetConfigAsJSON() (string, error) {
 		return "", errors.Wrap(err, "Model information for CassandraDatacenter resource was not properly configured")
 	}
 
-	if dc.Spec.Config != nil {
-		configParsed, err := gabs.ParseJSON([]byte(dc.Spec.Config))
+	if config != nil {
+		configParsed, err := gabs.ParseJSON(config)
 		if err != nil {
 			return "", errors.Wrap(err, "Error parsing Spec.Config for CassandraDatacenter resource")
 		}
--- a/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
+++ b/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
@@ -120,6 +120,52 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	configSecretMapFn := handler.ToRequestsFunc(func(mapObj handler.MapObject) []reconcile.Request {
+		log.Info("config secret watch called", "Secret", mapObj.Meta.GetName())
+
+		requests := make([]reconcile.Request, 0)
+		secret := mapObj.Object.(*corev1.Secret)
+		if v, ok := secret.Annotations[api.DatacenterAnnotation]; ok {
+			log.Info("adding reconciliation request for config secret", "Secret", secret.Name)
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: secret.Namespace,
+					Name: v,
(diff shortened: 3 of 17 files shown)


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -85,6 +85,8 @@ type CassandraDatacenterSpec struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Config json.RawMessage `json:"config,omitempty"`
 
+	ConfigSecret string `json:"configSecret,omitempty"`
```
burmanm, line 111: Missing description

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -120,6 +120,56 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	dcAnnotation := "cassandra.datastax.com/datacenter"
```
burmanm: This is used as a label (named DatacenterLabel from the cassdctypes) often, should probably use that one and explain why it's an annotation here instead of label.

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -120,6 +120,56 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	
+	configSecretMapFn := handler.ToRequestsFunc(func(mapObj handler.MapObject) []reconcile.Request {
+		log.Info("config secret watch called", "Secret", mapObj.Meta.GetName())
+
+		requests := make([]reconcile.Request, 0)
+		secret := mapObj.Object.(*corev1.Secret)
+		v, ok := secret.Annotations[dcAnnotation]
```
burmanm: Merge with the next if

operator/pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -325,6 +329,45 @@ func buildInitContainers(dc *api.CassandraDatacenter, rackName string, baseTempl
...
+
+		if configHash, ok := dc.Annotations[configHashAnnotation]; ok {
+			envVars = append(envVars, corev1.EnvVar{
+				Name: "CONFIG_HASH",
+				Value: configHash,
+			})
+		} else {
+			return nil, fmt.Errorf("datacenter %s is missing %s annotation", dc.Name, configHashAnnotation)
+		}
+		return envVars, nil
```
burmanm: Could be inside the if clause and that else removed.

operator/pkg/reconciliation/reconcile_configsecret.go
```diff
@@ -0,0 +1,158 @@
...
+
+		rc.ReqLogger.Info("updating datacenter config secret", "ConfigSecret", dcConfigSecret.Name)
+		dcConfigSecret.Data["config"] = config
+
+		if exists {
+			if err := rc.Client.Update(rc.Ctx, dcConfigSecret); err != nil {
+				rc.ReqLogger.Error(err,"failed to update datacenter config secret", "ConfigSecret", dcConfigSecret.Name)
+				return result.Error(err)
+			}
+		} else {
```
burmanm: Unnecessary else

operator/pkg/reconciliation/reconcile_configsecret.go
```diff
@@ -0,0 +1,158 @@
...
+// checkDatacenterNameAnnotation Checks to see if the secret has the
+// cassandra.datastax.com/datacenter annotation with the datacenter name as its value. If
+// the secret does not have the annotation, it is added, and the secret is patched. The
+// secret should be the one specifiied by ConfigSecret.
+func (rc *ReconciliationContext) checkDatacenterNameAnnotation(secret *corev1.Secret) error {
+	annotation := "cassandra.datastax.com/datacenter"
+
+	if v, ok := secret.Annotations[annotation]; ok &&  v == rc.Datacenter.Name {
+		return nil
+	} else {
```
burmanm: .. remove this one also.

operator/pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -325,6 +329,45 @@ func buildInitContainers(dc *api.CassandraDatacenter, rackName string, baseTempl
...
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: getDatacenterConfigSecretName(dc),
+					},
+					Key: "config",
+				},
+			},
+		})
+
+		configHashAnnotation := "cassandra.datastax.com/config-hash"
```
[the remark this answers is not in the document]
jsanda: What do you mean?

.github/workflows/kindIntegTest.yml
```diff
@@ -88,6 +88,8 @@ jobs:
         - additional_seeds
         - cluster_wide_install
         - config_change
+        - internode_encryption
```
burmanm: Remove, lets not re-enable it ;)

jsanda: Just my way of saying, are you sure.... 😆

operator/pkg/reconciliation/reconcile_configsecret.go
```diff
@@ -0,0 +1,159 @@
...
+	patch := client.MergeFrom(secret.DeepCopy())
+	secret.Annotations[api.DatacenterAnnotation] = rc.Datacenter.Name
+	return rc.Client.Patch(rc.Ctx, secret, patch)
+}
+
+// updateConfigHashAnnotation Adds the config hash annotation to the datacenter. The value
+// of the annotation is a hash of the secret. The datacenter is then patched. Note that the
+// secret should be the datacenter secret created by the operator, NOT the secret specified
+// by ConfigSecret.
+func (rc *ReconciliationContext) updateConfigHashAnnotation(secret *corev1.Secret) error {
```
burmanm, line 97: How do we ensure that the copy secret has the same security as the original secret? If the data is sensitive, we don't want this copy to be viewable anymore than the real secret. Or if the secret came from Hashicorp Vault or some other injection method (I don't know how the customer handles them if they're really sensitive).

tests/testdata/default-three-rack-three-node-dc.yaml
```diff
@@ -17,18 +17,11 @@ spec:
         resources:
           requests:
             storage: 1Gi
-  resources:
```
burmanm, line 20: Why are these removed?

operator/pkg/reconciliation/reconcile_configsecret.go
jsanda: >How do we ensure that the copy secret has the same security as the original secret? If the data is sensitive, we don't want this copy to be viewable anymore than the real secret. 

Great question. I think this falls more on the administrator or whoever is setting up RBAC for users and service accounts. The operator already has permissions to read/write secrets in the namespace in which it is deployed. I think handling other RBAC falls outside the scope of the operator.

tests/testdata/default-three-rack-three-node-dc.yaml
jsanda: I thought some of the initial failures may have been to do the limited resources. The max heap was basically the same as the memory limit specified under `resources`, leaving very little space for off-heap memory.

burmanm:
Just last questions:

If the secret is created after the Cassdc, what happens? I guess it will not be picked up, since it's not owned by the Cassdc - but do we reconcile all the secrets if that annotation was applied to it?

And if the secret is deleted - what happens?

(it's fine if these are documentation only issues - but just wanted to know the intended behavior)

jsanda:
>If the secret is created after the Cassdc, what happens?

I assume you are referring to the secret specified by the `ConfigSecret` property. Reconciliation fails. StatefulSets will not be created. 

>do we reconcile all the secrets if that annotation was applied to it?

Changes to any secret having the `cassandra.datastax.com/datacenter` annotation will trigger reconciliation on the CassandraDatacenter. The CassandraDatacenter is determined by the value of the annotation here:

```go
if v, ok := secret.Annotations[api.DatacenterAnnotation]; ok {
			log.Info("adding reconciliation request for config secret", "Secret", secret.Name)
			requests = append(requests, reconcile.Request{
				NamespacedName: types.NamespacedName{
					Namespace: secret.Namespace,
					Name: v,
				},
			})
		}
```

>And if the secret is deleted - what happens?

Same as above - reconciliation fails because `ConfigSecret` points to a secret that does not exist. Let's say you unset `ConfigSecret`. Then we fall back to using `Config` or defaults.

PR #18, opened 2021-05-03
Allow dns lookup by pod name for all pods

rchernobelskiy:
Some background for this change:
The purpose of the `ServiceName` field of a StatefulSet is to specify the name of the service under which DNS subdomain records are created for each pod: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id

The operator makes two services for dse pods, one with `PublishNotReadyAddresses` set to `true` and one with it set to `false`.

Setting the `ServiceName` to the service which has `PublishNotReadyAddresses` set to `true` allows dns lookups for pods that are not ready yet.

This can be useful in a number of scenarios, one of which is having the pods on an overlay network with stable IPs, where pod to pod communication happens via dns names: https://www.datastax.com/blog/2021/05/how-connect-stateful-workloads-across-kubernetes-clusters

### DIFF
--- a/operator/pkg/reconciliation/construct_statefulset.go
+++ b/operator/pkg/reconciliation/construct_statefulset.go
@@ -198,7 +198,7 @@ func newStatefulSetForCassandraDatacenterHelper(
 				MatchLabels: statefulSetSelectorLabels,
 			},
 			Replicas:             &replicaCountInt32,
-			ServiceName:          dc.GetDatacenterServiceName(),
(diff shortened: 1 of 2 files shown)


burmanm:
This type of change requires regenerating the client / sdk. So ``mage operator:sdkGenerate`` and ``mage operator:generateClient``.

rchernobelskiy:
@burmanm are you sure? Those commands produced no changes when ran

jsanda (approved):

PR #20, opened 2021-05-03
Always include `max_direct_memory` in examples

jimdickinson:
Given that the scripts for starting Cassandra do some calculations on the amount of memory that aren't container aware, I think it's important to have this in every example file.

### DIFF
--- a/operator/example-cassdc-yaml/dse-6.8.x/example-cassdc-minimal.yaml
+++ b/operator/example-cassdc-yaml/dse-6.8.x/example-cassdc-minimal.yaml
@@ -23,6 +23,7 @@ spec:
     jvm-server-options:
       initial_heap_size: "800M"
       max_heap_size: "800M"
+      max_direct_memory: "800M"
       additional-jvm-opts:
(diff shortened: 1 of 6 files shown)


jsanda (approved):

operator/example-cassdc-yaml/dse-6.8.x/example-cassdc-minimal.yaml
```diff
@@ -23,6 +23,7 @@ spec:
     jvm-server-options:
       initial_heap_size: "800M"
       max_heap_size: "800M"
+      max_direct_memory: "800M"
```
burmanm, line 26: What's the point here? The default from modern JDK should be the same?

jimdickinson: no - this is about the cassandra startup scripts calculating a poor value inside the container

burmanm: Is there a reason we wouldn't fix such calculations in cass-operator?

PR #33, opened 2021-05-10
When new docker image is created, push an update PR to k8ssandra

burmanm:
**What this PR does**:
Whenever the master branch is updated and new docker images are created, this sends a PR to the k8ssandra repo that updates the CRD and the values.yaml image path. If the PR hasn't been merged, it will be updated.

**Checklist**

* [ ] Changes manually tested
* [ ] Documentation added/updated

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-539) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/operatorBuildAndDeploy.yml
+++ b/.github/workflows/operatorBuildAndDeploy.yml
@@ -127,3 +127,42 @@ jobs:
           platforms: linux/amd64
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
+  push_k8ssandra_update:
+    runs-on: ubuntu-latest
+    needs: [build_system_logger, build_operator_docker]
+    name: Update dependency in k8ssandra
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: Set git parsed values
+        id: vars
+        run: |
+          echo ::set-output name=sha_short::$(git rev-parse --short=8 ${{ github.sha }})
+          echo ::set-output name=tag_name::${GITHUB_REF#refs/tags/}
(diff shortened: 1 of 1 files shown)


.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -127,3 +127,42 @@ jobs:
...
+          echo ::set-output name=tag_name::${GITHUB_REF#refs/tags/}
+      - name: Checkout k8ssandra charts
+        uses: actions/checkout@v2
+        with:
+          repository: k8ssandra/k8ssandra
+          path: k8ssandra
+      - name: Update cass-operator values in k8ssandra
+        run: |
+          # Update the CRD
+          cp charts/cass-operator-chart/templates/customresourcedefinition.yaml k8ssandra/charts/cass-operator/crds/customresourcedefinition.yaml
```
jsanda, line 151: Does this need to be changed to follow the steps described at https://github.com/k8ssandra/k8ssandra/blob/main/charts/cass-operator/UPGRADE.md?

.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -127,3 +127,42 @@ jobs:
           platforms: linux/amd64
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
+  push_k8ssandra_update:
```
jsanda, line 130: We are at a point now with the projects where we will be releasing new versions of component charts more frequently than the k8ssandra chart. Given that would it make more sense for this to run when a tag is pushed vs every push to master?

adejanovski:
Although it's not working flawlessly, we have the GHA dependency check workflow in K8ssandra which does exactly this.
Should we close this PR then?

PR #120, opened 2021-06-10
Update operator-sdk to 1.x from the old <0.19 structure (new kubebuilder struct)

burmanm:
**What this PR does**:
Updates the base of cass-operator to use operator-sdk 1.8 instead of the old pre-0.18.x line. This means new kubebuilder v3 layout as well as other controller-runtime changes.

Other changes include removal of Helm charts, reduction of mage use (intended to remove) and replaced with Makefile. 

TODO Still:

- [x] Re-enable webhooks
- [x] Re-enable dynamicSecrets (or replace with better approach)
- [x] Metrics (+prom) verification and Cassandra ServiceMonitor
- [x] ClusterRole / Role generation
- [x] Move operator/pkg to pkg/ 
- [x] All the reconciliation logic (except ReconciliationContext) to controllers/
- [x] Migrate crd.patch (x-kubernetes-keys removals) to Kustomize

Next PRs, not in this one to help review process:

- [ ] Rewrite the upgrade test to update from last datastax cass-operator (1.6.0) to this
- [ ] Scripts that allow to run kustomize scripts to create docs examples (and generate samples)
- [ ] Remove rest of mage, migrate to Makefile if there's still something useful

Future work:

- [ ] Add envtest usage for tests, replace some integ tests with envtest
- [ ] Merge integ-tests where necessary
- [ ] Rewrite fakeclient tests with envtest instead
- [ ] Fix bundle creation
- [ ] UBI-image support
- [ ] openapi-gen / client-gen if deemed necessary still (kubebuilder no longer supports them). controller-runtime clients do not need this feature.
- [ ] Deprecated PSP features and their future (remove or replace with more common solution not just VMware specific?)

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


jsanda:
@burmanm I realize that there is no way to avoid this being a massive PR, but it would be great if you are able to pull out anything into separate, smaller PRs. It will make reviewing a lot easier. Please let me know when you want me to start reviewing.

burmanm:
I'm not sure if separating to different PRs is easy (even if we accept that they won't compile / run tests until all of them are merged).

jsanda:
> I'm not sure if separating to different PRs is easy (even if we accept that they won't compile / run tests until all of them are merged).

Wishful thinking on my part :)

config/default/kustomization.yaml
```diff
@@ -0,0 +1,74 @@
...
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: cass-operator-
+
+# Labels to add to all resources and selectors.
+#commonLabels:
+#  someName: someValue
+
+bases:
```
jsanda, line 15: I believe `bases` is deprecated and should be replaced with `resources`.

Do we want to include certmanager by default?

config/manager/manager.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
```
jsanda, line 5: I am not a fan of the generated value `controller-manager` that is used all throughout the manifests. I prefer to change it to the name of the operator. In this case I would replace `controller-manager` with `cass-operator`. Thoughts?

jsanda (changes requested):

.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -38,21 +39,24 @@ jobs:
...
-        uses: magefile/mage-action@v1
-        with:
-          version: latest
-          args: operator:testSdkGenerate
-      - name: Test Client Generate
-        uses: magefile/mage-action@v1
-        with:
-          version: latest
-          args: operator:testGenerateClient
+      # - name: Test Sdk Generate
```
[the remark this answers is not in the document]
burmanm: Removed already in the follow-up PR.

.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -38,21 +39,24 @@ jobs:
...
+      #     args: operator:testGenerateClient
       - name: Unit Tests
-        uses: magefile/mage-action@v1
-        with:
-          version: latest
-          args: operator:testGo
+        run: |
+          export PATH=$GOROOT/bin:$GOPATH/bin:$PATH
+          make controller-gen test
+        # uses: magefile/mage-action@v1
```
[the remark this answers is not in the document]
burmanm: Removed already in the follow-up PR.

config/default/kustomization.yaml
burmanm: bases is the naming used by the newest version of operator-sdk

config/manager/manager.yaml
burmanm: We need to create a script then that modifies them after every time controller-gen is ran (manifest / generate), since the kubebuilder will always use that name (no configuration option).

jsanda: hmm... I tried the following:

* Add field to CassandraDatacenterSpec
* Modify RBAC annotation in cassandradatacenter_controller.go
* In manager.yaml replace controller-manager with with cass-operator
* Run `make manifests`

My changes in deployment.yaml stayed intact.

config/crd/patches/config_removal.yaml
```diff
@@ -0,0 +1,15 @@
...
+#
+# This is necessary because the config field has a dynamic
+# schema which depends on the DSE version selected, and
+# dynamic schema aren't possible to fully specify and
+# validate via openAPI V3.
+#
+# Instead, we remove the config field from the schema
+# entirely and instruct openAPI/k8s to preserve fields even
+# if they aren't specified in the CRD. The field itself is defined
+# as a json.RawMessage, see dsedatacenter_types.go in the
```
jsanda, line 12: Change `see dsedatacenter_types.go` to `see cassandradatacenter_types.go`

config/default/kustomization.yaml
jsanda: If it is deprecated though wouldn't it be better to update?

controllers/cassandradatacenter_controller.go
```diff
@@ -0,0 +1,288 @@
...
+			return oplabels.HasManagedByCassandraOperatorLabel(e.Object.GetLabels())
+		},
+	}
+
+	// NOTE: We do not currently watch PVC resources, but if we did, we'd have to
+	// account for the fact that they might use the old managed-by label value
+	// (oplabels.ManagedByLabelDefunctValue) for CassandraDatacenters originally
+	// created in version 1.1.0 or earlier.
+
+	c = c.Watches(
```
jsanda: Can we replace each of the `Watches` with the `Owns` method of `Builder` where we are using a `handler.EnqueueRequestForOwner`?

magefile.go
```diff
@@ -10,14 +10,8 @@ import (
```
jsanda, line 10: Can you create a separate ticket for removing mage? Unless there is a small amount of work left, I am favor of tackling it in a follow up PR.

jsanda (changes requested):
I think we will be able to merge pretty soon. So many of the changes are structural. One of the things I want to do as part of my next review is see how well I am able to consume the kustomize resources from another project, namely k8ssandra-operator.

magefile.go
burmanm: It's already removed in the follow-up PR #127

controllers/cassandradatacenter_controller.go
burmanm: I had it that way, but now I can't remember why I had to rewrite them back to watches..

burmanm: Okay, so I think I did the change to make all the other types consistent. But I can replace some of those Watches (not all) with Owns.

config/default/kustomization.yaml
burmanm: It will break users who use kustomize in kubectl <1.20. It will change behavior to those users.

config/manager/manager.yaml
burmanm: And your deployment has a funny name afterwards also ;)

```
NAME                                          READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/cass-operator-cass-operator   0/1     1            0           114s

NAME                                             READY   STATUS             RESTARTS   AGE
pod/cass-operator-cass-operator-b687684f-wc8bn   1/2     CrashLoopBackOff   2          114s

NAME                                                   DESIRED   CURRENT   READY   AGE
replicaset.apps/cass-operator-cass-operator-b687684f   1         1         0       114s
```

The name-prefix in the kustomize template hides most of the controller-manager instances. The label isn't modified, but all the names are. We might want to look at what to do with the control-plane, it has some effect on AKS at least..

config/default/kustomization.yaml
jsanda: Ah, I always had problems with the version of kustomize kubectl so I installed it separately.

For reference https://github.com/kubernetes/kubectl/issues/891#issuecomment-731238572.

config/manager/manager.yaml
jsanda: My comments were based off of what I have done elsewhere. I failed to mention that I had removed the prefix. If there are changes we want to make around the kustomize stuff, they can probably be done in follow up PRs. The tests are passing which is the biggest thing.

jsanda (approved):
Great job with this!

PR #148, opened 2021-07-18
Multigroup structure with configuration CRDs everywhere, move to UBI images

### LINKED ISSUE #145: K8SSAND-719 ⁃ ImageMap should be loaded from Object
**What is missing?**
We should have a ConfigMap or a CRD that would have the updated mapping between version of Cassandra and the image that is used. Such as 3.11.10 -> k8ssandra/cass-management-api:3.11.10-v0.1.26. 

Updating this mapping should be as easy as updating the object, such as with kubectl apply -f <path-to-repo>. And it should be usable from the Helm charts in k8ssandra also. 

**Why do we need it?**
At the moment we need to update the cass-operator source code to modify the mapping between version and management-API image.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-719) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-719
┆priority: Medium

### LINKED ISSUE #146: K8SSAND-729 ⁃ Create UBI images of cass-operator and system-logger
**What is missing?**
We don't create UBI images of cass-operator itself or system-logger. Needs modifications to Dockerfiles. We should create them If those images are not a lot bigger than our current ones, we should use them rather than keep creating dual images.

**Why do we need it?**
UBI images get better security rating and are also required for Red Hat's processes.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-729) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-729
┆priority: Medium

burmanm:
**What this PR does**:
Modifies the project to use kubebuilder multigroup structure to allow more types to be added later (such as configs). Also, modifies the operator and system-logger-image to use UBI8 as their base.

All the image properties are configurable from a CRD that's deployed with a ConfigMap, removing all the ENV parameters previously used (such as registry overrides, pull secrets etc). SKIP_VALIDATING_WEBHOOK was moved to a OperatorConfig also instead of ENV parameter.

Fixes #146
Fixes #145

**Checklist**

* [ ] Changes manually tested
* [ ] Documentation added/updated

┆Issue is synchronized with this [Jiraserver Task](https://k8ssandra.atlassian.net/browse/K8SSAND-770) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-770
┆Priority: Medium


apis/config/v1beta1/imageconfig_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+// ImageConfig is the Schema for the imageconfigs API
+type ImageConfig struct {
+	metav1.TypeMeta `json:",inline"`
+
+	Images *Images `json:"images,omitempty"`
+
+	ImageRegistry string `json:"imageRegistryOverride,omitempty"`
+
+	ImagePullSecret corev1.LocalObjectReference `json:"imagePullSecret,omitempty"`
```
jsanda, line 38: Should this be an array?

apis/config/v1beta1/imageconfig_types.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:images
+
+// ImageConfig is the Schema for the imageconfigs API
+type ImageConfig struct {
```
jsanda, line 31: I haven't finished looking through all the changes but let me ask some questions.

* Is there a 1-to-1 mapping between a CassandraDatacenter and ImageConfig? If so how is that enforced?
* If I change one of the images in my ImageConfig will that result in the underlying StatefulSet being updated?
* What is the default behavior if there is no ImageConfig?
* What happens if I delete the ImageConfig?

apis/config/v1beta1/imageconfig_types.go
burmanm: No? What would multiple secrets target?

apis/config/v1beta1/imageconfig_types.go
burmanm: No, ImageConfig is per cass-operator (it's a operator configuration). If you modify it and restart cass-operator, the behavior is the same as having currently newer version of operator (which would have different versions). This is just a static configuration at this point, no reconcile or updates while cass-operator is running.

If you delete ImageConfig, operator will refuse to start.

burmanm: At the moment the kustomize template generates the configuration, both operatorConfig as well as imageConfig. Missing either one would fail the operator creation.

So as long as the user doesn't intentionally override that creation then defaults are created.

jsanda (approved):

PR #155, opened 2021-07-26
Support for building OLM specific bundles and indexes

bradfordcp:
_What this PR does_*: Adds support for installation of cass-operator with OLM

**Which issue(s) this PR fixes**: N/A

**Checklist**

* [x] Changes manually tested
* [x] Documentation added/updated

**Note:** this PR is currently a draft while I run through a suite of manual testing.

┆Issue is synchronized with this [Jiraserver Task](https://k8ssandra.atlassian.net/browse/K8SSAND-768) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-768
┆Priority: Medium

### DIFF
--- a/charts/cass-operator-chart/Chart.yaml
+++ b/charts/cass-operator-chart/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: cass-operator
-version: 1.7.0
+version: 1.7.1
 description: Helm chart for Cass Operator.
-appVersion: 1.7.0
+appVersion: 1.7.1
 home: https://github.com/k8ssandra/cass-operator
 maintainers:
   - name: Cassandra Operator Team
--- a/charts/cass-operator-chart/values.yaml
+++ b/charts/cass-operator-chart/values.yaml
@@ -9,6 +9,6 @@ webhookClusterRoleName: cass-operator-webhook
 webhookClusterRoleBindingName: cass-operator-webhook
 deploymentName: cass-operator
 deploymentReplicas: 1
-defaultImage: "k8ssandra/cass-operator:v1.7.0"
+defaultImage: "k8ssandra/cass-operator:v1.7.1"
 imagePullPolicy: IfNotPresent
 imagePullSecret: ""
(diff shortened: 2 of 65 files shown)


bradfordcp:
@burmanm looking at the conflicts above is it safe to assume I can remove these files from my branch as they have been removed or relocated in `master`?

burmanm:
I'm afraid you'll have to recreate a new branch and start from scratch (and copy required files one by one). It's probably easier than trying to rebase this one to the new structure.

A lot of files you've pushed here and no longer there or no longer used in this sense. A pretty simple process you could start with is take the new master code, run the necessary make commands:

``make IMG=k8ssandra/cass-operator:$DEV_TAG_NAME VERSION=$VERSION bundle bundle-build catalog-build``

(there's also bundle-push and catalog-push, but those are ran at the release time automatically - unless you want to push them to your local repo, but modify the IMG tag in that case, the bundle and catalog image names are derived from it)

And before that obviously create the correct cass-operator image with the ``make docker-build``. Those should create the bundles, catalogs and such (with CRDs, deployment, CSVs etc). Then modifying certain Kustomize rules and PROJECT file etc should allow adding the necessary adjustments you need. The package-manifest structure is deprecated which you've used previously.

The operator-sdk's OLM integration document is a good starting point:

https://sdk.operatorframework.io/docs/olm-integration/

bradfordcp:
I thought I had moved to the bundle format (given all the Dockerfiles for storing CRDs, annotations, and CSVs). I am interested in comparing the CSVs created by operator-sdk to the ones we have hand-rolled here.

Regarding making the cass-operator docker containers will that be an issue given they are using previous versions of operator-sdk? Shouldn't I be able to use the containers already built for older versions of cass-operator? (This is just related to the cass-operator container, not bundle)

burmanm:
You mean using older 1.7.1 with the newer bundle / catalog creation? Should work just fine, the CRD API version and so on change, but there shouldn't be changes that prevent stuff from working.

burmanm:
If you do make bundle in the master, you can see in the bundle directory what it created for that bundle. It fetches samples from config/samples and somewhere the docs (I haven't checked where - maybe config/manifests). Mostly, everything is created from under config/ which means we update those parts (same kustomize is used everywhere).

PR #160, opened 2021-08-16
Update to operator-sdk 1.11.0 (controller-runtime 0.9.2)

burmanm:
**What this PR does**:
Updates the operator-sdk to 0.11.0 to match what was updated in k8ssandra-operator.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


go.mod
```diff
@@ -6,48 +6,48 @@ require (
...
 	github.com/pkg/errors v0.9.1
-	github.com/stretchr/testify v1.6.1
+	github.com/stretchr/testify v1.7.0
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.20.2
-	k8s.io/apimachinery v0.20.2
-	k8s.io/client-go v12.0.0+incompatible
-	k8s.io/code-generator v0.20.2
+	k8s.io/api v0.21.3
```
adutra: Out of curiosity how did you come up with these version numbers? Here for example I see that v0.21.4 could be used (published on August 12th).
And in fact, any version specified here will end up replaced with v0.21.2 because of the replace directive below. So why not use v0.21.2?

adutra (approved):
LGTM. I manually tested the new Makefile and it works 100%.

Out of curiosity: the PR says "Update to operator-sdk 0.11.0" but I guess it's 1.11.0 right?

go.mod
burmanm: Yeah, we can use 0.21.2 directly or update to newer one. Those versions come from whatever was created by the operator-sdk in it's go.mod example.

PR #163, opened 2021-08-24
If TargetRef is set for additional seed IP, do not remove it

burmanm:
**What this PR does**:
If additional-seds service has a TargetRef set for a IP address, the modifications to AdditionalSeeds will not remove those addresses, but retain them (considering them as managed from the outside).  Also, always add the additional seeds to cass-config-builder since we always create the service.

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -1803,7 +1803,13 @@ func (rc *ReconciliationContext) startOneNodePerRack(endpointData httphelper.Cas
 
 	// if the DC has no ready seeds, label a pod as a seed before we start Cassandra on it
 	// and also consider additional seeds
-	labelSeedBeforeStart := readySeeds == 0 && len(rc.Datacenter.Spec.AdditionalSeeds) == 0
+
+	existingEndpoints, err := rc.GetAdditionalSeedEndpoint()
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -1803,7 +1803,13 @@ func (rc *ReconciliationContext) startOneNodePerRack(endpointData httphelper.Cas
...
 	// if the DC has no ready seeds, label a pod as a seed before we start Cassandra on it
 	// and also consider additional seeds
-	labelSeedBeforeStart := readySeeds == 0 && len(rc.Datacenter.Spec.AdditionalSeeds) == 0
+
+	externalSeedPoints := 0
+	if existingEndpoints, err := rc.GetAdditionalSeedEndpoint(); err == nil {
+		externalSeedPoints = len(existingEndpoints.Subsets[0].Addresses)
+	}
+
+	labelSeedBeforeStart := readySeeds == 0 && len(rc.Datacenter.Spec.AdditionalSeeds) == 0 && externalSeedPoints == 0
```
jsanda, line 1812: Are we sure that we want to check external seeds here? I believe this is where the first seed (or seeds) is labeled and started. Subsequent seed labeling is done in `checkSeedLabels()`. If have an external seed here and have not yet started any C* nodes in this dc, I fear that this might prevent any nodes from starting.

jsanda (approved):

PR #173, opened 2021-09-16
Add support for mgmt api ListKeyspaces op

adejanovski:
**What this PR does**:
Adds support for the newly added list keyspaces operation in the management API: https://github.com/k8ssandra/management-api-for-apache-cassandra/commit/e77aa6ab32fb119fbb865cf8edc3f8a9a37eab95

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pkg/httphelper/client.go
+++ b/pkg/httphelper/client.go
@@ -288,8 +288,8 @@ func parseListKeyspacesEndpointsResponseBody(body []byte) ([]string, error) {
 	return keyspaces, nil
 }
 
-// ListKeyspaces calls the management API to list existing keyspaces
-func (client *NodeMgmtClient) ListKeyspaces(pod *corev1.Pod, keyspaceName string) ([]string, error) {
+// GetKeyspace calls the management API to check if a specific keyspace exists
+func (client *NodeMgmtClient) GetKeyspace(pod *corev1.Pod, keyspaceName string) ([]string, error) {
 	podHost, err := BuildPodHostFromPod(pod)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/httphelper/client.go
```diff
@@ -280,6 +280,37 @@ func (client *NodeMgmtClient) modifyKeyspace(endpoint string, pod *corev1.Pod, k
...
+func parseListKeyspacesEndpointsResponseBody(body []byte) ([]string, error) {
+	var keyspaces []string
+	if err := json.Unmarshal(body, &keyspaces); err != nil {
+		return nil, err
+	}
+	return keyspaces, nil
+}
+
+// ListKeyspaces calls the management API to list existing keyspaces
+func (client *NodeMgmtClient) ListKeyspaces(pod *corev1.Pod, keyspaceName string) ([]string, error) {
```
burmanm: Can there be multiple keyspaces with the same name? ;) Or why is "ListKeyspaces" requiring keyspaceName to list?

adejanovski: It was a way of filtering on the management API side if you're looking for a single keyspace and don't want to filter on the client side.
Dutra seemed to like the idea of doing this (or maybe it was Erik, can't remember), and I'm fine with it 🙂

burmanm: In that case, shouldn't it be called "GetKeyspace" ? Or how does one List those keyspaces?

adejanovski: Leaving it empty will list all keyspaces: https://github.com/k8ssandra/management-api-for-apache-cassandra/commit/e77aa6ab32fb119fbb865cf8edc3f8a9a37eab95#diff-51532ffcd7720e83141217c410cb08e76cf95fd93e2ad354abd72b71c78aa8d4R174

burmanm: Indeed, that request in the Test removes the parameter:

```java
MockHttpRequest request = MockHttpRequest.get(ROOT_PATH + "/ops/keyspace");
```

This implementation always adds the parameter keyspaceName (which is called filter in those Java tests). Thus, the implementation here should add the filter only ``if keyspaceName != ""`` 

```go
request := nodeMgmtRequest{
		endpoint: fmt.Sprintf("/api/v0/ops/keyspace?keyspaceName=%s", keyspaceName),
```

adejanovski: But it'll work either way as passing an empty value for `keyspaceName` will give the same result as not passing it.
Would you rather have two distinct functions here, one with the `keyspaceName` arg and the other without?

burmanm: You can't do that in Golang without naming them. What guarantees would the reader have that keyspaceName empty is same as not setting it? If even the management-api has no guarantees or test for that?

burmanm (approved):

Issue #175, opened 2021-09-19
K8SSAND-917 ⁃ Add full query logging to reconciliation loop


Miles-Garnsey:
We have a number of PRs + issues open across 3 repos to add full query logging capabilities to k8ssandra, since it was added in Cassandra 4.0.0.

Management API - https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/131
Cass config builder (definition submodule) - https://github.com/datastax/cass-config-definitions/pull/65
k8ssandra helm charts - https://github.com/k8ssandra/k8ssandra/issues/1102

I think we may also need to add a call to the management API  from cass-operator. This will ensure that the system state is brought into the target  state described in the `CassandraDatacenter` (and propagated to Cassandra via `cassandra.yaml`) during the reconciliation loop.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-917) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-917
┆priority: Medium

jsanda:
Why would we need to call the management-api? If query logging is enabled that will trigger a statefulset update, i.e., a rolling restart which means the change will take effect.

PR #176, opened 2021-09-19
Add logic to set full query logging enabled/disabled in reconciliation loop [K8SSAND-917]

Miles-Garnsey:
Draft PR to start work on getting full query logging to be enabled/disabled automatically via the mgmt API depending on whether it is enabled in the `cassandra.yaml`.


Miles-Garnsey:
This PR now ready for review. I have:
* Tested on Cassandra 4.0.0 with FQL turned on and off. It behaves as expected in both cases.
* Tested on Cassandra 3.0.0 with FQL turned on. This leads to Cassandra not starting and the Reconciler never calling `SetFullQueryLogging()`.

Miles-Garnsey:
We have failing tests here, but the failures may be spurious.

I got the following results when comparing a few of the integration tests across local runs on master, this branch, and GHA runs on this branch:

* `additional_serviceoptions`: Passes locally on master. Passes locally on my branch. Fails on my branch in GHA.
* `smoke_test_oss`: Passes on master locally. Passes on my branch locally (sometimes). Fails on my branch in GHA.
* `scale_up`: initially failed on my branch locally before I realised that it wants more kind nodes available. Rescaled kind cluster, got a cert-manager error. Ran it again, now it passes (I assume because cert-manager wasn't installed the first time and wasn't deleted after the first run.) It still fails as I only scaled to 4 kind nodes and it seems it wants more. Since the cluster does scale I'm calling this a pass.

It seems to me that there are intermittent failures in the integration tests which relate to (1) timeouts, and (2) whether or not cert-manager is installed on the cluster. 

I've just completed another additional round of manual testing on 3 4.0.0 nodes with FQL turned on, as my initial testing only covered single node clusters. The changes continue to behave as I'd expect, with an FQL log created on each of the 3 nodes.

I think this is ready for merge and will close #175 once it has been reviewed.

pkg/httphelper/client.go
```diff
@@ -456,3 +457,50 @@ func callNodeMgmtEndpoint(client *NodeMgmtClient, request nodeMgmtRequest, conte
...
+	if err != nil {
+		return false, err
+	}
+	request := nodeMgmtRequest{
+		endpoint: "/api/v0/ops/node/fullquerylogging",
+		host:     podHost,
+		method:   http.MethodGet,
+		timeout:  time.Minute * 2,
+	}
+	isEnabledBytes, err := callNodeMgmtEndpoint(client, request, "")
```
burmanm: In Go, opt for short variable names. This "isEnabled" indicates it's a Java boolean, but that's not true at all. It could be "payload" for example to indicate what was fetched.

pkg/httphelper/client.go
```diff
@@ -456,3 +457,50 @@ func callNodeMgmtEndpoint(client *NodeMgmtClient, request nodeMgmtRequest, conte
...
+	if err != nil {
+		client.Log.Error(err, "failed to call endpoint /api/v0/ops/node/fullquerylogging")
+		return false, err
+	}
+	var isEnabledMap map[string]interface{}
```
burmanm: Add whitespace between operations for easier read. And change variable name.

pkg/httphelper/client.go
```diff
@@ -456,3 +457,50 @@ func callNodeMgmtEndpoint(client *NodeMgmtClient, request nodeMgmtRequest, conte
...
+	err = json.Unmarshal([]byte(isEnabledBytes), &isEnabledMap)
+	if err != nil {
+		client.Log.Error(err, "failed to unmarshall JSON response from /api/v0/ops/node/fullquerylogging", "response", string(isEnabledBytes))
+		return false, err
+	}
+	isEnabledBool := isEnabledMap["entity"]
```
burmanm: This and the last line are unsafe assumptions. When fetching a key, you should check if that key actually exists in the payload. And when doing a cast, you need to that the cast is possible - otherwise you will panic the runtime and the cass-operator will go crashing down.

pkg/httphelper/client.go
```diff
@@ -456,3 +457,50 @@ func callNodeMgmtEndpoint(client *NodeMgmtClient, request nodeMgmtRequest, conte
...
+	if isEnabledBool == nil {
+		err := errors.New("failed to access entity key in response from /api/v0/ops/node/fullquerylogging ")
+		return false, err
+	}
+	return isEnabledBool.(bool), err
+}
+
+func (client *NodeMgmtClient) CallSetFullQuerylog(pod *corev1.Pod, enableFullQueryLogging bool) error {
```
burmanm: log and Log in different functions.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
 	return false, nil
 }
 
+func (rc *ReconciliationContext) SetFullQueryLogging() result.ReconcileResult {
```
burmanm: Is the visibility supposed to be public?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+	datacenter := rc.GetDatacenter()
+	var DCConfig map[string]interface{}
```
burmanm: Wrong visibility for DCConfig, start with lowerCase.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
 	return false, nil
 }
 
+func (rc *ReconciliationContext) SetFullQueryLogging() result.ReconcileResult {
+	datacenter := rc.GetDatacenter()
```
burmanm: This does not shorten much. Why not just .. dc?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+	}
+	shouldFQLBeEnabled := false
+	if err := DCConfig["full_query_logging_options"]; err == nil {
+		serverMajorVersion, err := strconv.ParseInt(strings.Split(datacenter.Spec.ServerVersion, ".")[0], 10, 8)
+		if err != nil {
+			rc.ReqLogger.Error(err, "error parsing server major version. Can't enable full query logging without knowing this")
+			return result.RequeueSoon(2)
+		}
+		if serverMajorVersion < 4 {
+			err := goerrors.New("full query logging only supported on OSS Cassandra 4x+")
```
burmanm: Since fmt is already imported, one can use fmt.Errorf() here to avoid additional import with rename.

pkg/httphelper/client.go
Miles-Garnsey: Good idea, will implement as suggested. The response does indeed contain some additional fields about the MIME media type and stuff, so it really isn't a bool at all.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+	if err := json.Unmarshal(datacenter.Spec.Config, &DCConfig); err != nil {
+		rc.ReqLogger.Error(err, "error unmarshalling DC config JSON")
+		return result.RequeueSoon(2)
+	}
+	shouldFQLBeEnabled := false
+	if err := DCConfig["full_query_logging_options"]; err == nil {
+		serverMajorVersion, err := strconv.ParseInt(strings.Split(datacenter.Spec.ServerVersion, ".")[0], 10, 8)
+		if err != nil {
+			rc.ReqLogger.Error(err, "error parsing server major version. Can't enable full query logging without knowing this")
+			return result.RequeueSoon(2)
```
burmanm: If we can't parse the version, will that magically work the next time? Or will we endlessly loop an error which is not transient?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+		}
+		if serverMajorVersion < 4 {
+			err := goerrors.New("full query logging only supported on OSS Cassandra 4x+")
+			rc.ReqLogger.Error(err, "full_query_logging_options is defined in Cassandra config, it is not supported on the version of Cassandra you are running")
+			return result.RequeueSoon(2)
+		}
+		rc.ReqLogger.Info("full_query_logging_options is defined in Cassandra config, we will try to enable it via the management API")
+		shouldFQLBeEnabled = true
```
burmanm: Do we need this parameter even? Can it ever be false and we would continue?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
 
+func (rc *ReconciliationContext) SetFullQueryLogging() result.ReconcileResult {
+	datacenter := rc.GetDatacenter()
+	var DCConfig map[string]interface{}
+	if err := json.Unmarshal(datacenter.Spec.Config, &DCConfig); err != nil {
+		rc.ReqLogger.Error(err, "error unmarshalling DC config JSON")
+		return result.RequeueSoon(2)
+	}
+	shouldFQLBeEnabled := false
+	if err := DCConfig["full_query_logging_options"]; err == nil {
```
burmanm: What if err is not nil?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+		shouldFQLBeEnabled = true
+	}
+
+	podList, err := rc.listPods(rc.Datacenter.GetClusterLabels())
+	if err != nil {
+		rc.ReqLogger.Error(err, "error listing all pods in the cluster to progress full query logging reconciliation")
+		return result.RequeueSoon(2)
+	}
+	for _, podPtr := range PodPtrsFromPodList(podList) {
+		fqlEnabledForPod, err := rc.NodeMgmtClient.CallIsFullQueryLogEnabledEndpoint(podPtr)
```
burmanm, line 2278: We should not make calls to the management-api if we're not going to use those results. If "shouldFqlBeEnabled" is disabled, we should not do this.

pkg/reconciliation/reconcile_racks.go
burmanm: Also, instead of checking for error, the correct use for map is:

```go
if fqlValue, found := dcConfig["full_query_logging_options"]; found { }
```

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+	if err := DCConfig["full_query_logging_options"]; err == nil {
+		serverMajorVersion, err := strconv.ParseInt(strings.Split(datacenter.Spec.ServerVersion, ".")[0], 10, 8)
+		if err != nil {
+			rc.ReqLogger.Error(err, "error parsing server major version. Can't enable full query logging without knowing this")
+			return result.RequeueSoon(2)
+		}
+		if serverMajorVersion < 4 {
+			err := goerrors.New("full query logging only supported on OSS Cassandra 4x+")
+			rc.ReqLogger.Error(err, "full_query_logging_options is defined in Cassandra config, it is not supported on the version of Cassandra you are running")
+			return result.RequeueSoon(2)
```
burmanm: Will the serverVersion be updated to 4.x on the next reconcile?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+func (rc *ReconciliationContext) SetFullQueryLogging() result.ReconcileResult {
+	datacenter := rc.GetDatacenter()
+	var DCConfig map[string]interface{}
+	if err := json.Unmarshal(datacenter.Spec.Config, &DCConfig); err != nil {
+		rc.ReqLogger.Error(err, "error unmarshalling DC config JSON")
+		return result.RequeueSoon(2)
+	}
+	shouldFQLBeEnabled := false
+	if err := DCConfig["full_query_logging_options"]; err == nil {
+		serverMajorVersion, err := strconv.ParseInt(strings.Split(datacenter.Spec.ServerVersion, ".")[0], 10, 8)
```
burmanm: And a simple note.. serverVersion isn't necessarily correct (even if it should be)  - so careful here. User can set serverVersion to 3.11.11 and serverImage to 4.0. Or the other way around.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2246,52 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+		return result.RequeueSoon(2)
+	}
+	for _, podPtr := range PodPtrsFromPodList(podList) {
+		fqlEnabledForPod, err := rc.NodeMgmtClient.CallIsFullQueryLogEnabledEndpoint(podPtr)
+		if err != nil {
+			rc.ReqLogger.Error(err, "can't get whether query logging enabled for pod ", "podName", podPtr.Name)
+			return result.RequeueSoon(2)
+		}
+		if fqlEnabledForPod != shouldFQLBeEnabled {
+			rc.ReqLogger.Info("enabling full query logging on ", "podIP", podPtr.Status.HostIP, "podName", podPtr.Name)
```
burmanm: .. or disabling?

burmanm (changes requested):
See comments.

pkg/httphelper/client.go
Miles-Garnsey: Oh jeez. So I couldn't figure this out because for some horrible reason a cast like this: `isEnabledBool.(bool)` returns `(bool, bool)` not `bool, err` like every other function call in Go. 

Same goes for when a map indexing operation can't find a key it appears. Do note that when indexing the map I actually was doing this check:

```
	if isEnabledBool == nil {
		err := errors.New("failed to access entity key in response from /api/v0/ops/node/fullquerylogging")
		return false, err
	}
```

I'm pretty sure that would avoid a panic but I'm curious if you think otherwise.

Either way, still a bit more practice with Go required for me it appears... I'll use `if !ok {` style checks on both to avoid the issues you've raised.

pkg/httphelper/client.go
Miles-Garnsey: Good catch, I'm happy to fix.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: I'm open to your thoughts on this. The other functions called within `ReconcileAllRacks()` are public, so I took my cue from there.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: Would visibility not be confined to the function given that this is a variable locally instantiated there? Or would this cause it to leak out into the scope of `rc` because this function has it as a receiver?

Either way I think your suggestion is more idiomatic and will implement.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: Can do.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: Happy to do this. I didn't love having to rename the import, so thanks for the suggestion.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: The truth is that we don't know. It might. We aren't in control of the `datacenter` instance right, so it makes sense to retry as long as there's a possibility of later success IMO.

We certainly don't want to silently continue, or return `result.done()`, and it doesn't make sense to panic. Is there a fifth option you would suggest?

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: Yes, if FQL is currently enabled, but has been removed from the `config.cassandra-yaml` YAML field, then FQL should be turned off at that point.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: So if `err != nil` then `shouldFQLBeEnabled = true` is never run which makes `shouldFQLBeEnabled` always false. It will end up with `rc.NodeMgmtClient.CallSetFullQuerylog(podPtr, false)` being run, which will disable FQL as intended.

> if fqlValue, found := dcConfig["full_query_logging_options"]; found { }

You're right, my original code actually makes very limited sense, this variable definitely should not have been called `err`! Thanks for picking this up.

If you have suggestions to add further clarity on the conditional logic I'm open to them, I don't particularly like the way this is written currently but it seemed to be the way to do it in Go.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: It could be, if the server version is changed?

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: That's a good point. In this instance we have protection at other layers of the system (especially, Cassandra actually won't start adn this code will never run if you try to add the `full_query_logging_options` into cassandra.yaml on 3x).

This is just here for completeness.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: Good point. I'll fix this.

Miles-Garnsey:
@burmanm I've made the requested changes, and some additional ones as well - 

* Some integration tests may have been failing due to the changes in Ginkgo that you mentioned. It seems it was not previously looking in subdirectories because the `./...` argument was not being picked up correctly. 
* I found that when the config key was undefined we saw errors as I wasn't catering to the case where it was nil. This should not have impacted integration tests (just causes more retries which may cause timeouts to be hit).
* I don't actually know why some of the logic around checking whether the `full_query_logging_options` key was present was working, the logic was flaky. Fixed now.
* Additional safety around all casts and map indexing as suggested.

I think this is ready for a second round of review now. There are three specific questions I have as well - 

* I'm not sure whether `result.RequeueSoon` is the right choice for my return values throughout the error paths. Would you prefer I returned `result.Error`, and would the `SetFullQueryLogging` function still keep getting called in the reconciler in that case? 
* I've chosen to call `result.Continue` instead of `result.Done` throughout, because I wasn't sure if `Done` stops all further reconciliation functions - let me know if you prefer `Done` instead of `Continue`.
* I vaguely recall that changes to racks don't always propagate (there's some forceRackUpgrade field somewhere in the cassDC resource, I think?) Do you foresee any problems with `ReconcileAllRacks` not being called on a change to the `full_query_logging_options?

**EDIT: integration tests are still failing. This could be because of -** 

* `smoke_test_oss`: I can get this to run locally, but only the second time through. The first time, the application of the cert-manager resources fails because of `failed calling webhook "webhook.cert-manager.io":` cert-manager webhook is not ready when the test tries to create certificates.
* `config_change_condition`: passes on my local machine. Probably failing here due to a timeout.

I'm not going to go through the remainder. I did try to run all tests locally with `M_INTEG_DIR=all` but I saw no output from this and when I cancelled the run after 15 minutes it had not progressed beyond the first test.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2245,65 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+		if !found {
+			return result.Continue()
+		}
+		casYamlMap, ok := casYaml.(map[string]interface{})
+		if !ok {
+			err := fmt.Errorf("type casting error")
+			rc.ReqLogger.Error(err, "couldn't cast cassandra-yaml value from config to map[string]interface{}")
+			return result.RequeueSoon(2)
+		}
+		if _, found := casYamlMap["full_query_logging_options"]; found {
```
jsanda: `SetFullQueryLogging` is called after all C* pods should be ready. It would be nice to pull the validation part out into a separate function that is called earlier, preferably before creating/updating StatefulSets.

It would nice to have some unit tests as well.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2245,65 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+	}
+	for _, podPtr := range PodPtrsFromPodList(podList) {
+		fqlEnabledForPod, err := rc.NodeMgmtClient.CallIsFullQueryLogEnabledEndpoint(podPtr)
+		if err != nil {
+			rc.ReqLogger.Error(err, "can't get whether query logging enabled for pod ", "podName", podPtr.Name)
+			return result.RequeueSoon(2)
+		}
+		rc.ReqLogger.Info("full query logging status:", "isEnabled", fqlEnabledForPod, "shouldBeEnabled", shouldFQLBeEnabled)
+		if fqlEnabledForPod != shouldFQLBeEnabled {
+			rc.ReqLogger.Info("Setting full query logging on ", "podIP", podPtr.Status.HostIP, "podName", podPtr.Name, "fqlDesiredState", shouldFQLBeEnabled)
```
jsanda: Should `podPtr.Status.HostIP` be changed to `podPtr.Status.PodIP`?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2245,65 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+			err := rc.NodeMgmtClient.CallSetFullQueryLog(podPtr, shouldFQLBeEnabled)
+			if err != nil {
+				rc.ReqLogger.Error(err, "couldn't enable full query logging on ", "podIP", podPtr.Status.HostIP, "podName", podPtr.Name)
```
jsanda: Should `podPtr.Status.HostIP` be changed to `podPtr.Status.PodIP`?

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: Sorry @jsanda I just saw this review. 

Should we not wait for all pods to be ready before calling either of the management API endpoints? Both depend on a live Cassandra cluster being in U/N as nodetool can't be called beforehand.

RE unit tests, I don't know whether this is unit-testable because all of the meaningful work is done via side effects. If you want me to look into some sort of envtest thing or Kuttl then let me know.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: Is hostIP actually the IP of the underlying k8s node? I assumed it was the podIP but clearly I'm wrong.

That does make me wonder why the CR even knows about the underlying node IPs?

pkg/reconciliation/reconcile_racks.go
jsanda: I would separate out the validation into a separate function that is called earlier. The validation check should be done before any StatefulSets are created or updated.

The code base overall does not lend itself to testing. The validation portion though could be refactored into a function instead of a method of ReconciliationContext. It would be a lot easier to add unit tests for that function. For the rest of `SetFullQueryLogging` I think a full e2e test makes sense so that we can validate that the changes are made on the running node.

pkg/reconciliation/reconcile_racks.go
jsanda: Not sure I understand the question. Are you asking why the CassandraDatacenter knows about the node IPs? If so, it doesn't really. Also keep in mind that pods allow you to use host networking and that is configurable through the CassandraDatacenter.

Miles-Garnsey: Apologies John, I've just re-read the code properly (this is several weeks old now - blame my goldfish memory) and I can see that actually the HostIP is a property on the `corev1.Pod` type, not one of our own CRs. I think k8s tracks that as part of pod statuses to ensure that pods have been assigned to hosts - so nothing to do with us. 

This should, as you say log information about the PodIP not the host IP - I will fix, thanks for catching it!

pkg/reconciliation/reconcile_racks.go
burmanm: It won't. The serverVersion will stay intact during the requeue. If the user changes the version, it will cause a queueing of the object. But a requeue with the same data is not going to fix this issue.

pkg/reconciliation/reconcile_racks.go
burmanm: Validation could probably be done in the webhook, as that's where the input validation should happen (that's the only way to prevent pushing broken data to api-server).

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2245,65 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+		if !ok {
+			err := fmt.Errorf("type casting error")
+			rc.ReqLogger.Error(err, "couldn't cast cassandra-yaml value from config to map[string]interface{}")
+			return result.RequeueSoon(2)
+		}
+		if _, found := casYamlMap["full_query_logging_options"]; found {
+			serverMajorVersion, err := strconv.ParseInt(strings.Split(dc.Spec.ServerVersion, ".")[0], 10, 8)
+			if err != nil {
+				rc.ReqLogger.Error(err, "error parsing server major version. Can't enable full query logging without knowing this")
+				return result.RequeueSoon(2)
```
burmanm: Is this recoverable by trying again? If not, no requeue with the same data.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2243,6 +2245,65 @@ func (rc *ReconciliationContext) fixMissingPVC() (bool, error) {
...
+			}
+			if serverMajorVersion < 4 {
+				err := fmt.Errorf("full query logging only supported on OSS Cassandra 4x+")
+				rc.ReqLogger.Error(err, "full_query_logging_options is defined in Cassandra config, it is not supported on the version of Cassandra you are running")
+				return result.RequeueSoon(2)
```
burmanm: Same as above, we can't recover with the same data.

pkg/reconciliation/reconcile_racks.go
burmanm: For testing, this type of "side-effect" testing can be done by creating an integration test. Create the Datacenter, enable fullQueryLogging, make some CQL statements and verify the FQL logs were created (perhaps even with some real knowledge of data). Then disable and verify it was disabled..

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: So `result.Error()` here makes more sense. Roger.

pkg/reconciliation/reconcile_racks.go
jsanda: I agree that it is better to call `result.Error()` here. It is worth noting though that the CassandraDatacenter still gets requeued when there is an error. The controller requeues it with a rate limiter with an exponential backoff.

pkg/reconciliation/reconcile_racks.go
jsanda: >Validation could probably be done in the webhook, as that's where the input validation should happen (that's the only way to prevent pushing broken data to api-server)

Agreed, but we do also perform validation in the controller (see [here](https://github.com/k8ssandra/cass-operator/blob/master/controllers/cassandra/cassandradatacenter_controller.go#L116)) which I think is good.

Miles-Garnsey:
@jsanda I've refactored FQL reconciliation logic into two functions as requested, it now errors out before creating the STS where an invalid config is passed.

I have comprehensive unit tests for `parseFQLFromConfig`, and have written as many as possible for `SetFullQueryLogging`. There are notes on additional tests we could look to add pending a refactor of the `ReconciliationContext` to make its easier to unit test.

All other changes requested by @burmanm RE `result` types implemented, thanks for the recommendations Michael.

pkg/reconciliation/reconcile_fql_test.go
```diff
@@ -0,0 +1,201 @@
...
+////////////////////////////////////
+var fqlEnabledConfig string = `cassandra-yaml:
+  full_query_logging_options:
+    log_dir: /var/log/cassandra/fql
+`
+
+func Test_parseFQLFromConfig_fqlEnabled(t *testing.T) {
+	// Test parsing when fql is set, should return (true, continue).
+	mockRC, _, cleanupMockScr := setupTest()
+	configJSON, err := createconfigJSON(fqlEnabledConfig)
```
burmanm: Wouldn't it have been easier to just do:

```go
fqlEnabledconfig := `{"cassandra-yaml": { ..`
.Spec.Config = []byte(fqlEnabledConfig)
```

Or ..

```go
fqlEnabledConfig := `[]byte("{\"cassandra-yaml\": ..
.Spec.Config = fqlEnabledConfig
```

?

pkg/reconciliation/reconcile_fql_test.go
```diff
@@ -0,0 +1,201 @@
...
+	mockRC.NodeMgmtClient = client
+	recResult := SetFullQueryLogging(mockRC, true)
+	assert.IsType(t, result.RequeueSoon(2), recResult)
+	cleanupMockScr()
+}
+
+// TDOO: we cannot currently write a func Test_SetFullQueryLogging_ListFail(t *testing.T) {...} test because we cannot inject a mock listPods method into the rc to test this code path.
+
+// TDOO: we cannot currently write a func Test_SetFullQueryLogging_StatusFail(t *testing.T) {...} test because the mocks do not include a mock sts controller to instantiate the pods.
+// While an e2e test could theoretically be written, it would be a significant investment. An ability to override the listPods method with a mock would probably allow us to inject a fake
```
burmanm: This isn't significant work, actually a lot less than this unit test. There's no need to override these methods to test the functioning behavior.

burmanm:
Instead of writing unit tests, it would have been nice to get an actual test (that tests the Reconciliation logic). That's done using the integration tests in tests directory, given that this feature actually requires running Cassandra pod. And the webhook would still provide a way to avoid those wrong config settings for 3.11 (webhooks are the Kubernetes way to verify configuration before applying them).

Now nothing checks if these changes to reconcile logic are broken.

pkg/reconciliation/reconcile_fql_test.go
Miles-Garnsey: While it would have been easier to use JSON directly, I like being able to test this using the native YAML you'd see in `cassandra.yaml` or the cassDC resource. 

I appreciate it does add a lot of code though (two whole functions) so I could be convinced to change this.

pkg/reconciliation/reconcile_fql_test.go
Miles-Garnsey: You're saying an integration test would be easier? I'm open to looking into further integration testing once this is merged. I just don't want to get hung up on it.

Irrespective of the integration tests, I'd still be interesting in factoring the listPods method (and related methods) out into an interface to improve unit testability. I find unit tests useful.

burmanm: In case of reconciliation logic, envtest testing without real pods would be far more useful than unit testing with mocked implementations. I wouldn't waste energy on adding interfaces when we could instead just do the real thing (the dependency to mgmt api is the one blocking use of envtest at the moment IIRC). But for now, the only real way to test that this feature works is building an integration test. For validation, a webhook unit test is enough as that's pretty close to the real thing (the webhook's validation is run with an integration testing also, so it's verified that the method is called - thus adding just new checks there does not require modifying the integration test for webhook).

Miles-Garnsey:
Integration test now added and passing.

tests/config_fql/config_fql_4x_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/k8ssandra/cass-operator/tests/kustomize"
+	ginkgo_util "github.com/k8ssandra/cass-operator/tests/util/ginkgo"
+	"github.com/k8ssandra/cass-operator/tests/util/kubectl"
+)
+
+var (
+	testName     = "FQL config is applied on Cassandra 4x"
+	namespace    = "fql-config"
+	dcName       = "dc1"
+	dcYaml       = "./config_fql_4x_test.yaml"
```
burmanm: This should be in the testate directory

tests/config_fql/config_fql_4x_test.go
```diff
@@ -0,0 +1,84 @@
...
+	operatorYaml = "../testdata/operator.yaml"
```
burmanm: Probably unused?

tests/config_fql/config_fql_4x_test.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc1
+spec:
+  podTemplateSpec:
```
burmanm: Do not define.

tests/config_fql/config_fql_4x_test.yaml
```diff
@@ -0,0 +1,31 @@
...
+    spec:
+      containers:
+        - name: "cassandra"
+          image: k8ssandra/mgmtapi-4_0:latest
+          imagePullPolicy: IfNotPresent
+  clusterName: cluster1
+  serverType: cassandra
+  serverVersion: 4.0.0
+  serverImage: k8ssandra/mgmtapi-4_0:latest
```
burmanm: No serverImage.

burmanm:
Still missing (after rebase, since there's conflicts). These should be relatively quick:

* webhook validation (+ unit test change for that), at least the 3.11.x <-> FQL
* Add the new test to github workflow

tests/config_fql/config_fql_4x_test.go
Miles-Garnsey: I was thinking about this one. What is the reasoning behind putting all the manifests in one directory? I considered doing the same, but it seemed like there was a risk that this would lead to a lot of dead/unused resources sitting in that directory. When a test is no longer needed - the manifests wouldn't be deleted alongside the test directory.  It also makes the link between the test resource and the test itself less clear.

I did also realise that the way I've done it means that parts of the manifest (e.g. the serverVersion) won't be updated from a central location. Would drawing from `config/samples` for a base DC manifest and then using a kustomization in `tests/config_fql` to add the FQL config be a good compromise?

tests/config_fql/config_fql_4x_test.go
Miles-Garnsey: I think `operatorYaml`, `dcResource` are unused. The rest appear to be used. Fixed.

tests/config_fql/config_fql_4x_test.yaml
Miles-Garnsey: Fixed.

tests/config_fql/config_fql_4x_test.yaml
Miles-Garnsey: Makes sense, thanks. Fixed.

pkg/reconciliation/reconcile_fql_test.go
Miles-Garnsey: Integration test added.

pkg/reconciliation/reconcile_fql.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/k8ssandra/cass-operator/pkg/internal/result"
+)
+
+func parseFQLFromConfig(rc *ReconciliationContext) (bool, result.ReconcileResult) {
```
jsanda: `parseFQLFromConfig` could be further refactored to be a stand alone function vs a method of ReconciliationContext. This would make it easier to unit test. I don't think it is a necessary change for this PR though.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2324,6 +2263,11 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
 
 	endpointData := rc.getCassMetadataEndpoints()
 
+	shouldFQLBeEnabled, recResult := parseFQLFromConfig(rc)
```
jsanda: 👍

pkg/reconciliation/reconcile_fql_test.go
jsanda: For high level tests, like envtest tests or full e2e tests I agree that having a YAML manifest is nice because it more blackbox testing. The unit test are white-box test that focus on particular functions/methods that aren't dealing with YAML. The `createconfigJSON` and `convert` functions strike me as unnecessary code that create extra maintenance.

tests/config_fql/config_fql_4x_test.go
jsanda: I agree that the manifest should be in the `testdata` directory. The tests were structured this way well before we came onto the project so we might not be able to give a great answer as to why. If we want to talk about changing it, I think it is a worthwhile discussion, but it is outside the scope of this PR.

Miles-Garnsey:
All comments addressed except for 

* Webhook validation: I would prefer if this was a part of a new ticket if we want additional functionality. I want to move on to other work and don't want this PR sitting open. (I also wouldn't mind doing this since I haven't touched the webhook yet.)
* Some discussion to be had around where manifests get stored for the tests. I've proposed a solution above.

pkg/reconciliation/reconcile_fql.go
Miles-Garnsey: It could take a `v1beta1.CassandraDatacenterSpec`. It can't take a `json.RawMessage` (config struct) because it needs to know the `ServerVersion`. 

I think for consistency with other functions called from `ReconcileRacks` it makes sense to have it take a `ReconciliationContext`, since the rest of the functions there take that parameter type as input. 

We could factor it out so that we have yet another function (`FQLFromRC(rc ReconciliationContext)` calling `FQLFromConfig(rawcfg RawMessage)`) but this seems like indirection for the sake of indirection. Once we start building a config component for `cass-operator` and need validation for it, we may want to refactor as you've suggested.

burmanm:
It isn't only historical reasons for those testdata directories, but also practical in a sense that multiple tests use the same deployment yaml. Not all of them require something special when setting up Cassandra, instead they do changes while Cassandra is running and check that something works that way.

burmanm:
Also, the test failures here are quite valid. The current implementation completely breaks the solution as it doesn't understand that certain versions that were deployed do not have that endpoint in management-api. As of now, this implementation simply requeues all the time. 

For this issue, I've proposed the FeatureSet (which will hopefully be merged to the next version). We should probably use only versions that have that feature to allow setting / checking FQL and otherwise simply pass to the next one. 

One more question though, why poll the management-api at all if the cassandra-yaml section doesn't have that FQL is set to run? Since that method has CassandraDatacenter, shouldn't it start by checking that one (modifications configs will always result in new reconcile).

burmanm:
After some more testing also, I'm struggling to think how this could work without update to cass-config-builder. I couldn't get it to work with the following test:

```sh
#!/bin/bash
mkdir -p output_config
export CONFIG_FILE_DATA='{"cassandra-yaml":{"full_query_logging_options": {"log_dir": "/var/log/cassandra"}, "allocate_tokens_for_local_replication_factor":3,"start_rpc": "true","authenticator":"PasswordAuthenticator","authorizer":"CassandraAuthorizer","credentials_update_interval_in_ms":3600000,"credentials_validity_in_ms":3600000,"num_tokens":16,"permissions_update_interval_in_ms":3600000,"permissions_validity_in_ms":3600000,"role_manager":"CassandraRoleManager","roles_update_interval_in_ms":3600000,"roles_validity_in_ms":3600000},"cluster-info":{"name":"k8ssandra-0","seeds":"k8ssandra-0-seed-service"},"datacenter-info":{"graph-enabled":0,"name":"dc1","solr-enabled":0,"spark-enabled":0},"jvm-server-options":{"additional-jvm-opts":["-Dcassandra.system_distributed_replication_dc_names=dc1","-Dcassandra.system_distributed_replication_per_dc=1"]}}'
export CONFIG_OUTPUT_DIRECTORY="./output_config"
export DEFINITIONS_LOCATION="build/docker/definitions"
export POD_IP="127.0.0.1"
export HOST_IP="127.0.0.2"
export PRODUCT_NAME="cassandra"
export PRODUCT_VERSION="4.0.1"
export RACK_NAME="dummyrack1"
java -jar build/docker/cass-config-builder-1.0.4-20210909-all.jar
```

The output has nothing about FQL:

```console
➜  cass-config-builder git:(master) ✗ grep -i full_query output_config/*
➜  cass-config-builder git:(master) ✗
```

Miles-Garnsey:
@burmanm 

> It isn't only historical reasons for those testdata directories, but also practical in a sense that multiple tests use the same deployment yaml. Not all of them require something special when setting up Cassandra, instead they do changes while Cassandra is running and check that something works that way.

See [Discord thread ](https://discord.com/channels/836217371453685760/897653115912851477/897654769412358165) for a proposed solution.

> One more question though, why poll the management-api at all if the cassandra-yaml section doesn't have that FQL is set to run? Since that method has CassandraDatacenter, shouldn't it start by checking that one (modifications configs will always result in new reconcile).

Because if a DC has FQL set and then it is removed then we need to disable FQL. We poll to confirm that it is disabled. We need logic to both enable and disable. We touched on this [here](https://github.com/k8ssandra/cass-operator/pull/176#discussion_r716480026), but I may not have expanded on it properly.

> Also, the test failures here are quite valid. The current implementation completely breaks the solution as it doesn't understand that certain versions that were deployed do not have that endpoint in management-api. As of now, this implementation simply requeues all the time.

> For this issue, I've proposed the FeatureSet (which will hopefully be merged  to the next version). We should probably use only versions that have that feature to allow setting / checking FQL and otherwise simply pass to the next one.

Can you link to where you've proposed the FeatureSet? That sounds interesting.

I don't know how to solve this issue off the top of my head. Should I examine the management API image before calling the FQL endpoints? Should the management API have a version endpoint that returns the version? We could then check the version endpoint ahead of calling the FQL endpoints on mgmt api.

In the short term, I wonder if it is a problem if the reconciliation just re-queues constantly. What error do you receive back, is it descriptive enough? What are the consequences if someone overrides the default Cassandra image with an older image? If the DC fails to come up and the error message is descriptive enough, this seems like the 'right' behaviour to me.

> After some more testing also, I'm struggling to think how this could work without update to cass-config-builder. I couldn't get it to work with the following test:

I don't have an answer here either, it is a mystery. It is working with the default deployment on my version of master. No explicit changes to the config builder image: `image: datastax/cass-config-builder:1.0.4-ubi7`. Can you use this image and tell me what you see? I don't know what branch/image/whatever you're on.

To re-iterate, the e2e test I have written as requested (`tests/config_fql`) passes. Nonetheless, I'm interested in why the config builder works under these circumstances, perhaps this version already has the commits in it to enable the FQL functionality?

burmanm:
> Can you link to where you've proposed the FeatureSet? That sounds interesting.

https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/141

Basically query /versions/features to get available features in the management-api version.

> In the short term, I wonder if it is a problem if the reconciliation just re-queues constantly. What error do you receive back, is it descriptive enough? What are the consequences if someone overrides the default Cassandra image with an older image? If the DC fails to come up and the error message is descriptive enough, this seems like the 'right' behaviour to me.

It will now requeue in a loop forever without ever proceeding. Customers with DSE deployments or older/custom serverImages are perfectly normal scenarios and should be supported. The reconcile loop should not fail if the management-api is missing a feature.

> I don't have an answer here either, it is a mystery. It is working with the default deployment on my version of master. No explicit changes to the config builder image: image: datastax/cass-config-builder:1.0.4-ubi7. Can you use this image and tell me what you see? I don't know what branch/image/whatever you're on.

My master was 1.0.4 equivalent. The full_query_logging is not in those definition files, so there's no way cass-config-builder could copy them there. It doesn't copy any non-defined values.

Are you sure the test actually tests the correct thing then (it's not enabled in the github workflow so I can't see the result here)? Or perhaps there's a bug in current version of Cassandra that it doesn't need the full_query_logging set in the config. If that's the case, we shouldn't depend on such behavior as 4.0.1 / 4.0.2 etc might fix it to work as intended in the documentation. So we need an updated cass-config-builder with newer definitions.

jsanda:
With respect to adding validation in the webhook, I think it is a great idea but should be addressed in a subsequent PR. We don't need to any additional scope here.

As for the e2e tests and the new management-api endpoint, the tests are failing due to this in the operator log:

```
2021-10-13T03:09:37.714Z        ERROR   controllers.CassandraDatacenter can't get whether query logging enabled for pod         {"cassandradatacenter": "fql-config/dc1", "requestNamespace": "fql-config", "requestName": "dc1", "loopID": "e20d9209-ce0c-433d-a83e-519f6ac166b3", "namespace": "fql-config", "datacenterName": "dc1", "clusterName": "cluster1", "podName": "cluster1-dc1-default-sts-0", "error": "incorrect status code of 404 when calling endpoint"}
github.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).ReconcileAllRacks
        /workspace/pkg/reconciliation/reconcile_racks.go:2365
github.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).CalculateReconciliationActions
        /workspace/pkg/reconciliation/handler.go:160
github.com/k8ssandra/cass-operator/controllers/cassandra.(*CassandraDatacenterReconciler).Reconcile
        /workspace/controllers/cassandra/cassandradatacenter_controller.go:140
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:298
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:253
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:21
``` 
This log is from the `config_fql` test. It is failing as well in GHA. @Miles-Garnsey you are going to have set `serverImage` in your test manifest to an image version that includes the new endpoint. You will also need to make sure that the test is using an appropriate config-builder image version.

The above changes won't address the other failing tests. Where are we with new management-api images? 

>It will now requeue in a loop forever without ever proceeding. Customers with DSE deployments or older/custom serverImages are perfectly normal scenarios and should be supported. The reconcile loop should not fail if the management-api is missing a feature.

At least in the short term I think this could be handled via documentation. 

For this PR if test failures are not resolved with new management-api images, I suggest something simple like an operator config for enabling the functionality.

burmanm:
> The above changes won't address the other failing tests. Where are we with new management-api images?

New management-api images won't solve this issue correctly. These tests should not fail with current management-api images. New management-api images can help to enable the feature, but the 404 handling is still wrong. 

> For this PR if test failures are not resolved with new management-api images, I suggest something simple like an operator config for enabling the functionality.

No. Only this single test should fail (in the end as the fql is not enabled), but not having any errors in the operator-logs. All the other tests should succeed with older management-api.

Miles-Garnsey:
> It will now requeue in a loop forever without ever proceeding. Customers with DSE deployments or older/custom serverImages are perfectly normal scenarios and should be supported. The reconcile loop should not fail if the management-api is missing a feature.

I understand what you're saying now. You're concerned that we're introducing an incompatibility between future versions of cass-operator and any older versions of management API. (I can now also see why you were asking if the FQL endpoint needs to always be polled even if FQL is turned off.)

It looks like we can use your featuresets endpoint which @emerkle826 kindly pushed out over (my) night to get a solution there. I'm happy to add an additional check so that FQL endpoints are only polled when the management API supports the feature.

Nice work on that PR, it will make future feature rollouts much easier.

> Are you sure the test actually tests the correct thing then (it's not enabled in the github workflow so I can't see the result here)? Or perhaps there's a bug in current version of Cassandra that it doesn't need the full_query_logging set in the config. If that's the case, we shouldn't depend on such behavior as 4.0.1 / 4.0.2 etc might fix it to work as intended in the documentation. So we need an updated cass-config-builder with newer definitions.

It is [passing](https://github.com/k8ssandra/cass-operator/pull/176/checks?check_run_id=3888235398) in the GHAs. Every other test is failing though. 

That test is definitely tests the right things, if you check the logs I've linked you'll see the FQL related files appearing in the directory. If you recall two weeks ago I did a demo showing that the files definitely didn't appear when FQL was turned off, and then showed up when it was turned on - so FQL isn't enabled by default in  4x.

I do have reservations about the rest of the tests now failing(!!) and will investigate this today.

jsanda:
With respect to the new feature set endpoint, I think we still need an operator config setting to enable it. 

More importantly, there needs to be a broader discussion about version compatibility.

burmanm:
> With respect to the new feature set endpoint, I think we still need an operator config setting to enable it.

Why? Do we explicitly want to disable (even if the pod supports it) it? So the users can't set it?

> More importantly, there needs to be a broader discussion about version compatibility.

cass-operator is supposed to support DSE 6.8.0 and up. There's no need to restrict this supported version set as we can always just ask the pod if it supports something or not.

burmanm:
> That test is definitely tests the right things, if you check the logs I've linked you'll see the FQL related files appearing in the directory. If you recall two weeks ago I did a demo showing that the files definitely didn't appear when FQL was turned off, and then showed up when it was turned on - so FQL isn't enabled by default in 4x.

That is not what I was saying. Why is it passing without the full_query_logging_options in the. config? cass-config-builder is stripping it out, so what's happening here?

burmanm:
Okay, now I can replicate it locally with cass-config-builder also, it indeed does put values that are not defined.. odd.

burmanm:
After PR #202 is merged, you can apply the following patch and it should fix all the test errors:

https://gist.github.com/burmanm/34930cdcdb62e6317582cb7232d83512

Miles-Garnsey:
Update;

* Test failures were caused by 3x deployments checking for FQL when FQL functionality not available for that version. I broke this while refactoring and somehow missed it. Many apologies, fixed.
* I refactored the FQL reconciliation logic (again) to avoid parsing the major version twice. I think this is clunky but in the absence of a dedicated validation package (under discussion) it will do.
* My local testing was failing all day yesterday because I only had a 3 node kind cluster and the test I was running needed 6 nodes. Idiot!
* I've now done some local testing and it all looks fine.

Once Michael's [PR 202](https://github.com/k8ssandra/cass-operator/pull/202) is merged I will adopt a patch similar to the one [he proposed](https://gist.github.com/burmanm/34930cdcdb62e6317582cb7232d83512) which checks for the presence of the FQL features. This will stay in a holding pattern until then.

Miles-Garnsey:
I've made the changes to 

* Check the new management API featuresset endpoint to confirm the FQL endpoints are available.
*  Move test resources into testdata. (I want to move this elsewhere but need to implement kustomizations to move down that path. We'll keep it consistent for now)
* Cater better to DSE including when the user is trying to use FQL on a DSE server.

Ready for another review.

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 * [BUGFIX] #141 #110 Force update of HostId after pod replace process to avoid stale HostId
 * [BUGFIX] #139 Bundle creation had some bugs after #120
 * [BUGFIX] #134 Fix cluster-wide-installation to happen with Kustomize also, from config/cluster
-
+* [ENHANCEMENT] #175 Add FQL reconciliation via parseFQLFromConfig and SetFullQueryLogging called from ReconcileAllRacks. CallIsFullQueryLogEnabledEndpoint and CallSetFullQueryLog functions to httphelper.
```
burmanm: This is in wrong place, should be at the unreleased part.

pkg/reconciliation/reconcile_fql.go
```diff
@@ -0,0 +1,100 @@
...
+			if serverMajorVersion < 4 || dc.Spec.ServerType != "cassandra" {
+				err := fmt.Errorf("full query logging only supported on OSS Cassandra 4x+")
+				rc.ReqLogger.Error(err, "full_query_logging_options is defined in Cassandra config, it is not supported on the version of Cassandra you are running")
+				return false, serverMajorVersion, result.Error(err)
+			}
+			rc.ReqLogger.Info("full_query_logging_options is defined in Cassandra config, we will try to enable it via the management API")
+			shouldFQLBeEnabled = true
+		}
+	}
+	if dc.Spec.ServerType != "cassandra" {
```
burmanm, line 50: Shouldn't this be before all the other logic? Unnecessary to do any work if the server is not ``cassandra``

pkg/reconciliation/reconcile_fql.go
```diff
@@ -0,0 +1,100 @@
...
+		for _, podPtr := range PodPtrsFromPodList(podList) {
+			features, err := rc.NodeMgmtClient.FeatureSet(podPtr)
+			if err != nil {
+				rc.ReqLogger.Error(err, "failed to verify featureset for FQL support")
+				return result.RequeueSoon(2)
+			}
+			if !features.Supports(httphelper.FullQuerySupport) {
+				if enableFQL {
+					err := errors.New("FQL should be enabled but we cannot verify if FQL is supported by mgmt api")
+					return result.Error(err)
```
burmanm, line 77: This will cause a loop. You can log, but do not return result.Error() or the reconcile will never finish.

CHANGELOG.md
Miles-Garnsey: Good catch, thanks.

pkg/reconciliation/reconcile_fql.go
```diff
@@ -0,0 +1,100 @@
...
+				} else {
```
burmanm: The else clause can be left out, this part will be executed anyway (go-staticcheck should complain about this)

pkg/reconciliation/reconcile_fql.go
Miles-Garnsey: I think this is the behaviour we want. If FQL is not supported at the management API layer but the user has requested it, the DC should not be created.

tests/config_fql/config_fql_4x_test.go
```diff
@@ -0,0 +1,82 @@
...
+			lsCmd := kubectl.KCmd{
+				Command: "exec",
+				Args:    []string{"cluster1-dc1-default-sts-0", "--", "ls", "/var/log/cassandra/fql"},
+				Flags:   map[string]string{"container": "cassandra"},
+			}
+			Eventually(func() bool {
+				res, _, err := ns.ExecVCapture(lsCmd)
+				if err != nil {
+					fmt.Println(err)
+					Fail("error running ls on Cassandra container")
```
burmanm, line 65: Don't fail inside Eventually, or this isn't Eventually at all.

burmanm: (I simplified this test in the patch I mentioned)

pkg/reconciliation/reconcile_fql.go
Miles-Garnsey: In your original patch it was, however we need to determine whether FQL is turned on in the `spec.config` and throw an error if so. Otherwise DSE clusters trying to use FQL will just silently fail to enable it.

pkg/reconciliation/reconcile_fql.go
Miles-Garnsey: Is it more idiomatic in Go to leave out the `else`?  It seemed clearer to me to include it but I don't mind either way.

tests/util/ginkgo/lib.go
```diff
@@ -54,10 +54,14 @@ func NewWrapper(suiteName string, namespace string) NsWrapper {
 	}
 }
 
+// Run KCmd via `kubectl` in the namspace (the receiver), where KCmd is a struct holding the kubectl command to run (not including `kubectl` itself), the args, and any flags.
```
burmanm: For public methods, the golang's documentation style starts with the function name. So this should be something like:

``// ExecVCapture runs ..``

Otherwise the generated godoc (kinda like javadoc) will not set this comment correctly.

tests/config_fql/config_fql_4x_test.go
Miles-Garnsey: They are different failures (I think). One detects whether the ls command is failing to run (should be fatal) while the other asserts that eventually the command returns the expected output.

Miles-Garnsey:
Seeing a GHA test failure here when I can get a pass on my local machine. 

Steps to make this pass (from within this branch):

```
setup-kind-multicluster.sh --kind-worker-nodes 1
kind load docker-image --name k8ssandra-0 k8ssandra/cass-operator:latest
kind load docker-image --name k8ssandra-0 k8ssandra/cass-management-api:4.0.0
kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
M_INTEG_DIR=config_fql
cd tests/${M_INTEG_DIR} && go test -v ./... -timeout 300m --ginkgo.progress --ginkgo.v
```

Have pinged @emerkle826 to see if this might be due to the `k8ssandra/cass-management-api:4.0.0` public image not incorporating last night's changes.

pkg/reconciliation/reconcile_fql.go
burmanm: Yeah, it is more idiomatic (and helps with my editor not spewing so many yellow lines ;))

Miles-Garnsey: OK, if it makes your IDE happier :)

burmanm (approved):

PR #180, opened 2021-09-26
Component based installation for Kustomize

### LINKED ISSUE #167: K8SSAND-877 ⁃ Default kustomization should not create namespace
**What happened?**
If I create the following:

```
CASS_OPERATOR_HOME=$(mktemp -d)
cat <<EOF >$CASS_OPERATOR_HOME/kustomization.yaml
resources:
- github.com/k8ssandra/cass-operator/config/default?ref=v1.8.0-rc.1
EOF
```

and then run:

```
kustomize build $CASS_OPERATOR_HOME | kubectl apply -f - 
```

cass-operator is deployed in the `cass-operator` namespace. This is basically just running the code generated by operator-sdk (or kubebuilder). It is perfectly fine if only want to deploy cass-operator.

If I want to create a kustomization that includes cass-operator as well other another operator (e.g., k8ssandra-operator) and want to deploy them in the same namespace, then this is a problem if both operators rely on that generated code. 

`kustomize build` will fail with an error that mentions `namespace transformation produces ID conflict`.

The problem is the namespace declaration in `config/manager/manager.yaml`. Let's suppose we have another operator, test-operator, that was also created with operator-sdk. I want to deploy both cass-operator and test-operator in the `test-operator` namespace. I could create the following:

```
KUSTOMIZE_HOME=$(mktemp -d)
cat <<EOF >$KUTSOMIZE_HOME/kustomization.yaml
namespace: test-operator
resources:
- github.com/k8ssandra/cass-operator/config/default?ref=v1.8.0-rc.1
- github.com/jsanda/test-operator/default?ref=main
EOF
```

This will fail when I run `kustomize build` because we create a Namespace object twice. I realize that this is what operator-sdk/kubebuilder generates, but after thinking about it for a while I think it is wrong or incomplete. 

There should be an overlay directory that does create the namespace. This would be appropriate to use when you want to deploy the operator on its own. There should be another overlay directory that does not create the namespace. This would be appropriate to use for composition, i.e., when you want to include the operator with other components. The one that does not include the namespace should the base since overlays can be built from it.

* Cass Operator version:
1.8.0-rc.1

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-877) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-877
┆priority: Medium

burmanm:
**What this PR does**:
Creates new structure of components to allow modification of the cass-operator installation.

Fixes #167

**Checklist**
- [x] Changes manually tested


burmanm:
Replaces #174

config/components/namespace/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+resources:
```
jsanda, line 4: I really like the changes inside of the `config` directory. If we make similar change in k8ssandra-operator, will we then be able to install both operators configured to be cluster-scoped and deployed into their own, respective namespaces?

burmanm: With a single kustomize file or in general? I think creating a layer on top would sadly overwrite the namespace for both, so it would require two kustomize builds..

jsanda (approved):
nice work!

PR #182, opened 2021-09-28
Non-root applied to server-logger

### LINKED ISSUE #157: K8SSAND-791 ⁃ Cass operator Image 1.7.1 running as root user
**What happened?**
A non default user is not utilized in the image. This is a security concern. 

Expected to see a non-root user such as cass-operator with restricted access. 

**Environment**
kubernetes 1.19

* Cass Operator version:
1.7.1
k8ssandra/cass-operator
sha256:dd57363b07948560e683610aa11715919b78f1c9a2671f0f4a1fb43a8f0f22b9
    
    
* Kubernetes cluster kind:
* Manifests:

```
```

* Cass Operator Logs:

```
```

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-791) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-791
┆priority: Medium

jeffbanks:
**What this PR does**:
Security applied to cass-operator.

Fixes 
#157 & [k8ssand-823](https://k8ssandra.atlassian.net/browse/K8SSAND-823), part of overall k8ssandra security [epic](https://k8ssandra.atlassian.net/browse/K8SSAND-705)

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/logger.Dockerfile
+++ b/logger.Dockerfile
@@ -4,6 +4,9 @@ FROM redhat/ubi8-micro:latest
 ENV TINI_VERSION v0.19.0
 ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /sbin/tini
 RUN chmod +x /sbin/tini
+
+# Non-root user, cassandra as default
+USER cassandra:cassandra
 ENTRYPOINT ["/sbin/tini", "--"]
 
 # Run your program under Tini


jeffbanks:
Verified with mgmt-api image, the permission state will be applied as follows:
```
drwxrwxrwx 1 cassandra cassandra   4096 Sep 28 20:03 cassandra
-rw-r--r-- 1 cassandra cassandra    52807 Sep 28 21:02 system.log
```
Since the system-logger is simply tailing the system.log, the `uid: 65532` will be used to perform the tail command.

logger.Dockerfile
```diff
@@ -4,6 +4,9 @@ FROM redhat/ubi8-micro:latest
 ENV TINI_VERSION v0.19.0
 ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /sbin/tini
 RUN chmod +x /sbin/tini
+
+# Non-root user
+USER 65532:65532
```
jsanda: Do we need this?

cass-operator already adds a pod security context by default:

```yaml
  securityContext:
    fsGroup: 999
    runAsGroup: 999
    runAsUser: 999
```

emerkle826: Is `65532` a specific `uid`/`gid` that is required here, or just something that isn't `0`/root? If it doesn't matter, maybe go with `999` for consistency?

jsanda: We just want non-root, but is it even necessary to add the `USER` instruction since cass-operator sets it by default?

jeffbanks: @jsanda 
Good point!  My understanding is that the container's `securityContext` will override the pod's `securityContext`.  And either of these will override the image's USER defined in the Dockerfile.

The question comes for k8ssandra related containers overall.  Do we want to provide group and user assignment at the securityContext only?  Speaking for our managed images in this case.

One thought is that if images are being utilized/executed (e.g. unit/integration tests) outside of an operator, a default user should still be applied at the Dockerfile level.  If we don't enforce the USER assignment as `non-root` in the docker file, this opens the image up to running as `root` in some cases.  Also, covers the case where there is a bug in the operator or there isn't security applied properly.  The image would still default to a non-root user.

Where I see this helping the most, is when used as a standalone test of the image, and this non-root default will prevent it from having root capabilities.  Otherwise, we may get false positives on the test runs due to root's capabilities for reading, writing, etc.

jeffbanks: @jsanda @emerkle826 
Are we good here with having the Dockerfile include the USER to protect us from defaulting to something that _could_ be less safe?

I'm o.k. with `cassandra` user since we are discussing the cass-operator, but the basic convention used in other images and [here](https://github.com/k8ssandra/cass-operator/blob/38f22c25fdb7f3c366adac06fb879f063e8ab797/config/manager/manager.yaml#L21) is the `65532` uid.

jsanda: The example you referenced is for the security context used for the operator pod itself. If we are going to add the `USER` instruction then I think it makes sense to set it to the cassandra uid since that it was it is actually set to by default.

emerkle826: > The example you referenced is for the security context used for the operator pod itself. If we are going to add the `USER` instruction then I think it makes sense to set it to the cassandra uid since that it was it is actually set to by default.

:+1:

jeffbanks: Updated to cassandra:cassandra in the image

jsanda (approved):
There is one integration test failing, but I don't think it is related.

PR #193, opened 2021-10-07
Add new Management API endpoints to HTTP Helper: GetKeyspaceReplication, ListTables, CreateTable [K8SSAND-926]

adutra:
**What this PR does**:

This PR adds new Management API endpoints to HTTP Helper: GetKeyspaceReplication, ListTables, CreateTable.

It requires https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/143.

This PRs is part of the fix for https://github.com/k8ssandra/k8ssandra-operator/issues/156.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pkg/httphelper/client.go
+++ b/pkg/httphelper/client.go
@@ -365,6 +365,167 @@ func (client *NodeMgmtClient) ListKeyspaces(pod *corev1.Pod) ([]string, error) {
 	return client.GetKeyspace(pod, "")
 }
 
+// GetKeyspaceReplication calls the management API to retrieve the replication settings of the
+// given keyspace.
+func (client *NodeMgmtClient) GetKeyspaceReplication(pod *corev1.Pod, keyspaceName string) (map[string]string, error) {
+	if keyspaceName == "" {
+		return nil, fmt.Errorf("keyspace name cannot be empty")
+	}
+	podHost, err := BuildPodHostFromPod(pod)
+	if err != nil {
+		return nil, err
+	}
+	endpoint := "/api/v0/ops/keyspace/replication?keyspaceName=" + keyspaceName
+	request := nodeMgmtRequest{
+		endpoint: endpoint,
+		host:     podHost,
+		method:   http.MethodGet,
+		timeout:  time.Second * 20,
+	}
+	body, err := callNodeMgmtEndpoint(client, request, "application/json")
+	if err != nil {
+		return nil, err
+	}
+	var replication map[string]string
+	if err := json.Unmarshal(body, &replication); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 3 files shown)


burmanm:
What if the pod that's being called does not have that information? Can you check that with the error code?

adutra:
> What if the pod that's being called does not have that information? Can you check that with the error code?

If the pod fails, or if it replies with a status code >= 200 and < 300, the call fails and an error is returned to the caller. This behavior is the same for all methods in `NodeMgmtClient`.

From what I understand of the `NodeMgmtClient` API, it is designed in such a way that retrying on another pod is the caller's responsibility. This is why we have `ManagementApiFacade` in k8ssandra-operator.

burmanm:
Code 404 and 500 were my interest though. I did actually implement a featureSet improvement in a PR to management-api to be able to see what features are available, but perhaps such check could be done later.

pkg/httphelper/client.go
```diff
@@ -320,6 +320,167 @@ func (client *NodeMgmtClient) ListKeyspaces(pod *corev1.Pod) ([]string, error) {
...
+
+func NewRegularColumn(name string, dataType string) *ColumnDefinition {
+	return &ColumnDefinition{
+		Name: name,
+		Type: dataType,
+		Kind: ColumnKindRegular,
+	}
+}
+
+func NewStaticColumn(name string, dataType string) *ColumnDefinition {
```
burmanm, line 489: Common to all these functions, why are they dealing with pointers all the time?

adutra: Just to avoid passing these structs by value and copying their contents each time. Would you prefer to pass them by value instead?

adutra:
> Code 404 and 500 were my interest though. I did actually implement a featureSet improvement in a PR to management-api to be able to see what features are available, but perhaps such check could be done later.

I confess I don't see exactly what you mean here.

Wrt to 404 and 500 codes though: it is true that the API is sometimes surprising:

* ListTables and ListKeyspaces return OK + empty array if the keyspace or table does not exist.
* GetReplication returns 404 if the keyspace does not exist.
* AlterKeyspace returns 500 if the keyspace does not exist.

Do you mean that we should harmonize which HTTP codes to return?

burmanm:
Those error codes and also the fact that how do you find the difference between 404 -> too old management-api and not-found? I understand 404 might be a correct return code, but we have an issue here that we can't detect if that method is there.

burmanm:
Can you rebase so I can merge?

adutra:
> Can you rebase so I can merge?

Done.

burmanm (approved):

PR #202, opened 2021-10-14
Implement /versions/feature endpoint and Supports feature in the client

burmanm:
**What this PR does**:
Adds ability to poll the management-api what features it provides.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


burmanm:
I need one more fix to this for DSE..

pkg/httphelper/client.go
```diff
@@ -74,6 +74,40 @@ func newNoPodIPError(pod *corev1.Pod) NoPodIPError {
 	return fmt.Errorf("pod %s has no IP", pod.Name)
 }
 
+type Feature string
+
+type FeatureSet struct {
+	CassandraVersion string
```
Miles-Garnsey, line 89: I'm just thinking about this (and I realise I'm late to the party with the management API PR already merged). Would it be worth having both a CassandraVersion and a MgmtAPIVersion here? These two can be mixed and matched and what we're really after for my PR is the management API version (since we can infer FQL availability in Cassandra from the Cassandra version which we already have).

Miles-Garnsey:
I've done a read through and it looks great in most respects. The only thing that I keep asking myself is whether we should be using the swagger docs instead of a custom defined FeatureSet. 

I know that the swagger docs are not currently complete/accurate. My question: if they are accurate for both AsyncSSTables and FQL, is that enough for us to use them to describe supported features now? (Even acknowledging other endpoints may not be accurately represented.)

The alternative here is that we keep building on the FeatureSet logic until the swagger docs are all validated and then move to using them. That sounds like a more complex path though since we'd have to deprecate the FeatureSet API, and we then have two different discovery mechanisms for some time. << I am OK with this if it is the path we choose to go down, but suggest we think about how it will look in 1-2 years.

_(I'd also like to see the mgmt API version and perhaps whether it is a DSE or Cassandra server included in the response, but both of these can be put into a later ticket once/if we need them, they don't need to be in scope here.)_

pkg/httphelper/client.go
burmanm: (for clarity I'll reply here also). The management-api doesn't know its own version as of yet, that's why the reply is always empty. Once that's done, then this will be modified also. But mgmtVersion <-> featuremapping is a difficult case also.

Miles-Garnsey: I'm good with that answer :)

Miles-Garnsey (approved):
From discussion tonight it seems we all agree that this should proceed. 

It would be nice if we could use error codes from the management API or the openAPI spec to determine supported features, but we need to mature the management API before going down that path.

The changes in this PR will be a good solution until that happens. Thanks @burmanm!

PR #216, opened 2021-10-22
Modify FQL reconcile function to be part of the ReconciliationContext

### LINKED ISSUE #214: Add FQL validation to the validating Webhooks
**What is missing?**
FQL support is doing its validation only in the reconcile phase, but we should prevent users from deploying invalid CassandraDatacenter to the api-server. This would be prevented by adding the same validation rules to the cassandradatacenter_webhook.go

**Why do we need it?**
If a user sends an invalid configuration to the api-server, the FQL process will block all the other features (such as replacing dead nodes) until the user sees from the logs that their configuration was incorrect.

burmanm:
**What this PR does**:
CheckFullQuery will now be part of the ReconciliationContext, while the parsing and the validation will be in the CassandraDatacenter. This allows to reuse the validation and parsing logic in the webhook and in the handler's isValid functionality (which is ran before the actual reconcile)

Fixes #214 

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
I did have the functions set up like this originally, but @jsanda suggested making these standalone functions instead of methods on ReconciliationContext (because it makes them easier to test.)

I did agree with him at the time but if you feel strongly that the FQL methods should be consistent with the current methods I'm OK with approving this.

burmanm:
I don't see how they're easier to test, since they require the ReconciliationContext in any case. I'd rather have the consistency if there's no real benefits to keep them outside. These are not stateless functions, but bound to the RC.

Miles-Garnsey:
From my observation, I've noticed that methods defined on concrete types in Go are harder to *use in tests* because they cannot be mocked/stubbed. 

For example, imagine I was testing a function which calls `parseFQLFromConfig()` but I want `parseFQLFromConfig()` to return a specific result (e.g. an error). This becomes hard to achieve if it is a method of `ReconciliationContext` because I cannot override the behaviour of that single function within `ReconciliationContext`. 

So I can't override the function with a stub to get the response I need. But I also can't implement a mock of `ReconciliationContext` and override `parseFQLFromConfig()` there, because it is a concrete type and any function that takes `ReconciliationContext` needs to take that exact type. So I don't have any option to ever stub `parseFQLFromConfig()`

It basically means that you can't unit test any function which calls the FQL functions in isolation. Let me know if I'm wrong here, I've only figured out this behaviour in Go in the past month as I've started to add unit tests to the code.

pkg/reconciliation/reconcile_fql.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // parseFQLFromConfig parses the DC config field to determine whether FQL should be enabled based on the presence of full_query_logging_options within the cassandra-yaml field.
 // To ease integration into the reconciliation process, it returns (shouldFQLBeEnabled, serverMajorVersion, ReconcileResult) where the ReconcileResult may be result.Error() or result.Continue().
-func parseFQLFromConfig(rc *ReconciliationContext) (bool, int64, result.ReconcileResult) {
+func (rc *ReconciliationContext) parseFQLFromConfig() (bool, int64, result.ReconcileResult) {
```
jsanda: As I had mentioned in Miles' PR, I would rather see `parseFQLFromConfig` be stand alone function. It would easier to unit test. I don't see why it needs to be a method of ReconciliationContext.

pkg/reconciliation/reconcile_fql.go
```diff
@@ -56,7 +56,7 @@ func parseFQLFromConfig(rc *ReconciliationContext) (bool, int64, result.Reconcil
 
 // SetFullQueryLogging sets FQL enabled or disabled based on the `enableFQL` parameter, and takes serverMajorVersion for additional validation.
 // It calls the NodeMgmtClient which calls the Cassandra management API and returns a result.ReconcileResult.
-func SetFullQueryLogging(rc *ReconciliationContext, enableFQL bool, serverMajorVersion int64) result.ReconcileResult {
+func (rc *ReconciliationContext) SetFullQueryLogging(enableFQL bool, serverMajorVersion int64) result.ReconcileResult {
```
jsanda: Changing this `SetFullQueryLogging` makes more sense to me considering it needs access to `NodeMgmtClient`.

pkg/reconciliation/reconcile_fql.go
burmanm: How is it easier to unit test? It's not anonymous function as a parameter anywhere. If it's not stateful, then removing rc would make sense.

burmanm:
Rebased to fix the GHA errors (merge process is a bit odd, it merged new tests, but not new actions from .github). And although the commit itself looks odd on Github, basically the SetFullQueryLogging name was only changed, but parseFQL was moved to apis.

Miles-Garnsey:
@burmanm I'm marked as a reviewer and haven't come back to you as soon as I should, apologies. Has this PR evolved from refactoring FQL functions to moving validation logic in the webhook instead of the controller?

Should we discuss this design on ticket #214 as @jsanda and I had some ideas on this. (We might also want to change the name of this PR to reflect any new direction?)

I don't have any issues with doing validation in webhooks, I can see the advantages. I do want to explore what happens though when `cass-operator` is deployed without the webhook, whether that causes failures, and what they look like. From my understanding the webhook is an optional component?

burmanm:
This is a stricter form of validation, it will still reject it even if there's no webhook installed.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -747,6 +748,28 @@ func (dc *CassandraDatacenter) GetContainerPorts() ([]corev1.ContainerPort, erro
 	return ports, nil
 }
 
+func (dc *CassandraDatacenter) GetFullQueryStatus() (bool, error) {
```
jsanda: I find the method name a bit confusing. I was expecting the method to return something based on the status. What do you think something like `FullQueryLoggingEnabled` or `FQLEnabled`?

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -226,3 +227,29 @@ func (dc *CassandraDatacenter) ValidateUpdate(old runtime.Object) error {
 func (dc *CassandraDatacenter) ValidateDelete() error {
 	return nil
 }
+
+var (
+	ErrFQLNotSupported = fmt.Errorf("full query logging is only supported on OSS Cassandra 4.0+")
+)
+
+func (dc *CassandraDatacenter) ValidateFQLConfig() error {
```
jsanda: I see where `ValidateFQLConfig` is called in the Reconciler. Is it also called in the webhook? Not sure if I just missed it. If it isn't call in the webhook, it's fine with me if that change is done in a subsequent PR.

jsanda (commented):
The refactoring is pretty much in line with what I had in mind in terms of improving the testability of the code, particularly with the `GetFullQueryStatus` and `ValidateFQLConfig` methods.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -747,6 +748,28 @@ func (dc *CassandraDatacenter) GetContainerPorts() ([]corev1.ContainerPort, erro
 	return ports, nil
 }
 
+func (dc *CassandraDatacenter) FullQueryEnabled() (bool, error) {
```
Miles-Garnsey, line 752: Can we make this `GetFullQueryEnabled() (` to avoid confusion with `enabled:` keys in k8ssandra's helm charts?

Also, I think my comments from earlier on apply here. I'm not sure that creating methods on concrete types is a good idea as the method then can't be stubbed when unit testing other functions that call it. Could this be a function like `GetFullQueryEnabled(dc *CassandraDatacenter)` instead of a method of `CassandraDatacenter`?

apis/cassandra/v1beta1/webhook_test.go
```diff
@@ -758,3 +759,106 @@ func Test_ValidateDatacenterFieldChanges(t *testing.T) {
...
+
+var fqlEnabledConfig string = `{"cassandra-yaml": { 
+	"full_query_logging_options": {
+		"log_dir": "/var/log/cassandra/fql" 
+		}
+	}
+}
+`
+
+func getCassDc(serverType string) CassandraDatacenter {
```
Miles-Garnsey: Can this be `NewCassDC` so it is clear that we are creating something rather than returning an existing DC object?

pkg/reconciliation/reconcile_fql.go
```diff
@@ -1,98 +1,53 @@
...
-	"strings"
 
 	"github.com/k8ssandra/cass-operator/pkg/httphelper"
 	"github.com/k8ssandra/cass-operator/pkg/internal/result"
 )
 
-// parseFQLFromConfig parses the DC config field to determine whether FQL should be enabled based on the presence of full_query_logging_options within the cassandra-yaml field.
-// To ease integration into the reconciliation process, it returns (shouldFQLBeEnabled, serverMajorVersion, ReconcileResult) where the ReconcileResult may be result.Error() or result.Continue().
-func parseFQLFromConfig(rc *ReconciliationContext) (bool, int64, result.ReconcileResult) {
+// CheckFullQueryLogging sets FQL enabled or disabled based on the `enableFQL` parameter, and takes serverMajorVersion for additional validation.
```
Miles-Garnsey: This isn't quite accurate since it doesn't take an `enableFQL` parameter or serverMajorVersion.

pkg/reconciliation/reconcile_fql.go
```diff
@@ -1,98 +1,53 @@
...
+// It calls the NodeMgmtClient which calls the Cassandra management API and returns a result.ReconcileResult.
+func (rc *ReconciliationContext) CheckFullQueryLogging() result.ReconcileResult {
 	dc := rc.GetDatacenter()
-	serverMajorVersion, err := strconv.ParseInt(strings.Split(dc.Spec.ServerVersion, ".")[0], 10, 8)
+	enableFQL, err := dc.FullQueryEnabled()
```
Miles-Garnsey, line 17: I'm a bit confused in this part. It looks like we are determining whether to enable FQL, but we aren't validating whether we actually can validate FQL. 

I'm wondering what happens if the user is not using the webhook (which does do the validation) and tries to enable FQL on DSE (for example). We'd end up in a requeue loop - which is fine - but we should probably error out?

Should we consider calling `ValidateFQLConfig()` in the controllers as well as the webhooks to avoid that sort of issue?

pkg/reconciliation/reconcile_fql.go
```diff
@@ -1,98 +1,53 @@
...
+
+	podList, err := rc.listPods(rc.Datacenter.GetClusterLabels())
+	if err != nil {
+		rc.ReqLogger.Error(err, "error listing all pods in the cluster to progress full query logging reconciliation")
+		return result.RequeueSoon(2)
+	}
+	for _, podPtr := range PodPtrsFromPodList(podList) {
+		features, err := rc.NodeMgmtClient.FeatureSet(podPtr)
+		if err != nil {
+			rc.ReqLogger.Error(err, "failed to verify featureset for FQL support")
```
Miles-Garnsey, line 30: Could this error message be "failed to call NodeMgmtClient.FeatureSet()", and could we add the pod IPs or hostnames?

I'm a bit fanatical about informative error messages sorry😅  In this case we aren't failing to verify the featureSet, we just aren't able to get it at all from the HTTP client.

pkg/reconciliation/reconcile_fql.go
```diff
@@ -1,98 +1,53 @@
...
 	"github.com/k8ssandra/cass-operator/pkg/httphelper"
 	"github.com/k8ssandra/cass-operator/pkg/internal/result"
 )
 
-// parseFQLFromConfig parses the DC config field to determine whether FQL should be enabled based on the presence of full_query_logging_options within the cassandra-yaml field.
-// To ease integration into the reconciliation process, it returns (shouldFQLBeEnabled, serverMajorVersion, ReconcileResult) where the ReconcileResult may be result.Error() or result.Continue().
-func parseFQLFromConfig(rc *ReconciliationContext) (bool, int64, result.ReconcileResult) {
+// CheckFullQueryLogging sets FQL enabled or disabled based on the `enableFQL` parameter, and takes serverMajorVersion for additional validation.
+// It calls the NodeMgmtClient which calls the Cassandra management API and returns a result.ReconcileResult.
+func (rc *ReconciliationContext) CheckFullQueryLogging() result.ReconcileResult {
```
Miles-Garnsey, line 11: My earlier comments RE methods on concrete types apply here. It will be very hard to stub out this method to unit test any callers. Could we consider changing it to `func CheckFullQueryLogging(rc *ReconciliationContext)`?

Miles-Garnsey (changes requested):
This was a bit hard to get my head around as a lot of the logic is moved around. 

I think it is good, and I'd ask that we consider the following changes - 

1. I think we should call `ValidateFQLConfig()` in `CheckFullQueryLogging()` to cover cases where the webhook isn't running. If the config doesn't validate (e.g.  Cassandra < 4 or DSE is running) then we should raise an informative error message.
2. I'm not clear on we make `rc.CheckFullQueryLogging()` and `dc.FullQueryEnabled()` methods of `ReconciliationContext` and `CassandraDatacenter`. If they were functions of these types instead, then we can stub them more easily to unit test their callers. 
3. I can see both `ValidateFQLConfig()` and `dc.FullQueryEnabled()` are tested in `webhook_test.go`. My only question is whether `CheckFullQueryLogging()` should have some unit tests too? I'm less interested in the happy path there, but I'd like to see what happens if it receives invalid configs etc. (obviously this connects to point 1 above).

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: You can't stub methods like that in any case, since this isn't an interface. I don't follow the enabled part of k8ssandra, this isn't a Helm chart.

pkg/reconciliation/reconcile_fql.go
burmanm: The ValidateFQLConfig() is called in the reconciliation method, it's done in the isValid.

apis/cassandra/v1beta1/cassandradatacenter_types.go
Miles-Garnsey: If you really had to stub it you could pass the `FullQueryEnabled()`function as a parameter to the caller. You can't even do that though if it is a method of `CassandraDatacenter`. I'd 100% support you making `CassandraDatacenter` an interface though, it would be quite helpful in testing 😉

pkg/reconciliation/reconcile_fql.go
Miles-Garnsey: Ah, OK sorry. I missed that as I haven't played around in `handler` too much. You're right, ignore me :)

pkg/reconciliation/reconcile_fql.go
burmanm: The connection error text comes in the err part of the logging. So that's not removed.

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: A struct shouldn't be an interface. Never. I think that approach is looking Go from wrong direction, this isn't Java and its reflection capabilities. 

This is a method that works as a method for CassandraDatacenter and should never be stubbed, if you have to do that, then you're doing something wrong. This is very stateful one.

pkg/reconciliation/reconcile_fql.go
burmanm: A method alone shouldn't be stubbed in this case, as the test should verify the behavior of ReconciliationContext. If you need to test something else, then make it accept something else than ReconciliationContext as input.

Miles-Garnsey: I'll agree to disagree here too. I'm not going to get dogmatic about the 'right way' to do testing (if I was even 100% certain of it myself 🙂)

apis/cassandra/v1beta1/cassandradatacenter_types.go
Miles-Garnsey: > A struct shouldn't be an interface. Never. I think that approach is looking Go from wrong direction, this isn't Java and its reflection capabilities.

So I'm not proposing that the struct be an interface. I'm proposing to embed an interface `FQLthingDoer` in `CassandraDatacenter` and attaching FQL functions to it. 

I'm not sure where reflection comes into this, and I steer people away from using reflection in JVM languages because it is almost always bad practice IMO.

_* NB: `FQLFunctionDoer` is clearly a placeholder name and we should come up with something better, please not `FQLManager` because I hate that Javaism too_ 

> This is a method that works as a method for CassandraDatacenter and should never be stubbed, if you have to do that, then you're doing something wrong. This is very stateful one.

I don't understand what statefulness has to do with stubbing methods. I think we disagree on how to construct unit tests. I'm not one to be dogmatic on this, so I'll resolve this comment, but I just want to make sure I'm expressing my point clearly.

Miles-Garnsey:
All of my questions have been addressed here except for one:

> I can see both ValidateFQLConfig() and dc.FullQueryEnabled() are tested in webhook_test.go. My only question is whether CheckFullQueryLogging() should have some unit tests too? I'm less interested in the happy path there, but I'd like to see what happens if it receives invalid configs etc. (obviously this connects to point 1 above).

If we can add some unit tests for `CheckFullQueryLogging()` then I'm happy to approve 🙂

burmanm:
CheckFullQueryLogging is the method you created ;)

burmanm:
Added unit tests and renames.

pkg/reconciliation/reconcile_fql_test.go
```diff
@@ -0,0 +1,272 @@
...
+		Protocol: "http",
+	}
+
+	r := rc.CheckFullQueryLogging()
+	if r != result.Continue() {
+		t.Fatalf("expected result of result.Continue() but got %s", r)
+	}
+}
+
+func TestCheckFullQueryNotSupportedTriedToUse(t *testing.T) {
```
Miles-Garnsey, line 251: +1.

Miles-Garnsey (commented):
Some tests still failing, but I like the new unit tests. Should we also test the code paths [here](https://github.com/k8ssandra/cass-operator/blob/c2832a3d74d297bf5f60880425856a4e2882ab53/pkg/reconciliation/reconcile_fql.go#L25), [here](https://github.com/k8ssandra/cass-operator/blob/c2832a3d74d297bf5f60880425856a4e2882ab53/pkg/reconciliation/reconcile_fql.go#L38) and [here](https://github.com/k8ssandra/cass-operator/blob/c2832a3d74d297bf5f60880425856a4e2882ab53/pkg/reconciliation/reconcile_fql.go#L45)?

As you mentioned, you've added more unit tests than I previously had, so I'd probably approve (once all checks pass) even if you don't want to add those additional tests. Just pointing them out in case we want to do so for completeness.

burmanm:
I think all of those are http connection issues, I don't really see the point in testing them as that's not really logic related. The only test really is that it requeues, but even that isn't such a big deal in the end.

burmanm:
The tests have started failing a lot after FQL reconcile addition in the master also. There's something going on, since this is the log in canary-upgrade (it does not enable FQL):

```
2021-11-04T01:29:12.646Z	ERROR	controllers.CassandraDatacenter	failed to call endpoint /api/v0/ops/node/fullquerylogging	{"cassandradatacenter": "test-canary-upgrade/dc1", "requestNamespace": "test-canary-upgrade", "requestName": "dc1", "loopID": "e87dbbe5-2da6-4c95-b38c-4ba9e55f056d", "namespace": "test-canary-upgrade", "datacenterName": "dc1", "clusterName": "cluster1", "error": "incorrect status code of 500 when calling endpoint"}
github.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).CheckFullQueryLogging
	/workspace/pkg/reconciliation/reconcile_fql.go:37
github.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).ReconcileAllRacks
	/workspace/pkg/reconciliation/reconcile_racks.go:2360
github.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).CalculateReconciliationActions
	/workspace/pkg/reconciliation/handler.go:160
github.com/k8ssandra/cass-operator/controllers/cassandra.(*CassandraDatacenterReconciler).Reconcile
	/workspace/controllers/cassandra/cassandradatacenter_controller.go:140
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:298
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:253
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:214
```

I'm surprised it tries to call these, as these are 3.11.7 pods.

burmanm:
Ah, now I know.. the pod reports it supports FQL properties (newer management-api) even for 3.11.*

Miles-Garnsey:
> Ah, now I know.. the pod reports it supports FQL properties (newer management-api) even for 3.11.*

So that's intended behaviour from management API because the server does not know which Cassandra version it is actually working against. It only talks to a generic `CassandraAPI` not a versioned `CassandraAPI4x`/`CassandraAPI3x`. I don't particularly like that fact, but that's how the project is structured, and it is major surgery to fix it.

This is why I was asking where we call ValidateFQL - doesn't that catch this condition in `handler.go` [here](https://github.com/k8ssandra/cass-operator/pull/216/files#diff-376efce46fc7fb911b13fb34f94409c8d0fd68e586036ee6e72c401cf935f4f1R207)? I was under the impression that we couldn't proceed to `CheckFullQueryLogging()` if the config wasn't valid, but I don't fully understand how the handler works?

burmanm:
We can proceed to CallFullQueryLogging. There's no enabled FQL or mention in any of the 3.11.* configs, but since CallFullQueryLogging wants to verify in any case, it tries to call the pod. As such, I added a check that if it's a dse/3.11 it just exits and doesn't verify the status.

Miles-Garnsey:
That sounds fine in theory... I just want to confirm that [this line](https://github.com/k8ssandra/cass-operator/pull/216/commits/a43052aa8520c57a53c6a6e1cb32f38a69de2c5a#diff-37da9c8b9cffdc407986a7dc0893e5aa0c68cd6bc0fa450d0c78ab7bb4d84ce8R14) is called after [this line](https://github.com/burmanm/cass-operator/blob/small_fql_change/pkg/reconciliation/handler.go#L195)?

I am struggling to get my head around this because you are validating in the DC controller but you are reconciling in the racks controller. I'm not sure about this 😒

burmanm:
Yes, the IsValid is in the beginning of the reconciling. 

The IsValid is called here:

https://github.com/k8ssandra/cass-operator/blob/65e47b5ac16adfccfbb335abf9723263554fc3ee/controllers/cassandra/cassandradatacenter_controller.go#L116

But the ReconcileAllRacks process starts only from:

https://github.com/k8ssandra/cass-operator/blob/65e47b5ac16adfccfbb335abf9723263554fc3ee/controllers/cassandra/cassandradatacenter_controller.go#L140

Which then leads to:

https://github.com/k8ssandra/cass-operator/blob/65e47b5ac16adfccfbb335abf9723263554fc3ee/pkg/reconciliation/handler.go#L160

Miles-Garnsey:
Righto. I don't love having the validation in a different place to the reconciliation. But I guess the logic is that we want to validate the resource as a whole and then proceed to call different reconcilers for its component parts, and I can understand that argument. 

I just need to remember that the reconciliation is occurring serially on a single resource, rather than multiple parallel reconciliation loops being run on component parts of the resource. 

Once the tests pass I'm happy to approve (again 😅)

Miles-Garnsey (approved):
Just one failing test. @burmanm advises he will fix separately and it is not related to this PR. Approved.

PR #221, opened 2021-10-28
Improve bundle support for automated Openshift push

burmanm:
**What this PR does**:
Tries to build the correct shippable bundle when doing `make bundle`.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


burmanm:
Tested deployment with OLM:

```console
➜  cass-operator git:(csv) ✗ operator-sdk run bundle docker.io/burmanm/cass-operator-bundle:latest
INFO[0010] Successfully created registry pod: docker-io-burmanm-cass-operator-bundle-latest 
INFO[0010] Created CatalogSource: cass-operator-catalog 
INFO[0010] OperatorGroup "operator-sdk-og" created      
INFO[0010] Created Subscription: cass-operator-v1-9-0-sub 
INFO[0015] Approved InstallPlan install-knqrs for the Subscription: cass-operator-v1-9-0-sub 
INFO[0015] Waiting for ClusterServiceVersion "default/cass-operator.v1.9.0" to reach 'Succeeded' phase 
INFO[0015]   Waiting for ClusterServiceVersion "default/cass-operator.v1.9.0" to appear 
INFO[0027]   Found ClusterServiceVersion "default/cass-operator.v1.9.0" phase: Pending 
INFO[0029]   Found ClusterServiceVersion "default/cass-operator.v1.9.0" phase: Installing 
INFO[0040]   Found ClusterServiceVersion "default/cass-operator.v1.9.0" phase: Succeeded 
INFO[0040] OLM has successfully installed "cass-operator.v1.9.0" 
➜  cass-operator git:(csv) ✗
```

config/manifests/bases/cass-operator.clusterserviceversion.yaml
```diff
@@ -16,33 +22,145 @@ spec:
...
       name: cassandradatacenters.cassandra.datastax.com
       version: v1beta1
-  description: '-'
-  displayName: DataStax Apache Cassandra Operator
+  description: ../../docs/developer/csv-description.md
+  displayName: DataStax Kubernetes Operator for Apache Cassandra
   icon:
-  - base64data: ""
-    mediatype: ""
+  - base64data: |
```
bradfordcp, line 28: Is there anyway we could store this in a separate file and include it similar to the `spec.description` field?

Bonus points if we don't have to store it in base64 format and can just read in the binary PNG.

.gitignore
```diff
@@ -253,3 +253,7 @@ tags
 # make temps
 bin/
 testbin/
+
+# bundle files+directories are generated
+bundle.Dockerfile
+bundle/
```
bradfordcp, line 259: I'm trying to understand the logic a bit here. Is the expectation that we don't want to check in the generated artifacts?

If so, is it safe to assume that we could regenerate these files at any time by just checking out the appropriate commit (or tag) and running the appropriate make target?

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
```diff
@@ -1,7 +1,29 @@
 apiVersion: cassandra.datastax.com/v1beta1
 kind: CassandraDatacenter
 metadata:
-  name: cassandradatacenter-sample
+  name: dc2
```
bradfordcp: When in the OpenShift OperatorHub catalog we need to include the image coordinates which are UBI based at a minimum. Preferably we would point to the Red Hat container registry coordinates.

In either case we should update this sample with the appropriate keys to override the images.

* serverImage
* configBuilderImage
* systemLoggerImage

Additionally when running on OpenShift we must override the SecurityContext with an empty block as it creates per pod users and injects them.

```yaml
spec:
  dockerImageRunsAsCassandra: false
  podTemplateSpec:
    securityContext: {}
    containers:
      - name: cassandra
        securityContext: {}
```

Note the `spec.dockerImageRunsAsCassandra` is set to false. This is due to needing to override the spec with an empty block so that OpenShift can template out the values it needs. See [construct_podtemplatespec.go](https://github.com/k8ssandra/cass-operator/blob/ed4a19bdd2cad356467872c899c47afb758581e8/pkg/reconciliation/construct_podtemplatespec.go#L575) and [construct_statefulset.go](https://github.com/k8ssandra/cass-operator/blob/ed4a19bdd2cad356467872c899c47afb758581e8/pkg/reconciliation/construct_statefulset.go#L57-L64) for a context.

This feels a bit counterintuitive though. We should consider changes needed to make this a little less confusing. (@jsanda has a little bit of experience here given our recent call with a user)

scripts/postprocess-bundle.sh
```diff
@@ -0,0 +1,24 @@
...
+yq eval -i '.annotations."com.redhat.openshift.versions" headComment = "Certified Openshift required labels"' bundle/metadata/annotations.yaml
+
+# This file is extra from creation process on config/manifests, should not be in the bundle itself
+rm -f bundle/field-config_v1_configmap.yaml
+
+# Use yq to set that date to the createdAt field
+createdAt=$(date +%Y-%m-%d) yq eval '.metadata.annotations.createdAt = env(createdAt)' -i bundle/manifests/cass-operator.clusterserviceversion.yaml 
+
+# Use the correct containerImage from the deployment
+yq eval '.metadata.annotations.containerImage = .spec.install.spec.deployments[0].spec.template.spec.containers[0].image' -i bundle/manifests/cass-operator.clusterserviceversion.yaml
```
bradfordcp: Minor: Missing newline at end of file.

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,454 @@
...
+      displayName: Access Modes
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:select:ReadWriteOnce
+    - path: storageConfig.cassandraDataVolumeClaimSpec.resources
+      description: |
+        Storage requirements for the Cassandra data volume. Only specify
+        the Storage column here.
+      displayName: Resources
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:resourceRequirements
```
bradfordcp: We should consider changing this definition out for something a bit more targeted. The issue being that the UI element rendered here while providing a field for specifying storage uses the `ephemeralStorage` key instead of `storage`. This would prevent users from solely creating a `CassandraDatacenter` via the OpenShift UI.

Consider:

```yaml
    - path: storageConfig.cassandraDataVolumeClaimSpec.resources.requests.storage
      description: |
        Storage requirements for the Cassandra data volume in GBs.
      displayName: Data volume size
      x-descriptors:
        - urn:alm:descriptor:com.tectonic.number
```

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
bradfordcp: It's worth noting that whatever sample we include in the `ClusterServiceVersion` is used to prepopulate the `CassandraDatacenter` creation UI. Providing the appropriate values here will default them for users even if we don't have a UI element specified.

bradfordcp: Should we choose a different name for this sample?

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
```diff
@@ -1,7 +1,29 @@
...
 spec:
-  # Add fields here
-  foo: bar
+  clusterName: cluster2
```
bradfordcp: Should we choose a different name for this sample?

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
```diff
@@ -1,7 +1,29 @@
...
+  serverType: cassandra
+  serverVersion: "3.11.11"
```
bradfordcp: Let's consider getting this updated to the 4.x line.

bradfordcp (changes requested):
Awesome work @burmanm! I've added a few questions / comments here and there.

Would you like the Red Hat registry repo and credentials added as secrets on the repo for automation purposes?

bradfordcp:
Our previous installation instructions included the creation of three resources before attempting to install the operator here. This included

1. Secret for TLS certificates
2. ValidatingWebhook configuration
3. webhook service

since these objects were not supported as part of the ClusterServiceVersion spec for describing the deployment of an operator. Given there are newer versions of Operator SDK is that still the case? Should we consider making any changes to our operator to support installation without the creation of these components?

.gitignore
burmanm: Yes, the intention is that make command generates all the correct files from scratch.

config/manifests/bases/cass-operator.clusterserviceversion.yaml
burmanm: That should be possible with postprocessing (and setting with yq), but I don't have the png file.

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
burmanm: The images section is something I'd still have to think about. I wanted this image to be deployable to OperatorHub also, in which case we probably want to fetch the images from hub.docker.com / quay.io. And then a certified version which would use the containerRegistry from Red Hat. I think the latter requires login with RH account to get the pull token?

burmanm: Yes, do you have a proposal? dc1 & cluster1? Or something fancier?

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
burmanm: Agreed.

burmanm:
None of those steps are necessary with OLM bundle format at least. The OLM creates certs (we remove the cert-manager mapped mounts in the OLM bundle creation) and attaches them to the container. So the steps I mentioned in the earlier comments (run bundle + kubectl apply) were all that I needed to get things running and a Cassandra deployed correctly.

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
burmanm: As for images, systemLoggerImage & configBuilderImage are always UBI based when cass-operator deploys them (1.8.0 and up). 

We can push modified ImageConfig for OLM that uses serverImages with UBI-ending, but do they exists? I don't see us building them in the management-api.

burmanm: I set them to dc1 & development for now..

config/manifests/patches/spec-descriptors.yaml
burmanm: I modified the storageConfig.additionalVolumes[0].resources.requests.storage also

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
burmanm: As for DSE-images. Is there any need ever to not use the non-UBI images? Or should we deploy by default the UBI-images (like datastax/dse-server:6.8.17-ubi7) ?

Makefile
```diff
@@ -38,9 +51,11 @@ M_INTEG_DIR ?= all
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 
 # Image URL to use all building/pushing image targets
-IMG ?= $(IMAGE_TAG_BASE):latest
+IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
+IMG_LATEST ?= $(IMAGE_TAG_BASE):latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+#CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
```
bradfordcp, line 57: Should we leave this line commented out since we're setting it explicitly in the next line?

config/manifests/bases/cass-operator.clusterserviceversion.yaml
bradfordcp: I decoded the base64 here and attached the resulting PNG

ds

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+        Exposed ports on the Kubernetes node level for forwarding services to Cassandra pods. Most deployments will not use this functionality.
+      displayName: Node Port Network Configuration
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+    - path: networking.nodePort.internode
+      description: |
+        Internode port to use when leveraging Node Port network configuration.
+      displayName: Internode Port
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
```
bradfordcp: Should this be `urn:alm:descriptor:com.tectonic.ui:number`?

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+    - path: networking.nodePort.internodeSSL
+      description: |
+        Internode SSL port to use when leveraging Node Port network configuration.
+      displayName: Internode SSL Port
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
```
bradfordcp: Let's use the `urn:alm:descriptor:com.tectonic.ui:number` descriptor instead of text.

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+    - path: networking.nodePort.native
+      description: |
+        Native transport port to use when leveraging Node Port network configuration.
+      displayName: Native Port
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
```
bradfordcp: `urn:alm:descriptor:com.tectonic.ui:number` instead of text

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+    - path: networking.nodePort.nativeSSL
+      description: |
+        Native transport SSL port to use when leveraging Node Port network configuration.
+      displayName: Native SSL Port
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
```
bradfordcp: `urn:alm:descriptor:com.tectonic.ui:number` instead of `text`

config/manifests/patches/spec-descriptors.yaml
bradfordcp:
```suggestion
        - urn:alm:descriptor:com.tectonic.ui:number
```

config/manifests/patches/spec-descriptors.yaml
bradfordcp:
```suggestion
        - urn:alm:descriptor:com.tectonic.ui:number
```

config/manifests/patches/spec-descriptors.yaml
bradfordcp:
```suggestion
        - urn:alm:descriptor:com.tectonic.ui:number
```

config/manifests/patches/spec-descriptors.yaml
bradfordcp:
```suggestion
        - urn:alm:descriptor:com.tectonic.ui:number
```

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+        Access mode of the volume
+      displayName: Access Modes
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:select:ReadWriteOnce
+    - path: storageConfig.cassandraDataVolumeClaimSpec.resources.requests.storage
+      description: |
+        Storage requirements for the Cassandra data volume.
+      displayName: Data volume size
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.number
```
bradfordcp: I think this has to be a text field given it is represented as a number and the unit.

```suggestion
    - path: storageConfig.cassandraDataVolumeClaimSpec.resources.requests.storage
      description: |
        Storage requirements for the Cassandra data volume. Ex: 2048 Gi
      displayName: Data volume size
      x-descriptors:
        - urn:alm:descriptor:com.tectonic.text
```

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+      description: |
+        Volume access modes
+      x-descriptors:
+        - 'urn:alm:descriptor:com.tectonic.ui:text'
+    - path: storageConfig.additionalVolumes[0].resources.requests.storage
+      description: |
+        Storage requirements for the additional data volume.
+      displayName: Data volume size
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.number
```
bradfordcp: Same as the Cassandra data volume, this should be a text field.

```suggestion
    - path: storageConfig.additionalVolumes[0].resources.requests.storage
      description: |
        Storage requirements for the additional data volume. Ex: 1 Gi
      displayName: Data volume size
      x-descriptors:
        - urn:alm:descriptor:com.tectonic.text
```

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+      displayName: Cluster Name
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
+    - path: size
+      description: Size
+      displayName: Size
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:podCount
+    - path: racks
+      description: "Collection of logical rack identifiers, these may be named \nand specify a zone to limit pod deployment.\n"
```
bradfordcp: I'm noticing a number of the `description` fields in this file are single line with `\n` characters. Is there any reason why we don't use the multi-line option like line `16` in this file?

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+    - path: disableSystemLoggerSidecar
+      description: |
+        Configuration for disabling the simple log tailing sidecar
+        container. Our default is to have it enabled.
+      displayName: Disable System Logger Sidecar
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:booleanSwitch
+    - path: dockerImageRunsAsCassandra
+      description: |
+        Does the Server Docker image run as the Cassandra user?
```
bradfordcp: This should include a bit more context, especially for the OpenShift UI.

```suggestion
        Indicates whether the Cassandra container runs as the built-in `cassandra` user. 
        On OpenShift clusters this field MUST be set to false given the random UID and GID
        assigned to each pod / container.
```

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+      description: |
+        Additional seeds for Cassandra
+      displayName: Additional Seeds
+      x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+    - path: additionalServiceConfig
+      description: |
+        AdditionalServiceConfig allows to define additional parameters
+        that are included in the created Services. Note, user can override
```
bradfordcp:
```suggestion
        that are included in the created Services. Note, users can override
```

config/manifests/patches/spec-descriptors.yaml
```diff
@@ -0,0 +1,452 @@
...
+        otherwise just a regular string.
+      x-descriptors:
+        - "urn:alm:descriptor:com.tectonic.ui:text"
+    - path: users
+      description: |
+        Cassandra users to bootstrap
+      displayName: Users
+    - path: users[0].secretName
+      description: |
+        Kubernetes secret containing user credentials
```
bradfordcp: Should we specify which keys are required within the Secret?

bradfordcp: Something like

```suggestion
        Kubernetes secret containing user credentials within the keys `username` and `password`.
```

bradfordcp: Note, I do not know if those are the correct keys.

config/manifests/patches/status-descriptors.yaml
```diff
@@ -0,0 +1,49 @@
...
+      displayName: Quiet Period
+    - path: conditions
+      description: conditions
+      displayName: Conditions
+    - path: message
+      description: message
+      displayName: Message
+    - path: reason
+      description: reason
+      displayName: Reason
```
bradfordcp: I think these keys may be removed. Looking at the status spec in the [CRD](https://github.com/k8ssandra/cass-operator/blob/master/config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml#L7383-L7446) these fields are under conditions and not separate.

What do you think?

bradfordcp:
```suggestion
```

bradfordcp (commented):
Looking good @burmanm. I've included a few tweaks with suggestions, let me know where you land on them.

Separately I see a number of GHA unit test warnings. Are those going to be tackled here or separately?

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
bradfordcp: We can use images hosted outside of the RH registry, they just need to be UBI-based. I thought we were going to mirror Docker Hub to quay.io via Harbor, but it's more likely we should just push images to both locations from the GitHub actions. I can take a look at that or we can put it in the backlog (CC @jdonenine). 

It would be nice to have UBI images for all containers we're pushing. Is that something we're working on @jdonenine? (I think this question focuses on the `cassandra` container).

bradfordcp: Great!

jdonenine: > I can take a look at that or we can put it in the backlog (CC @jdonenine).

I certainly seems simpler to do the push from the actions we already have, or some sort of a copy as a follow up action.  If the requirement is that they are UBI images and not where they live, we can probably deal with this later and we can put it in the backlog for now.

> It would be nice to have UBI images for all containers we're pushing. Is that something we're working on @jdonenine? (I think this question focuses on the cassandra container).

We generally just went through this with with the older stuff that we patched up, although we didn't have to touch the cassandra container at that point.  It makes sense I think that we based everything on the UBI images generally, it will simplify for us here and also help us avoid vulnerability issues down the road.

What new images do we need at this point?

bradfordcp: IMHO we should strive to **always** use the UBI images.

Edit: Specifically ones that are available from a public registry like Docker Hub or quay.io. When inside certain environments we can provide guidance around private registries.

config/manifests/patches/spec-descriptors.yaml
bradfordcp: Excellent, thanks!

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
burmanm: We don't have UBI images of Open source versions of Cassandra (the management-api doesn't build Cassandra itself)

config/manifests/patches/spec-descriptors.yaml
burmanm: No idea. This is from your PR, I don't know where these came from originally. I can modify them.

burmanm:
Those unit test warnings are not targeted for 1.9 (they're just some Github beta feature - we run staticcheck on PRs, but we do not enforce them at this point)

Makefile
burmanm: Yes, these are for older versions of controller-gen (only `crd` is what we'll use in newer ones)

bradfordcp:
I'm getting an error when running `make bundle`.

```console
~/s/cass-operator (csv|✚1…) $ make bundle
/home/bradfordcp/src/cass-operator/bin/controller-gen "crd" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
go: creating new go.mod: module tmp
Downloading sigs.k8s.io/kustomize/kustomize/v4@v4.4.0
go: downloading sigs.k8s.io/kustomize/kustomize/v4 v4.4.0
go: downloading sigs.k8s.io/kustomize/cmd/config v0.10.1
go: downloading sigs.k8s.io/kustomize/api v0.10.0
go: downloading sigs.k8s.io/kustomize/kyaml v0.12.0
go: downloading k8s.io/kube-openapi v0.0.0-20210421082810-95288971da7e
go get: added sigs.k8s.io/kustomize/kustomize/v4 v4.4.0
/home/bradfordcp/bin/operator-sdk generate kustomize manifests -q
WARN[0000] Skipping definitions parsing for API {}: Go type not found 
WARN[0000] Skipping definitions parsing for API {}: Go type not found 
cd config/manager && /home/bradfordcp/src/cass-operator/bin/kustomize edit set image controller=k8ssandra/cass-operator:1.9.0-dev.57f83b2-20211129
/home/bradfordcp/src/cass-operator/bin/kustomize build --load-restrictor LoadRestrictionsNone config/manifests | /home/bradfordcp/bin/operator-sdk generate bundle -q --overwrite --version 1.9.0-dev.57f83b2-20211129 --channels="stable" --default-channel="stable" 
INFO[0001] Creating bundle.Dockerfile                   
INFO[0001] Creating bundle/metadata/annotations.yaml    
INFO[0001] Bundle metadata generated suceessfully       
scripts/postprocess-bundle.sh
base64: invalid option -- 'b'
Try 'base64 --help' for more information.
/home/bradfordcp/bin/operator-sdk bundle validate ./bundle
INFO[0000] All validation tests have completed successfully 
```

Looking at my `base64` command I see the following information

```console
~/s/cass-operator (csv|✚2…) $ base64 --help
Usage: base64 [OPTION]... [FILE]
Base64 encode or decode FILE, or standard input, to standard output.

With no FILE, or when FILE is -, read standard input.

Mandatory arguments to long options are mandatory for short options too.
  -d, --decode          decode data
  -i, --ignore-garbage  when decoding, ignore non-alphabet characters
  -w, --wrap=COLS       wrap encoded lines after COLS character (default 76).
                          Use 0 to disable line wrapping

      --help     display this help and exit
      --version  output version information and exit

The data are encoded as described for the base64 alphabet in RFC 4648.
When decoding, the input may contain newlines in addition to the bytes of
the formal base64 alphabet.  Use --ignore-garbage to attempt to recover
from any other non-alphabet bytes in the encoded stream.

GNU coreutils online help: <https://www.gnu.org/software/coreutils/>
Full documentation <https://www.gnu.org/software/coreutils/base64>
or available locally via: info '(coreutils) base64 invocation'
~/s/cass-operator (csv|✚2…) $ base64 --version
base64 (GNU coreutils) 8.32
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Simon Josefsson.
```

I _think_ on my machine `-w` is the right flag instead of `-b`. What version of `base64` are you using?

bradfordcp:
After switching out that flag I'm met with a separate error message.

```console
...
INFO[0001] Bundle metadata generated suceessfully       
scripts/postprocess-bundle.sh
base64: config/manifests/images/ds.png: No such file or directory
/home/bradfordcp/bin/operator-sdk bundle validate ./bundle
INFO[0000] All validation tests have completed successfully 
```

which appears to be due to the PNG file being missing.

config/samples/cassandra.datastax.com_v1beta1_cassandradatacenter.yaml
bradfordcp: Right and that feels like a miss. It's not a blocker for this PR, but something we should consider. (@jdonenine we need to sync on this)

bradfordcp:
After bringing in the PNG file in my local checkout the bundle builds successfully. How do you want to move forward @burmanm?

burmanm:
Right, the -b is a BSD version it seems. Ouh well, I'll swap to -w

burmanm:
If there's otherwise nothing wrong, I'd merge and try to build an image to OperatorHub (from 1.9 branch with this merged to it)

bradfordcp (approved):
Excellent work!

PR #232, opened 2021-11-09
fix: check both secrets when reconciling InternodeCredentialSecret

### LINKED ISSUE #225: K8SSAND-1023 ⁃ Cassandra operator fails to start the Cassandra cluster due to a missing secret caused by unexpected crash
### Problem description
We find that the operator fails to create the Cassandra cluster if it crashes at some point and restarts.
More concretely, after restart the Cassandra Datacenter StatefulSet hangs at the init stage with the following error messages:

```java
Unable to attach or mount volumes: unmounted volumes=[encryption-cred-storage], unattached volumes=[encryption-cred-storage server-config default-token-gc5kq server-logs server-data]: timed out waiting for the condition with the reason secret "cassandra-datacenter-keystore" not found
```

From the error message, the `cassandra-datacenter-keystore` secret is missing.
After inspecting the source code, we find this is caused by a bug in `pkg/reconciliation/secrets.go`

```java
func (rc *ReconciliationContext) retrieveInternodeCredentialSecretOrCreateDefault() (*corev1.Secret, error) {
  secret, retrieveErr := rc.retrieveSecret(rc.keystoreCASecret())
  if retrieveErr != nil {  
    if errors.IsNotFound(retrieveErr) { // if cassandra-datacenter-ca-keystore doesn't exist
      createInternodeCACredential()  // create cassandra-datacenter-ca-keystore secret
      createCABootstrappingSecret()  // create cassandra-datacenter-keystore secret
    }
  } else {
    return
  }
```

Say the operator crashes right after creating the `cassandra-datacenter-ca-keystore` secret and before creating `cassandra-datacenter-keystore`. After restart, the operator finds `cassandra-datacenter-ca-keystore` already exists and will never try to create `cassandra-datacenter-keystore`. Thus, the missing `cassandra-datacenter-keystore` prevents the operator from setting up the Cassandra cluster.

### Potential solution
Check if both of the secrets exist in the if condition and create any missing one.
We are willing to issue a PR to help fix thie problem.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-1023) by [Unito](https://www.unito.io)
┆affectedVersions: cass-operator-1.7.1,cass-operator-1.8.0,k8ssandra-operator-1.0.0-alpha.1
┆friendlyId: K8SSAND-1023
┆priority: Low

srteam2020:
**What this PR does**:
- Check both CASecret and BootStrappingSecret when reconciling InternodeCredentialSecret
- Create both secrets if both of them are not present. Create CABootStrappingSecret based on CASecret if CASecret is present but CABootStrappingSecret is not present.

Fixes #225 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/secrets.go
+++ b/pkg/reconciliation/secrets.go
@@ -192,8 +192,13 @@ func (rc *ReconciliationContext) keystoreCASecret() types.NamespacedName {
 
 func (rc *ReconciliationContext) retrieveInternodeCredentialSecretOrCreateDefault() (*corev1.Secret, error) {
 	secret, retrieveErr := rc.retrieveSecret(rc.keystoreCASecret())
-	if retrieveErr != nil {
-		if errors.IsNotFound(retrieveErr) {
+	_, retrieveBootStrappingErr := rc.retrieveSecret(types.NamespacedName{
+		Name:      fmt.Sprintf("%s-keystore", rc.Datacenter.Name),
+		Namespace: rc.Datacenter.Namespace,
+	})
+	if retrieveErr != nil || retrieveBootStrappingErr != nil {
+		if errors.IsNotFound(retrieveErr) && errors.IsNotFound(retrieveBootStrappingErr) {
+			// both secrets are not found
 			secret, err := rc.createInternodeCACredential()
 
 			if err == nil && secret == nil {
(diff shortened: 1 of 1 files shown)


pkg/reconciliation/secrets.go
```diff
@@ -192,8 +192,13 @@ func (rc *ReconciliationContext) keystoreCASecret() types.NamespacedName {
 
 func (rc *ReconciliationContext) retrieveInternodeCredentialSecretOrCreateDefault() (*corev1.Secret, error) {
 	secret, retrieveErr := rc.retrieveSecret(rc.keystoreCASecret())
-	if retrieveErr != nil {
-		if errors.IsNotFound(retrieveErr) {
+	_, retrieveBootStrappingErr := rc.retrieveSecret(types.NamespacedName{
+		Name:      fmt.Sprintf("%s-keystore", rc.Datacenter.Name),
+		Namespace: rc.Datacenter.Namespace,
+	})
+	if retrieveErr != nil || retrieveBootStrappingErr != nil {
```
burmanm: This if clause jungle is a bit difficult to follow. Wouldn't it make more sense to check one by one and continue? Such as:

```go
secret, err := rc.retrieveSecret(rc.keystoreCASecret())
if err != nil {
  if  errors.IsNotFound(err) { // Handle here }
  return nil, err
}

// The same checks for retrieveBootstrapping
```

That would avoid the original problem. Also, the ``rc.createCABootstrappingSecret`` logic is now repeated in two different conditional clauses and that could cause issues in some future refactoring and it duplicates the code.

burmanm (changes requested):

pkg/reconciliation/secrets.go
srteam2020: Your suggestion makes more sense. I refactored the code a bit, can you check if it looks better to you?

burmanm (approved):

burmanm:
Thank you for the contribution.

PR #233, opened 2021-11-11
Allow configuration of default image repository and set system-logger when tagged

burmanm:
**What this PR does**:
To support efforts in rebuilding the UBI images, this PR adds the ability to set both a suffix for the version that's going to be deployed, as well as the default repository is configurable. Previously these were hardcoded in the images.go, but that doesn't work if OLM requires different images than what we usually use.

Also, when creating a tag, we now set the system-logger to that tag version also and configure the shipped image_config.yaml with it. This allows us to rebuild the UBI images if there's a CVE in the base image.

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


apis/config/v1beta1/imageconfig_types.go
```diff
@@ -55,6 +57,19 @@ type Images struct {
 	ConfigBuilder string `json:"config-builder"`
 }
 
+type DefaultImages struct {
```
jsanda, line 60: Do we want to also have default image components for system-logger and config-builder?

Should a tag be specified as part of the `ImageComponent.Repository` field?

CHANGELOG.md
```diff
@@ -14,7 +14,9 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 * [CHANGE] #213 Integration tests in Github Actions are now reusable across different workflows
 * [FEATURE] #193 Add new Management API endpoints to HTTP Helper: GetKeyspaceReplication, ListTables, CreateTable
 * [FEATURE] #175 Add FQL reconciliation via parseFQLFromConfig and SetFullQueryLogging called from ReconcileAllRacks. CallIsFullQueryLogEnabledEndpoint and CallSetFullQueryLog functions to httphelper.
+* [FEATURE] Allow overriding default Cassandra and DSE repositories and give versions a default suffix
```
jsanda: Need to add the PR number here

CHANGELOG.md
```diff
@@ -14,7 +14,9 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
...
 * [ENHANCEMENT] #185 Add more app.kubernetes.io labels to all the managed resources
+* [ENHANCEMENT] Simplify rebuilding the UBI images if vulnerability is found in the base images
```
jsanda: Need to add the PR number here

CHANGELOG.md
burmanm: That's not available when making the PR (these always need to be changed afterwards).

apis/config/v1beta1/imageconfig_types.go
burmanm: system-logger / config-builder are only available as ubi versions. Tag is not defined in the ImageComponent.Repository, this is only a struct to build using the ServerVersion.

If one wishes to override everything, then ServerImage (and eq) will not use any of these parameters.

jsanda (approved):

PR #234, opened 2021-11-15
Requeue should not happen instantly, fix upgrade_operator test

burmanm:
**What this PR does**:
Current timeouts on upgrade_operator tests are not enough for some machines so the reconcile function is still going when the timeout hits. This adds another timeout after the upgrade process.

Also, prevent cass-operator from instantly requeueing. This reduces chatting with the server, since now it would requeue after 1ms, which isn't enough time for anything to have changed (and thus, another requeue is called). Make 500ms the minimum time.

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/cassandra/cassandradatacenter_controller.go
+++ b/controllers/cassandra/cassandradatacenter_controller.go
@@ -142,6 +142,13 @@ func (r *CassandraDatacenterReconciler) Reconcile(ctx context.Context, request c
 		logger.Error(err, "calculateReconciliationActions returned an error")
 		rc.Recorder.Eventf(rc.Datacenter, "Warning", "ReconcileFailed", err.Error())
 	}
+
+	// Prevent immediate requeue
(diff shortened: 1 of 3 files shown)


controllers/cassandra/cassandradatacenter_controller.go
```diff
@@ -142,6 +142,13 @@ func (r *CassandraDatacenterReconciler) Reconcile(ctx context.Context, request c
 		logger.Error(err, "calculateReconciliationActions returned an error")
 		rc.Recorder.Eventf(rc.Datacenter, "Warning", "ReconcileFailed", err.Error())
 	}
+
+	// Prevent immediate requeue
+	if res.Requeue {
```
jsanda, line 147: This change seems simple enough but curious as to how exactly it helps.

jsanda (approved):

PR #236, opened 2021-11-17
Adding AdditionalLabels

### LINKED ISSUE #235: K8SSAND-1052 ⁃ No options to add labels on the statefulset object
**What happened?**
I trying to add labels to the statefulset because of our policy in our cluster. Does it possible currently with the operator?

I saw that we have `podTemplateSpec` and I added those. But seems to add it only on the pod not the statefulset object?

Add the possibility to add the labels on all resources created.

**Environment**

* Cass Operator version:

    * Kubernetes version information:docker.io/k8ssandra/cass-operator:v1.7.1

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1052) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1052
┆priority: Medium

piclemx:
**What this PR does**:
Adding the possibility to people to add `Labels` and  the all objects that will be created by the operator.

Fixes #235 

**Checklist**
- [ ] Changes manually tested


apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -212,6 +212,12 @@ type CassandraDatacenterSpec struct {
 
 	// Tolerations applied to the Cassandra pod. Note that these cannot be overridden with PodTemplateSpec.
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// Additional Labels allows to define additional labels that will be included in all object created by the operator. Not, user can override values set by default from the cass-operator and doing so could break cass-operator functionality.
+	AdditionalLabels map[string]string `json:"additionalLabels,omitempty"`
+
+	// Additional annotations allows to define additional annotations that will be included in all object created by the operator. Not, user can override values set by default from the cass-operator and doing so could break cass-operator functionality.
```
jsanda: `object` is missing an `s`, and `Not` is missing an `e` at the end.

pkg/oplabels/labels.go
```diff
@@ -27,6 +27,12 @@ func AddKubernetesLabels(m map[string]string, dc *api.CassandraDatacenter) {
 
 	instanceName := fmt.Sprintf("cassandra-%s", dc.Spec.ClusterName)
 	m[InstanceLabel] = instanceName
+
+	if len(dc.Spec.AdditionalLabels) != 0 {
```
jsanda, line 31: It feels a bit odd to add the labels here considering this function is specifically adding well-known kubernetes labels.

pkg/reconciliation/construct_service_test.go
```diff
@@ -74,3 +74,37 @@ func TestServiceNameGeneration(t *testing.T) {
 	service := newSeedServiceForCassandraDatacenter(dc)
 	assert.Equal(t, "notcool-bob-seed-service", service.Name)
 }
+
+func TestAddingAdditionalLabels(t *testing.T)  {
```
jsanda: It would be nice to see a test that demonstrates expected behavior when labels are added from both `AdditionalLabels` and `AdditionalServiceConfig`.

jsanda (changes requested):
You're off to a good start. There are some other objects that the operator like the superuser secret and endpoints for the additional seeds service. There might be more, but those are the ones I know of immediately.

Are you going to add corresponding functionality for annotations?

pkg/oplabels/labels.go
burmanm: No, technically this is where I would add them also. I'd make a small change though. Rename this function (refactoring is easiest way to get it everywhere) and then make those kubernetes labels as one step. I don't want to litter the code everywhere with billion different "add this label, add that label" when we could do them all here. It makes the code easier to read when resources are discussed.

Also, this method is called everywhere we currently create resources, so it hits the correct spots.

pkg/reconciliation/construct_service_test.go
piclemx: Yeah I would add one. I was thinking that I did one. Forgive me.

piclemx:
> You're off to a good start. There are some other objects that the operator like the superuser secret and endpoints for the additional seeds service. There might be more, but those are the ones I know of immediately.
> 
> Are you going to add corresponding functionality for annotations?

I think for now I will only add the `labels` for the first step. I can do the `annotations` later if you want? If you want both of them now I can also work on getting both.

pkg/oplabels/labels.go
piclemx: I agree with @burmanm, Will be easier to add other stuff and refactor the method if needed. Since the pattern, when creating an object is to call that method to put the labels.

I changed the name of the method to `AddOperatorLabels`.

piclemx:
I will be adding the missing test tomorrow.

pkg/reconciliation/construct_service_test.go
piclemx: Done

piclemx:
@jsanda @burmanm could you do another round of review?

For the annotation, do you mind I do that in another merge request?

piclemx:
@jsanda @burmanm do you still have sometime to review this one 🥺

jsanda:
Last week was a really short week for me due to the Thanksgiving holiday, and I just didn't have time to get back to it. I will do another review today or tomorrow.

jsanda (commented):
My apologies for the delayed review. The changes look good. There is another place you need to update for the default super user secret. See [here](http://github.com/k8ssandra/cass-operator/blob/master/pkg/reconciliation/secrets.go#L52-L52).

piclemx:
> My apologies for the delayed review. The changes look good. There is another place you need to update for the default super user secret. See [here](https://github.com/k8ssandra/cass-operator/blob/master/pkg/reconciliation/secrets.go#L52-L52).

Does it make sense to use the `AddOperatorLabels` there also?

You don't need to apologies @jsanda  ;)

jsanda:
>Does it make sense to use the `AddOperatorLabels` there also?

yeah, I think so

piclemx:
> > Does it make sense to use the `AddOperatorLabels` there also?
> 
> yeah, I think so

Done

jsanda (commented):
Thanks for adding the changes for the superuser secret. Pending the test run, I think it is good to go :)

piclemx:
Nice thank you ;)

jsanda (approved):
Thanks!

PR #243, opened 2021-12-07
Task controller for multi-pod jobs in cass-operator

burmanm:
**What this PR does**:
Runs jobs in serial in every pod part of the Datacenter, either using the async functions or sync functions (depending on the version of the cassandra container).

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -0,0 +1,187 @@
...
+
+package v1alpha1
+
+import (
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO Labels to add that indicate the status of the task? So they can be searched? Like jobs have.
```
jsanda: When you implemented the async cleanup I had suggested using labels instead of annotations precisely so that they can be searched. I'd still be in favor of that change :)

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,507 @@
...
+	SyncFunc     SyncTaskExecutorFunc
+	AsyncFeature httphelper.Feature
+	Arguments    map[string]string
+}
+
+//+kubebuilder:rbac:groups=control.k8ssandra.io,namespace=cass-operator,resources=cassandratasks,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=control.k8ssandra.io,namespace=cass-operator,resources=cassandratasks/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=control.k8ssandra.io,namespace=cass-operator,resources=cassandratasks/finalizers,verbs=update
+
+// Do we need to repeat this? It's already on the cassandradatacenter_controller.go
```
jsanda: Good question. My guess is no, but I think it is good to leave it because 1) it provides documentation and 2) it avoid an unnecessary coupling on cassandradatacenter_controller.go.

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,507 @@
...
+//+kubebuilder:rbac:groups=core,namespace=cass-operator,resources=pods;events,verbs=get;list;watch;create;update;patch;delete
+
+func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	var cassJob api.CassandraTask
+	if err := r.Get(ctx, req.NamespacedName, &cassJob); err != nil {
+		logger.Error(err, "unable to fetch CassandraTask", "Request", req.NamespacedName)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
jsanda, line 85: I wasn't aware of `client.IgnoreNotFound`. I like that.

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -0,0 +1,187 @@
...
+			concurrencyPolicy: Forbid
+			ttlSecondsAfterFinished: 86400
+	*/
+}
+
+type CassandraJob struct {
+	Name string `json:"name"`
+
+	// TODO Should this be typed to match what jobs this version of cass-operator can run?
+	Command string `json:"command"`
```
jsanda: 👍 to making `Command` more strongly typed.

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -0,0 +1,187 @@
...
+
+// TODO Labels to add that indicate the status of the task? So they can be searched? Like jobs have.
+
+// CassandraTaskSpec defines the desired state of CassandraTask
+type CassandraTaskSpec struct {
+
+	// ScheduledTime indicates the earliest possible time this task is executed. This does not necessarily
+	// equal to the time it is actually executed (if other tasks are blocking for example)
+	// +optional
+	ScheduledTime *metav1.Time `json:"scheduledTime,omitempty"`
```
jsanda, line 32: If `ScheduledTime` is `nil`, will the task be executed immediately, assuming no other tasks are already running? The comment should mention what the behavior is when it is `nil`.

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// TODO BackOffLimit for RetryPolicy and maxMissedSeconds deadline for ScheduledTime?
+
+	// Which datacenter this task is targetting. Note, this must be a datacenter which the current cass-operator
+	// can access
+	Datacenter corev1.ObjectReference `json:"datacenter,omitempty"`
```
jsanda, line 38: `Datacenter` is a required field, right? We have places in k8ssandra-operator that use references to a CassandraDatacenter,  although I don't think they are using ObjectReference. What do you think about making a CassandraDatacenterReference type?

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,507 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *CassandraTaskReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&api.CassandraTask{}).
+		Complete(r)
+}
+
+func (r *CassandraTaskReconciler) activeTasks(ctx context.Context, dc *cassapi.CassandraDatacenter) ([]api.CassandraTask, error) {
+	var taskList api.CassandraTaskList
+	// TODO This should add per-datacenter active (we can execute concurrently in different datacenters)
```
jsanda: Agreed that it should be per datacenter.

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// Jobs defines the jobs this task will execute (and their order)
+	Jobs []CassandraJob `json:"jobs,omitempty"`
```
jsanda, line 41: Curious, what is the use case for multiple jobs?

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,507 @@
...
+	if cassJob.GetLabels() == nil {
+		cassJob.Labels = make(map[string]string)
+	}
+
+	if _, found := cassJob.GetLabels()[taskStatusLabel]; found {
+		if res.RequeueAfter == 0 && !res.Requeue {
+			logger.Info("Setting this to be complete..\n")
+			// Job has been completed
+			cassJob.GetLabels()[taskStatusLabel] = completedTaskLabelValue
+			if err = r.Client.Update(ctx, &cassJob); err != nil {
```
jsanda: patch?

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,507 @@
...
+				return res, err
+			}
+
+			// This status update could be at the end also..
```
jsanda: 👍  to doing the status update at the end

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,507 @@
...
+			cassJob.Status.CompletionTime = &timeNow
+			if err = r.Client.Status().Update(ctx, &cassJob); err != nil {
+				return res, err
+			}
+		}
+	} else {
+		// Starting the run, set the Active label so we can quickly fetch the active ones
+		cassJob.GetLabels()[taskStatusLabel] = activeTaskLabelValue
+
+		if err = r.Client.Update(ctx, &cassJob); err != nil {
```
jsanda: patch?

apis/control/v1alpha1/cassandratask_types.go
burmanm: Well, this is actually implemented for the active status, not otherwise. The original behavior did not need it, this only because it's no longer tied to the CassandraDatacenter directly.

apis/control/v1alpha1/cassandratask_types.go
burmanm: Yes, it will be executed immediately. I can add that to the comment.

apis/control/v1alpha1/cassandratask_types.go
burmanm: What's the benefit over standard ObjectReference? One can set "Kind: CassandraDatacenter" in the ObjectReference also.

apis/control/v1alpha1/cassandratask_types.go
burmanm: One of those "what ifs". It doesn't really add any extra code to the processing itself (a single for clause). I guess you could do something like cleanup + repair?

apis/control/v1alpha1/cassandratask_types.go
jsanda: It's more explicit and self-documenting.

apis/control/v1alpha1/cassandratask_types.go
burmanm: Also, forgot.. this is task label, not pod annotations (those still remain annotations).

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -0,0 +1,188 @@
...
+	// from suspension. It is represented in RFC3339 form and is in UTC.
+	// +optional
+	StartTime *metav1.Time `json:"startTime,omitempty"`
+
+	// Represents time when the job was completed. It is not guaranteed to
+	// be set in happens-before order across separate operations.
+	// It is represented in RFC3339 form and is in UTC.
+	// The completion time is only set when the job finishes successfully.
+	// +optional
+	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
```
jsanda, line 107: >The completion time is only set when the job finishes successfully.

Is this correct? If I understand `reconcileEveryPodTask` correctly, it looks like the job should eventually reach a completion state even if it has failed.

burmanm: No, that text is copied over from Job. I don't know yet how to approach this, should it be "CompletionTime + Failed count" to notice that job failed or how would one check that it will no longer be executed (there are cases where we won't retry).

burmanm:
Trying to write down missing features:

- [x] TTL to delete expired tasks
- [x] Output modifications with kubebuilder annotations so that display looks nice with kubectl
- [ ] Condition updates to Status
- [ ] Partial failure mode and CompletionTime
- [x] Decommission support (requires changes elsewhere in the cass-operator)
- [ ] More than one job? (and its failure mode)
- [ ] Event support to track the process for some tools
- [ ] RetryPolicy extensions? MaxRetries and maxMissedSeconds for a scheduled time?
- [ ] ScheduledTimeWindow? Only start jobs between time n -> n and otherwise wait?
- [ ] Webhooks for validation
- [ ] CassandraTaskStatus has still fields from Job definition that will not be correct

None critical, but something I had planned..

controllers/control/cassandratask_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+var (
+	mockServer  *httptest.Server
+	callDetails *httphelper.CallDetails
+	// testNamespaceName  = ""
+	testDatacenterName = "dc1"
+)
+
+func createDatacenter(dcName, namespace string) func() {
+	return func() {
```
jsanda, line 30: Why does this need to be wrapped in a function?

pkg/httphelper/server_test_utils.go
```diff
@@ -0,0 +1,115 @@
...
+
+var featuresReply = `{
+	"cassandra_version": "4.0.1",
+	"features": [
+		"async_sstable_tasks",
+		"rebuild"
+	]
+	}`
+
+var jobDetailsCompleted = `{"submit_time":"1638545895255","end_time":"1638545895255","id":"%s","type":"Cleanup","status":"COMPLETED"}`
```
jsanda, line 20: When I first saw this I wondered whether there should be a separate job details for the rebuild, but looking again at `reconcileEveryPodTask` I see that it only checks the status. Is `type` actually used anywhere?

controllers/control/cassandratask_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+				},
+			},
+		},
+		Status: api.CassandraTaskStatus{},
+	}
+
+	return taskKey, task
+}
+
+var _ = Describe("Execute jobs against all pods", func() {
```
jsanda, line 166: Thanks for writing these tests. This is good stuff. There are probably more scenarios we could test, but they could make for good follow up PRs, particularly for someone else looking to get more familiar with the code.

pkg/httphelper/server_test_utils.go
burmanm: Not by cass-operator. But if one would have "getAllJobs" in the management-api, you could create external tool to monitor the jobs from pods.

controllers/control/cassandratask_controller_test.go
burmanm: Because of the way ginkgo works `Describe("...", func() ..)` and that's how it was previously used (until I refactored a bit to do it in the beforesuite instead of inside the test)

controllers/control/cassandratask_controller_test.go
burmanm: There are more scenarios, but writing them as envtest is difficult. Something like "prevent this from running concurrently" / "allow concurrently".

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,480 @@
...
+		features, err := nodeMgmtClient.FeatureSet(&pod)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if pod.Annotations == nil {
+			pod.Annotations = make(map[string]string)
+		}
+
+		if podJobId, found := pod.Annotations[podJobIdAnnotation]; found {
```
jsanda: I have been testing your branch with my k8ssandra-operator [branch](https://github.com/jsanda/k8ssandra-operator/tree/add-dc-dev) and hit some issues. I am testing adding a new DC to an existing cluster. I first tested a with a single-node DC. I grepped Cassandra's logs to confirm that the rebuild operation. Yay for that!

I then decided to try my test with 2 nodes. When the new DC, dc2, comes up in this case, cleanup runs first. When cleanup finishes, the job id and job status annotations are still present on the pods. The rebuild task is created and reported as finished but not rebuild operations are actually performed. This is because at line 354 we check for the job id annotation. It is already present from cleanup, so we assume that the job is done. 

Once I realized what was going on, I thought I might have a quick fix by simply adding a statement to delete the job id annotation at line 384. Unfortunately it leads to additional problems.

This situation made it abundantly clear that there needs to be an integration for this, i.e., two tasks run in succession.

Removing both the job id and job status annotations when the job completes won't work either because on a subsequent reconciliation we won't be able to determine if the job has already run on that pod. I think that storing the completed pods in the CassandraTask status would make this easier.

burmanm: I think I'll handle this the same way as Kubernetes handles affinity status in annotations (json-serialized job information).

jsanda: If we update TaskConfig to include the task name, then we might be able to add a check to see if both the job id and job status are set but belong to a different task other than the current one, then we would know we can proceed with execution. I can try to test this out today.

burmanm: Well, I did get it to run twice in the test at least with my approach. The current commit is a bit hacky, but I don't have the energy to make it pretty tonight - I'll continue tomorrow. But it seems to pass the test with two tasks in the same datacenter at least.

jsanda: Thanks for pushing the commit. I tested and it didn't resolve the issue.

burmanm: Hmm, I can't see them clashing at least on the annotations part:

```console
Annotations:  control.k8ssandra.io/job-4d7dc3a5-68c4-40cc-9136-cd9b3daac215:
                {"jobId":"448f3370-2aab-42de-8262-257911379814","jobStatus":"COMPLETED","jobHandler":"management-api"}
              control.k8ssandra.io/job-93af73b5-12aa-446e-9e12-720a85227e66:
                {"jobId":"964845be-a7dd-4c1c-977c-08988d8dbf1c","jobStatus":"COMPLETED","jobHandler":"management-api"}
```

burmanm: I also see that the cleanup endpoint is called multiple times. Hmm..

```console
INFO  [nioEventLoopGroup-2-1] 2021-12-13 10:04:59,643 Cli.java:617 - address=/10.244.1.2:35656 url=/api/v0/metadata/versions/features status=200 OK
INFO  [nioEventLoopGroup-2-2] 2021-12-13 10:04:59,646 Cli.java:617 - address=/10.244.1.2:35658 url=/api/v1/ops/keyspace/cleanup status=202 Accepted
INFO  [nioEventLoopGroup-2-1] 2021-12-13 10:04:59,670 Cli.java:617 - address=/10.244.1.2:35660 url=/api/v0/metadata/versions/features status=200 OK
INFO  [nioEventLoopGroup-2-2] 2021-12-13 10:04:59,672 Cli.java:617 - address=/10.244.1.2:35662 url=/api/v0/ops/executor/job status=200 OK
INFO  [nioEventLoopGroup-2-1] 2021-12-13 10:04:59,708 Cli.java:617 - address=/10.244.1.2:35664 url=/api/v0/metadata/versions/features status=200 OK
INFO  [nioEventLoopGroup-2-2] 2021-12-13 10:05:03,608 Cli.java:617 - address=/10.244.4.1:37512 url=/api/v0/probes/readiness status=200 OK
INFO  [nioEventLoopGroup-2-1] 2021-12-13 10:05:09,670 Cli.java:617 - address=/10.244.1.2:35666 url=/api/v0/metadata/versions/features status=200 OK
```

This was with live pods, third task to execute.

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,559 @@
...
+func (r *CassandraTaskReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&api.CassandraTask{}).
+		Complete(r)
+}
+
+func (r *CassandraTaskReconciler) activeTasks(ctx context.Context, dc *cassapi.CassandraDatacenter) ([]api.CassandraTask, error) {
+	var taskList api.CassandraTaskList
+	matcher := client.MatchingLabels(utils.MergeMap(dc.GetDatacenterLabels(), map[string]string{taskStatusLabel: activeTaskLabelValue}))
+	if err := r.Client.List(ctx, &taskList, client.InNamespace(dc.Namespace), matcher); err != nil {
```
jsanda, line 301: This always returns an empty result set because there are no datacenter labels on a CassandraTask. You need to remove the datacenter labels and then I would filter the results  based on the `.Spec.Datacenter` field.

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,559 @@
...
+				if details.Id == "" {
+					// This job was not found, pod most likely restarted. Let's retry..
+					delete(pod.Annotations, getJobAnnotationKey(taskId))
+					err = r.Client.Patch(ctx, &pod, podPatch)
+					if err != nil {
+						return ctrl.Result{}, err
+					}
+					return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
+				} else if details.Status == podJobError {
+					// Log the error, move on
```
jsanda, line 491: When the job finishes, either successfully or fails, the job annotation never gets removed from the pod so you end up with an accumulation of job annotations, e.g., 

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    control.k8ssandra.io/job-0f003348-33db-4272-bfcd-790ad27785cf: '{"id":"0fc51790-c522-4f74-9d87-a93a08d252af","status":"COMPLETED","handler":"management-api"}'
    control.k8ssandra.io/job-91f0112a-5cf6-4e65-b889-54d6676d2d8d: '{"id":"c38a54e5-5a57-4d2e-9067-7957370466cf","status":"COMPLETED","handler":"management-api"}'
    control.k8ssandra.io/job-b7946ffd-2456-4544-bab8-c6f314c6fcc4: '{"id":"8b04c4f2-632b-429d-b678-aeeefe5489a6","status":"COMPLETED","handler":"management-api"}'
  creationTimestamp: "2021-12-13T18:29:52Z"
  generateName: test-dc2-default-sts-
```
We should be removing those once it is safe to do so.

Since the annotation refers to the CassandraTask uid, do you think it might provide more clarity to change the prefix to `task`, e.g., `control.k8ssandra.io/task-b7946ffd-2456-4544-bab8-c6f314c6fcc4`?

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,559 @@
...
+		if err = r.Client.Patch(ctx, &cassTask, taskPatch); err != nil {
+			return res, err
+		}
+
+		cassTask.Status.StartTime = &timeNow
+		cassTask.Status.Active = 1 // We don't have concurrency inside a task at the moment
+	}
+
+	if cassTask.Status.Succeeded != completedCount {
+		cassTask.Status.Succeeded = completedCount
```
jsanda: I'm not sure I understand this `if` block, but maybe that's why `.Status.Succeeded` isn't getting updated properly. I am testing with 2 C* nodes and it has a value of 1 when the tasks finish. Maybe it should be `cassTask.Status.Succeeded += completedCount`.

jsanda (changes requested):

pkg/reconciliation/reconcile_racks.go
```diff
@@ -780,9 +780,9 @@ func (rc *ReconciliationContext) CheckRackScale() result.ReconcileResult {
 				updated = rc.setCondition(
 					api.NewDatacenterCondition(
 						api.DatacenterResuming, corev1.ConditionTrue)) || updated
-			} else {
+			} else if dc.GetConditionStatus(api.DatacenterReady) == corev1.ConditionTrue {
 				// We weren't resuming from a stopped state, so we must be growing the
-				// size of the rack
+				// size of the rack and this isn't the initialization stage
```
jsanda, line 785: Currently (in master) the scaling up condition does get set to true initially before the datacenter is ready.  Are there any scenarios in which the Ready condition could be false and scaling up would happen? I assume no but want to be sure. If it is possible would it be better/safer to handle this in the `CheckClearActionConditions` function. The function could be updated to only call `cleanupAfterScaling` after the Initialized condition is true.

burmanm: No, there's no such thing. This is the only place where ScalingUp is set (and I don't think it should be set when the cluster is starting).

controllers/control/cassandratask_controller.go
burmanm: It's not a pod count, it's a job count inside the task (increased after a job has finished).

controllers/control/cassandratask_controller.go
burmanm: I actually need to add those annotations (so that all resources have those annotations).

pkg/reconciliation/reconcile_racks.go
jsanda: My main point is that ScalingUp is set before the datacenter becomes ready, or more precisely, before the Initialized condition is set. What is the rationale for changing that existing behavior?

burmanm: Because creating a new cluster is not scaling up existing one. I don't feel like these two should share the same status - especially if we have a different process for both cases.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -394,6 +394,10 @@ type CassandraDatacenterStatus struct {
 
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// TrackedTasks tracks the tasks for completion that were created by the cass-operator
+	// +optional
+	TrackedTasks []corev1.ObjectReference `json:"trackedTasks,omitempty"`
```
jsanda, line 400: Since you have added the DC labels to the CassandraTask is this necessary any more? I see where it is used in `reconcile_racks.go`, but that could be replaced with a query to fetch the CassandraTask. There are a couple benefits. First, it eliminates and therefore simplifies some code. Secondly, it cuts down on requeues which is big. The status updates for a task will trigger a lot of unnecessary requeues not only for cass-operator but also for k8ssandra-operator.

controllers/control/cassandratask_controller.go
```diff
@@ -0,0 +1,593 @@
...
+		logger.V(1).Info("This task has completed, scheduling for deletion in " + nextRunTime.String())
+	}
+
+	if cassTask.Status.Succeeded != completedCount {
+		cassTask.Status.Succeeded = completedCount
+	}
+
+	// TODO Add conditions also
+	if err = r.Client.Status().Update(ctx, &cassTask); err != nil {
+		return ctrl.Result{}, err
```
jsanda, line 270: At one point I tested with a bad management-api image which resulted in an error when `taskConfig.AsyncFunc` is called. The failure was due to the RPC endpoint not being properly exposed. The errors happens against the first pod, subsequent pods are never tried. At some point the job should be marked failed. k8ssandra-operator ends up in a requeue loop waiting for the  job to complete, but it never does.

jsanda (changes requested):

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: Status updates should not trigger a reconcile. That's a bug in the k8ssandra-operator if that happens and means the watcher is missing version check.

jsanda: TBH I forgot about GenerationChangedPredicate when I wrote my comment. Status updates won't trigger a reconcile for cass-operator but it will for k8ssandra-operator since the latter does not use a GenerationChangedPredicate for its CassandraDatacenter watch. This is by design though. k8ssandra-operator doesn't use GenerationChangedPredicate since it does not manage the CassandraDatacenter. We do want status updates to trigger reconciliation in k8ssandra-operator. Aside from requeues (or lack thereof) I would still be in favor removing TrackedTasks since the information is already available through the CassandraTask.

burmanm: That information isn't actually available, since TrackedTasks only include tasks generated by the cass-operator, while fetching with labels would give back all the tasks generated by anyone. At that point, I would need to know which ones have been created before the operation I ran in the cass-operator and which later. I don't have that information at the moment.

controllers/control/cassandratask_controller.go
burmanm: It seems my reply never got here, odd.. well, the process catches failed jobs and exits with them correctly and cass-operator should kill pods that are not healthy. However, you hit an edge case where management-api is just simply not working correctly but cass-operator is not killing it.

jsanda (approved):

PR #250, opened 2021-12-23
Add support for decommissioning an entire datacenter [K8SSAND-1057][K8SSAND-619]

### LINKED ISSUE #125: K8SSAND-619 ⁃ No way to gracefully decommission an entire datacenter
**What happened?**

I was in the process of migrating an existing cassandra cluster to k8ssandra using the instructions described here: https://docs.k8ssandra.io/tasks/migrate/

After succesfully adding the k8ssandra datacenter to my existing cluster,  I wanted to abort the procedure and decomission the k8ssandra cluster again.   I tried scaling down the `cassandradatacenter` to size `0` however `cass-operator` threw an error saying that the size needs to be at least `1`. 

I then decided to delete the `cassandradatacenter` hoping that would gracefully decomission the datacenter. However instead it just deleted all the pods ands pvc at once and left all the nodes in a `DN` state.   Luckily I hadn't streamed any data yet between the datacenters and the k8sssandra datacenter was not in use yet, but it was a bit of an awkward bad state to end up in. I had to then manually `nodetool assasinate` each node that was in the k8ssandra datacenter from the old cassandra datacennter.

I expected it to be possible to set `size` to `0` so I can decomission an entire datacenter before deleting it from the cluster.

OR deleting a cassandradatacenter that is part of a larger cluster to gracefully decomission all nodes on deletion

Set size to `0` and see that it gets rejected AND delete the cassandradatacenter afterwards

**Environment**

* Cass Operator version:

    * Kubernetes cluster kind:```insert how you created your cluster: kops, bootkube, etc.```* Manifests:```insert manifests relevant to the issue```* Cass Operator Logs:```
```

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-619) by [Unito](https://www.unito.io)
┆epic: Decomission Datacenters
┆fixVersions: k8ssandra-operator-v1.0.0
┆friendlyId: K8SSAND-619
┆priority: Medium

### LINKED ISSUE #238: K8SSAND-1057 ⁃ Add support for automated decommissioning of datacenters
**What is missing?**
Support for the automated decommissioning of datacenters.  The management-api recently [added](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/161) support for triggering rebuild, this needs to be supported within cass-operator before it can be propagated up to support within k8ssandra-operator.

**Why do we need it?**
To support requirements upstream in K8ssandra Operator.

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-1057) by [Unito](https://www.unito.io)
┆Reviewer: John Sanda
┆epic: Decomission Datacenters
┆fixVersions: k8ssandra-operator-v1.0.0
┆friendlyId: K8SSAND-1057
┆priority: Medium

### PR DESCRIPTION
**Pull request by burmanm**
**What this PR does**:
In a multi-datacenter cluster, when deleting a CassandraDatacenter the cass-operator will run a decommission for all the remaining nodes in that datacenter. This ensures the cluster stays healthy and no ghost-nodes are left behind.

Still some rough edges (unit tests are probably failing, there's too many unnecessary logging lines when something really isn't going wrong and the decommission is using the old api/v0 method at this point only).

Fixes #125 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pkg/reconciliation/decommission_node.go
+++ b/pkg/reconciliation/decommission_node.go
@@ -79,7 +79,7 @@ func (rc *ReconciliationContext) DecommissionNodes(epData httphelper.CassMetadat
 		lastPodSuffix := stsLastPodSuffix(maxReplicas)
 
 		if maxReplicas > desiredNodeCount {
-			logger.V(1).Info(fmt.Sprintf("reconcile_racks::DecommissionNodes::maxReplicas::%d > %d", maxReplicas, desiredNodeCount), "Rack", rackInfo.RackName)
+			logger.V(1).Info("reconcile_racks::DecommissionNodes::scaleDownRack", "Rack", rackInfo.RackName, "maxReplicas", maxReplicas, "desiredNodeCount", desiredNodeCount)
 
 			dcPatch := client.MergeFrom(dc.DeepCopy())
 			updated := false
--- a/pkg/reconciliation/decommission_node.go
+++ b/pkg/reconciliation/decommission_node.go
@@ -62,7 +62,6 @@ func (rc *ReconciliationContext) DecommissionNodes(epData httphelper.CassMetadat
 	}
 
 	if currentSize <= targetSize {
-		logger.Info(fmt.Sprintf("reconcile_racks::DecommissionNodes::reached-target: %d <= %d", currentSize, dc.Spec.Size))
 		return result.Continue()
 	}
 
@@ -79,9 +78,9 @@ func (rc *ReconciliationContext) DecommissionNodes(epData httphelper.CassMetadat
 		maxReplicas := *statefulSet.Spec.Replicas
 		lastPodSuffix := stsLastPodSuffix(maxReplicas)
 
-		logger.Info(fmt.Sprintf("reconcile_racks::DecommissionNodes::maxReplicas::%d > %d", maxReplicas, desiredNodeCount))
-
 		if maxReplicas > desiredNodeCount {
+			logger.V(1).Info(fmt.Sprintf("reconcile_racks::DecommissionNodes::maxReplicas::%d > %d", maxReplicas, desiredNodeCount), "Rack", rackInfo.RackName)
+
 			dcPatch := client.MergeFrom(dc.DeepCopy())
 			updated := false
 
@@ -120,8 +119,6 @@ func (rc *ReconciliationContext) DecommissionNodes(epData httphelper.CassMetadat
 		}
 	}
 
-	logger.Info(fmt.Sprintf("reconcile_racks::DecommissionNodes::continue-after-decommrack: %d <= %d", currentSize, dc.Spec.Size))
-
 	return result.Continue()
 }
 
@@ -142,7 +139,7 @@ func (rc *ReconciliationContext) DecommissionNodeOnRack(rackName string, epData 
 				return err
 			}
 
-			rc.ReqLogger.Info("Marking node as decommissioning")
+			rc.ReqLogger.V(1).Info("Marking node as decommissioning")
 			patch := client.MergeFrom(pod.DeepCopy())
 			metav1.SetMetaDataLabel(&pod.ObjectMeta, api.CassNodeState, stateDecommissioning)
 			if err := rc.Client.Patch(rc.Ctx, pod, patch); err != nil {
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Comment by burmanm** [comment]
@jsanda One additional thing, do we want to protect this feature behind an annotation? So that we wouldn't even try to decommission/multi-dc check if ``k8ssandra.io/decommission-on-need`` (just example name) isn't set for the DC?

File: pkg/reconciliation/reconcile_datacenter.go [hunk 1]
```diff
@@ -27,6 +28,47 @@ func (rc *ReconciliationContext) ProcessDeletion() result.ReconcileResult {
...
+
+	if rc.Datacenter.Status.GetConditionStatus(api.DatacenterScalingDown) == corev1.ConditionTrue {
+		// ScalingDown is still happening
+		rc.Recorder.Eventf(rc.Datacenter, corev1.EventTypeNormal, events.DecommissionDatacenter, "Datacenter is decommissioning")
+		rc.ReqLogger.V(1).Info("Waiting for the decommission to complete first, before deleting")
+		return result.Continue()
+	}
+
+	// If there are still pods up, check the cluster datacenter count..
+	if _, found := rc.Datacenter.Annotations[api.SkipDecommissionAnnotation]; !found {
```
jsanda: With the skip decommission annotation the user can opt out. Why not require the user to opt in? If you do that, then I don't see why we would need to bother with calling `rc.getClusterDatacenters(dcPods)`. k8ssandra-operator knows if it is a multi-dc cluster. This would avoid extra calls to management-api/Cassandra. And if someone is using cass-operator on its own and wants to decommission when there is only a single dc, would there be any harm?

pkg/reconciliation/decommission_node.go
```diff
@@ -54,7 +58,7 @@ func (rc *ReconciliationContext) DecommissionNodes(epData httphelper.CassMetadat
 		}
 	}
 
-	if currentSize <= dc.Spec.Size {
```
jsanda, line 57: I did a manual test with 2 DCs and 2 nodes per DC. When I deleted dc2, one node is decommissioned, but no more.  The pod is eventually recreated since the liveness check will fail and because the replica count on the statefulset isn't getting changed.

I haven't looked into it enough but it seems like it would be tricky to scale the StatefulSets down without also changing the size of the CassandraDatacenter. Definitely not a fan of that.

I am inclined to think that this might be easier to do from k8ssandra-operator. It would simply reduce the size of the dc, requeue until the decommission completes, reduce the size again, and so on until it hits zero. Other than allowing the last node to be decommissioned, would there be other changes required in cass-operator?

jsanda (changes requested):

pkg/reconciliation/reconcile_datacenter.go
burmanm: We did discuss this option in the chat, opt-in vs opt-out. I'm fine with the opt-in also.

pkg/reconciliation/decommission_node.go
burmanm: After the decommission has succeeded, it will scale down StatefulSet. If your pod came up again it failed to decommission for some reason. Also, the CassandraDatacenter size is reduced during the operation.

Allowing user to set size to 0 would be bad, since it will never succeed. One can't call decommission to the last node if there are no other DCs. It will simply never finish.

burmanm: The scale down process itself was not changed, if it failed to reduce StS size then that issue is affecting all scale down operations

jsanda: The decommission did succeed. Then it looks like we requeue. The replica count of the StatefulSet doesn't change.

>Allowing user to set size to 0 would be bad, since it will never succeed. One can't call decommission to the last node if there are no other DCs. It will simply never finish.

But it's ok if we have another DC, right?

burmanm: If cass-operator sees decommission succeeding, it should remove the pod from StS:

https://github.com/k8ssandra/cass-operator/blob/master/pkg/reconciliation/decommission_node.go#L171

There's no requeue before doing this once it sees decommission has succeeded. Apparently it has missed the success for some reason. 

> But it's ok if we have another DC, right?

Assuming one rewrites the current rack scaling process, yes. The current rack scaling (up or down) is unable to function with size 0. It isn't simply a validation issue.

burmanm: Okay, I found the bug.

pkg/reconciliation/decommission_node.go
```diff
@@ -155,23 +186,25 @@ func (rc *ReconciliationContext) CheckDecommissioningNodes(epData httphelper.Cas
...
-					rc.ReqLogger.Info("Node decommissioning, reconciling again soon")
 				}
 			} else {
-				rc.ReqLogger.Info("Node finished decommissioning")
+				rc.ReqLogger.V(1).Info("Node finished decommissioning", "Pod", pod.Name)
 				if res := rc.cleanUpAfterDecommissionedPod(pod); res != nil {
 					return res
 				}
 			}
+			// TODO Add event here to indicate decommissioning this node is still taking place?
```
jsanda, line 213: I wouldn't. We could get spammy with events. I'd say just check the logs. If Cassandra exposes any sort of metrics on the progress of the decomm, then it would be really nice to expose those (but not suggesting anything like that for this PR).

pkg/reconciliation/reconcile_datacenter.go
```diff
@@ -27,6 +30,59 @@ func (rc *ReconciliationContext) ProcessDeletion() result.ReconcileResult {
 		return result.Error(err)
 	}
 
+	origSize := rc.Datacenter.Spec.Size
+	if rc.Datacenter.Status.GetConditionStatus(api.DatacenterDecommission) == corev1.ConditionTrue {
+		rc.Datacenter.Spec.Size = 0
```
jsanda, line 33: How come you set `.Spec.Size` here? Is it necessary?

jsanda (approved):
I have a question about setting `.Spec.Size = 0`; otherwise, this looks good to go.

pkg/reconciliation/reconcile_datacenter.go
burmanm: This code is called before actual decommission:

https://github.com/burmanm/cass-operator/blob/decommission_dc/pkg/reconciliation/reconcile_racks.go#L51

So setting it here ensures that the CalculateRackInformation has Size set to 0

PR #305, opened 2022-03-24
Set the ServiceName for new StatefulSets

### LINKED ISSUE #304: K8SSAND-1349 ⁃ Hostname lookups on Cassandra pods fail
**What happened?**
I created a CassandraDatacenter with cluster name of test and no racks and 3 C* nodes. This results in the following pods:

* test-dc1-default-sts-0
* test-dc1-default-sts-1
* test-dc1-default-sts-2

As per the Kubernetes [docs](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id) there should be a DNS record for each pod such that the following should be resolvable (assuming the CassandraDatacenter is created in cass-operator namespace),

`test-dc1-default-sts-0.test-dc1-all-pods-service.cass-operator.svc.cluster.local`

Note that `test-dc1-all-pods-service` is the name of the headless service with which cass-operator configures the StatefulSet. Specifically, cass-operator sets the `StatefulSet.Spec.ServiceName` field to the name of the all-pods service.

The DNS lookup fails because the `StatefulSet.Spec.ServiceName` is set to the empty string.

The `StatefulSet.Spec.ServiceName` property should be set to the all-pods service name then hostname lookups will work.

Create a CassandraDatacenter, e.g., 

```yaml
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: test
  config:
    jvm-server-options:
      initial_heap_size: 512M
      max_heap_size: 512M
  serverType: cassandra
  serverVersion: 4.0.3
  size: 1
  storageConfig:
    cassandraDataVolumeClaimSpec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
      storageClassName: standard
```

Wait for the operator to create the StatefulSet and then see that the `ServiceName` property is not set.

**Environment**

* Cass Operator version:

```
I believe this affects versions 1.7.1 to 1.10.1
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1349) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1349
┆priority: Medium

jsanda:
**What this PR does**:

Fixes #304

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 * [CHANGE] [#264](https://github.com/k8ssandra/cass-operator/issues/264) Generate PodTemplateSpec in CassandraDatacenter with metadata
 * [ENHANCEMENT] []() Update to Go 1.17 with updates to dependencies: Kube 1.23.4 and controller-runtime 0.11.1 
 * [BUGFIX] [#298](https://github.com/k8ssandra/cass-operator/issues/298) EndpointState has incorrect json key 
+* [BUGFIX] [#304](https://github.com/k8ssandra/cass-operator/issues/304) Hostname lookups on Cassandra pods fail
 
 ## v1.10.1
 
--- a/pkg/reconciliation/construct_statefulset.go
+++ b/pkg/reconciliation/construct_statefulset.go
@@ -179,7 +179,7 @@ func newStatefulSetForCassandraDatacenter(
 	}
 	result.Annotations = map[string]string{}
 
-	if sts != nil && sts.Spec.ServiceName != result.Spec.ServiceName {
+	if sts != nil && sts.Spec.ServiceName != "" && sts.Spec.ServiceName != result.Spec.ServiceName {
 		result.Spec.ServiceName = sts.Spec.ServiceName
 	}
 
--- a/pkg/reconciliation/construct_statefulset_test.go
+++ b/pkg/reconciliation/construct_statefulset_test.go
@@ -6,6 +6,8 @@ package reconciliation
 import (
 	"fmt"
 	"github.com/k8ssandra/cass-operator/pkg/oplabels"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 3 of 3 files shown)


pkg/reconciliation/construct_statefulset.go
```diff
@@ -179,7 +179,7 @@ func newStatefulSetForCassandraDatacenter(
 	}
 	result.Annotations = map[string]string{}
 
-	if sts != nil && sts.Spec.ServiceName != result.Spec.ServiceName {
+	if sts != nil && sts.Spec.ServiceName != "" && sts.Spec.ServiceName != result.Spec.ServiceName {
```
burmanm, line 182: This would indicate that the dc.GetAllPodsServiceName() is returning empty string (set in the line 174)

burmanm: Wonder what would've created the empty sts originally

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 * [BUGFIX] [#272](https://github.com/k8ssandra/cass-operator/issues/272) Strip password (if it has one) from CreateRole
 * [BUGFIX] [#254](https://github.com/k8ssandra/cass-operator/pull/254) Safely set annotation on datacenter in config secret
 * [BUGFIX] [#261](https://github.com/k8ssandra/cass-operator/issues/261) State of decommission can't be detected correctly if the RPC address is removed from endpoint state during decommission, but before it's finalized.
+* [BUGFIX] [#304](https://github.com/k8ssandra/cass-operator/issues/304) Hostname lookups on Cassandra pods fail
```
burmanm: This is in wrong place, it should be placed under ## unreleased

pkg/reconciliation/construct_statefulset.go
jsanda: Turns out we always start with an empty sts initially. Due to this bug for new CassandraDatacenters the `ServiceName` will always be empty string.

`dc.GetAllPodsServiceName()` does not return empty string.

burmanm (approved):

PR #309, opened 2022-03-31
Recreate StatefulSet if required for updates to StS

### LINKED ISSUE #304: K8SSAND-1349 ⁃ Hostname lookups on Cassandra pods fail

### LINKED ISSUE #311: K8SSAND-1410 ⁃ ScalingUp Status is not always removed
**What happened?**
While running scale_up integration test, I've noticed that the second / third cleanup task is not always created as it seems the DatacenterController thinks there's still an active one (and as such, no new one is required). There are few indicators why this could be happening:

```
1.6488026926732535e+09	ERROR	controller.cassandratask	Reconciler error	{"reconciler group": "control.k8ssandra.io", "reconciler kind": "CassandraTask", "name": "cleanup-1648802692", "namespace": "test-scale-up", "error": "Operation cannot be fulfilled on cassandratasks.control.k8ssandra.io \"cleanup-1648802692\": the object has been modified; please apply your changes to the latest version and try again"}
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:266
```

That task did complete (the controller was able to handle it), but it indicates we have some concurrent update / missing refresh. The next one completed without issues, but on this run, the third one was never created - but there's no logging to indicate why this might have happened.

Sadly, we don't have CassandraDatacenter status logged in the test framework, so tracking why that might have happened is difficult. Tracking this in #312 

From further debugging, it seems to be that CassandraDatacenter stays in "ScalingUp" and updating and that's why CassandraTask is never created:

```yaml
  - lastTransitionTime: "2022-04-01T09:59:05Z"
    message: ""
    reason: ""
    status: "True"
    type: ScalingUp
  - lastTransitionTime: "2022-04-01T10:01:58Z"
    message: ""
    reason: ""
    status: "True"
    type: Updating
```

Yet there's no action going on in the cass-operator, since:

```
1.6488073383268092e+09  INFO    controllers.CassandraDatacenter.cassandradatacenter_controller.controller.cassandradatacenter-controller        All StatefulSets should now be reconciled.{"reconciler group": "cassandra.datastax.com", "reconciler kind": "CassandraDatacenter", "name": "dc2", "namespace": "test-scale-up", "namespace": "test-scale-up", "datacenterName": "dc2", "clusterName": "cluster2"}
```

**Environment**

* Cass Operator version:

    * Kubernetes cluster kind:```
```

* Manifests:

```
```

* Cass Operator Logs:

```
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1410) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1410
┆priority: Medium

### LINKED ISSUE #312: K8SSAND-1409 ⁃ Test dump does not output all objects
**What is missing?**
After running the test, the build directory has information such as operator logs, pod logs and some objects. However, we miss all our CRD statuses and that makes it difficult to track last status of an object, or which objects were actually created and how they looked.

**Why do we need it?**
To improve debug of test runs.

**Environment**

* Cass Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1409) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1409
┆priority: Medium

### PR DESCRIPTION
**Pull request by burmanm**
**What this PR does**:

Fixes #312
Fixes #311 
Fixes #304 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,12 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 * [CHANGE] [#183](https://github.com/k8ssandra/cass-operator/issues/183) Move from PodDisruptionBudget v1beta1 to v1 (min. Kubernetes version 1.21)
 * [CHANGE] [#264](https://github.com/k8ssandra/cass-operator/issues/264) Generate PodTemplateSpec in CassandraDatacenter with metadata
-* [ENHANCEMENT] []() Update to Go 1.17 with updates to dependencies: Kube 1.23.4 and controller-runtime 0.11.1 
+* [FEATURE] [#309](https://github.com/k8ssandra/cass-operator/pull/309) If StatefulSets are modified in a way that they can't be updated directly, recreate them with new specs
+* [ENHANCEMENT] [#292](https://github.com/k8ssandra/cass-operator/issues/292) Update to Go 1.17 with updates to dependencies: Kube 1.23.4 and controller-runtime 0.11.1 
+* [ENHANCEMENT] [#312](https://github.com/k8ssandra/cass-operator/issues/312) Integration tests now output CassandraDatacenter and CassandraTask CRD outputs to build directory
 * [BUGFIX] [#298](https://github.com/k8ssandra/cass-operator/issues/298) EndpointState has incorrect json key 
 * [BUGFIX] [#304](https://github.com/k8ssandra/cass-operator/issues/304) Hostname lookups on Cassandra pods fail
+* [BUGFIX] [#311](https://github.com/k8ssandra/cass-operator/issues/311) Fix cleanup retry reconcile bug
 
 ## v1.10.1
 
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -164,17 +164,13 @@ func (rc *ReconciliationContext) CheckRackCreation() result.ReconcileResult {
 func (rc *ReconciliationContext) desiredStatefulSetForExistingStatefulSet(sts *appsv1.StatefulSet, rackName string) (desiredSts *appsv1.StatefulSet, err error) {
 	dc := rc.Datacenter
 
-	// have to use zero here, because each statefulset is created with no replicas
-	// in GetStatefulSetForRack()
-	replicas := 0
-
 	// when Cass Operator was released, we accidentally used the incorrect managed-by
 	// label of "cassandra-operator" we have since fixed this to be "cass-operator",
 	// but unfortunately, we cannot modify the labels in the volumeClaimTemplates of a
 	// StatefulSet. Consequently, we must preserve the old labels in this case.
 	usesDefunct := usesDefunctPvcManagedByLabel(sts)
 
-	return newStatefulSetForCassandraDatacenter(sts, rackName, dc, replicas, usesDefunct)
+	return newStatefulSetForCassandraDatacenter(sts, rackName, dc, int(*sts.Spec.Replicas), usesDefunct)
 }
 
 func (rc *ReconciliationContext) CheckRackPodTemplate() result.ReconcileResult {
@@ -275,11 +271,14 @@ func (rc *ReconciliationContext) CheckRackPodTemplate() result.ReconcileResult {
 			statefulSet.SetResourceVersion(resVersion)
 			err = rc.Client.Update(rc.Ctx, statefulSet)
 			if err != nil {
-				logger.Error(
-					err,
-					"Unable to perform update on statefulset for config",
-					"statefulSet", statefulSet)
-				return result.Error(err)
+				if errors.IsInvalid(err) {
+					if err = rc.deleteStatefulSet(statefulSet); err != nil {
+						// logger.Error(err, "Failed to delete the StatefulSet", "Invalid", errors.IsInvalid(err), "Forbidden", errors.IsForbidden(err))
+						return result.Error(err)
+					}
+				} else {
+					return result.Error(err)
+				}
 			}
 
 			if err := rc.enableQuietPeriod(20); err != nil {
@@ -334,7 +333,7 @@ func (rc *ReconciliationContext) CheckRackForceUpgrade() result.ReconcileResult 
 		return result.Continue()
 	}
 
-	for idx := range rc.desiredRackInformation {
+	for idx, nextRack := range rc.desiredRackInformation {
 		rackName := rc.desiredRackInformation[idx].RackName
 		if slice.ContainsString(forceRacks, rackName, nil) {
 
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: pkg/reconciliation/reconcile_racks.go [hunk 1]
```diff
@@ -408,6 +409,37 @@ func (rc *ReconciliationContext) CheckRackForceUpgrade() result.ReconcileResult
 	return result.Done()
 }
 
+func (rc *ReconciliationContext) deleteStatefulSet(statefulSet *appsv1.StatefulSet) error {
+	// Remove reference to the StatefulSet from pods
```
jsanda: Can you remove this comment block?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2222,7 +2254,7 @@ func (rc *ReconciliationContext) CheckClearActionConditions() result.ReconcileRe
 	// Explicitly handle scaling up here because we want to run a cleanup afterwards
 	if dc.GetConditionStatus(api.DatacenterScalingUp) == corev1.ConditionTrue {
 		// Call the first node with cleanup, wait until it has finished and then move on to the next pod..
-		if res := rc.cleanupAfterScaling(); !res.Completed() {
+		if res := rc.cleanupAfterScaling(); res.Completed() {
```
jsanda, line 2238: Can you explain this? I don't follow. I read this as, if the cleanup task is done then return and do not flip the status on the ScalingUp condition.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -275,11 +271,14 @@ func (rc *ReconciliationContext) CheckRackPodTemplate() result.ReconcileResult {
...
 			err = rc.Client.Update(rc.Ctx, statefulSet)
 			if err != nil {
-				logger.Error(
-					err,
-					"Unable to perform update on statefulset for config",
-					"statefulSet", statefulSet)
-				return result.Error(err)
+				if errors.IsInvalid(err) {
+					if err = rc.deleteStatefulSet(statefulSet); err != nil {
+						// logger.Error(err, "Failed to delete the StatefulSet", "Invalid", errors.IsInvalid(err), "Forbidden", errors.IsForbidden(err))
```
jsanda: Can you remove this commented out line?

pkg/reconciliation/reconcile_racks.go
burmanm: No, that's what I thought it would do also when I wrote it originally. That's not what Completed() in result is however. Completed() in this case means the processing should not continue. ``res.Completed()`` is true when a) Error happens or b) Requeue is requested. If the cleanup is done, the cleanupAfterScaling returns ``result.Continue()`` which actually has ``res.Completed()`` returning false.

jsanda: Thanks for the explanation 👍

tests/util/kubectl/kubectl.go
```diff
@@ -351,6 +351,43 @@ func DumpLogs(path string, namespace string) KCmd {
 	return KCmd{Command: "cluster-info", Args: args, Flags: flags}
 }
 
+// DumpClusterInfo Executes `kubectl cluster-info dump -o yaml` on each cluster. The output
+// is stored under <project-root>/build/test.
+func DumpClusterInfo(path string, namespace string) error {
```
jsanda, line 356: It would be nice to put the output under a test-specific subdirectory. I realize though that it would require passing the test name through several functions in the ginkgo package. If there is a desire to do that it could be done in a later PR.

CHANGELOG.md
```diff
@@ -13,9 +13,12 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 * [CHANGE] [#183](https://github.com/k8ssandra/cass-operator/issues/183) Move from PodDisruptionBudget v1beta1 to v1 (min. Kubernetes version 1.21)
 * [CHANGE] [#264](https://github.com/k8ssandra/cass-operator/issues/264) Generate PodTemplateSpec in CassandraDatacenter with metadata
-* [ENHANCEMENT] []() Update to Go 1.17 with updates to dependencies: Kube 1.23.4 and controller-runtime 0.11.1 
+* [FEATURE] [#309](https://github.com/k8ssandra/cass-operator/pull/309) If StatefulSets are modified in a way that they can't be updated directly, recreate them with new specs
```
jsanda, line 16: Do you really consider this a feature? :)

tests/util/kubectl/kubectl.go
burmanm: It is done already (the path is test specific with date and test name)

CHANGELOG.md
burmanm: Yes. It's not available in Kubernetes ;)

jsanda: Curious, outside of the bug fix what is the use case here that would be considered a feature?

jsanda (approved):
I created a CassandraDatacenter using 1.10.1. Then I updated to the latest in your branch and verified that the `ServiceName` property of the StatefulSet was updated as expected.

CHANGELOG.md
burmanm: We've modified the StatefulSet before, but previously it has been protected by not making the changes if it was something else previously. I'd say this is the intended behavior of StatefulSets as of now. Recreating it is more a workaround for not being able to modify it and as such I don't think it's a bugfix.

PR #322, opened 2022-04-19
Add missing requeue to Datacenter decommission

### LINKED ISSUE #323: K8SSAND-1465 ⁃ Datacenter decommission is missing a requeue to prevent early deletion
**What happened?**
When cass-operator is deployed in a multicluster environment and annotation is set for decommission of the dc on delete (`cassandra.datastax.com/decommission-on-delete`), we first decommission the nodes and then delete them. However, even after finishing the decommission, the node isn't instantly deleted and so our decommission happens too early. Instead, we need to wait until all the pods are deleted and then do the decommission.

Sadly, this doesn't happen with the last pod as it there was missing requeue and so the datacenter was deleted after the decommission had finished.

**Environment**

* Cass Operator version:

    * Kubernetes cluster kind:```
```

* Manifests:

```
```

* Cass Operator Logs:

```
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1465) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1465
┆priority: Medium

### PR DESCRIPTION
**Pull request by burmanm**
**What this PR does**:
Decommissioning datacenter is missing a requeue after scaling down is done, but the pods haven't been removed yet. This can cause the operator to remove PVCs before Cassandra has properly shutdown (it has 30s wait after decommission).

Fixes #323

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_datacenter.go
+++ b/pkg/reconciliation/reconcile_datacenter.go
@@ -70,6 +70,8 @@ func (rc *ReconciliationContext) ProcessDeletion() result.ReconcileResult {
 				// Exiting to let other parts of the process take care of the decommission
 				return result.Continue()
 			}
+			// How could we have pods if we've decommissioned everything?
+			return result.RequeueSoon(5)
 		}
 	}
 
--- a/.github/workflows/kindIntegTest.yml
+++ b/.github/workflows/kindIntegTest.yml
@@ -60,46 +60,46 @@ jobs:
       matrix:
         integration_test:
         # Single worker tests:
-        - additional_serviceoptions
-        - additional_volumes
-        # - delete_node_terminated_container # This does not test any operator behavior
-        - podspec_simple
-        # - smoke_test_oss # See next job
-        - smoke_test_dse
-        # - terminate # Completes too fast, the test doesn't catch it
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by jsanda** [comment]
@burmanm can you please create an issue for this?

File: pkg/reconciliation/reconcile_datacenter.go [hunk 1]
```diff
@@ -70,6 +70,8 @@ func (rc *ReconciliationContext) ProcessDeletion() result.ReconcileResult {
 				// Exiting to let other parts of the process take care of the decommission
 				return result.Continue()
 			}
+			// How could we have pods if we've decommissioned everything?
+			return result.RequeueSoon(5)
```
jsanda, line 74: Are you adding the requeue here to handle the scenario where `len(dcs) == 1`?

burmanm: Why would we requeue in that case? If len(dcs) == 1, we go and delete the DC.

jsanda: It's not obvious to me why the need for the requeue here. Can you explain?

jsanda (approved):

Issue #192, opened 2021-10-07
[Discuss] Simplify CassDC services


Miles-Garnsey:
**What is missing?**

The current services in the cassDC are slightly confusing to users at times. We have `CLUSTERNAME-DCNAME-all-pods-service` as well as a `CLUSTERNAME-DCNAME-dc1-service` which differ only in that `all-pods-service` has additional ports and also exposes pods which are not yet passing readiness probes. 

We also have the `additional-seeds-service` which is empty for all single DC deployments (most deployments).

**Why do we need it?**

Services are usually subject to monitoring and having excess services clutters the operator experience. For developers, it is currently confusing which service should be used to talk to the Cassandra API on 9042. While experienced users will be able to infer this from which ports are open, it makes things less intuitive.

* Proposed solution**

This is just a suggestion, there may be additional factors to consider given commentary in [k8ssandra-operator issue #67](https://github.com/k8ssandra/k8ssandra-operator/issues/67).

I suggest we rename:

`CLUSTERNAME-DCNAME-service` -> `CLUSTERNAME-DCNAME-client-apis`
{{CLUSTERNAME-DCNAME-all-pods-service }} -> `CLUSTERNAME-DCNAME-internal-monitoring`

(No need for the type of the resource at the end of the name as this is obvious when inspecting it.)

I also suggest we find a way to avoid creating the `additional-seeds-service` unless additional seeds are actually required. Testing will be needed to ensure we continue to avoid an additional STS rolling restart - which is not desirable.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-55) by [Unito](https://www.unito.io)
┆Issue Number: CASS-55

burmanm:
This would break **ALL** existing cluster installations of cass-operator. Not cass-operator installation itself, but all those applications that connect to the Cassandra. The service names can be mentioned in the documentation on how to connect to Cassandra inside Kubernetes, but renaming them makes very little sense. I don't actually remember tickets about anyone running into issues due to service naming.

Should such rename occur, it should happen in a version where we have to break all applications in any case. I don't foresee such version at this point in time however.

Miles-Garnsey:
> This would break ALL existing cluster installations of cass-operator.

So firstly, let's separate the topic of renaming the services from the topic of not instantiating the `additional-seeds` service, the latter doesn't break anything. 

[...]

PR #328, opened 2022-05-06
CDC Integration [K8SSAND-1426]

### LINKED ISSUE #317: K8SSAND-1426 ⁃ Add fields to allow CDC configuration in CassandraDatacenter
We need to add fields to control CDC configuration in the CassandraDatacenter resource. 

These should look like

```
cdc:
  enabled: true
  pulsar_url: "..."
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1426) by [Unito](https://www.unito.io)
┆epic: CDC integration
┆friendlyId: K8SSAND-1426
┆priority: Medium

### LINKED ISSUE #336: K8SSAND-1527 ⁃ Diagnose potential incompatibility with CDC agent
**What happened?**

According to [this FAQ](https://docs.datastax.com/en/cdc-for-cassandra/cdc-apache-cassandra/1.0.3/faqs.html) the DataStax CDC agent should interoperate with cass-operator. However, when I tried to reproduce the integration test demonstrating the compatibility, I wasn't able to successfully get events to feed into a Pulsar deployment.

Initially we had issues in two areas:

1. The management API needs to be added to the classpath so that the custom SeedProvider class we use (K8SeedProvider) is available to the CDC agent.
1. The CDC agent needs to be started after the management API, or management API doesn't start.

While we were able to resolve those two problems, we still aren't seeing events output to the Pulsar topic.

When I check the logs, I can see:

1. The CDC agent starts
1. Management API returns 200s in response to liveness/readiness probes.
1. CDC is enabled on the table created in Cassandra.
1. The data is being flushed (using `nodetool flush`) and then cassandra-stress is being used to flood the commitlogs.

These conditions should lead to the cdc agent picking up the flushed commitlogs and sending them to the Pulsar connector, but we aren't seeing the events appear in Pulsar.

We are currently waiting on the Pulsar team to help us troubleshoot. This ticket to track any further work as it comes through.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1527) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1527
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Adds CDC types to the `CassandraDatacenter` API. 

Adds a function that takes json data (formatted as required by the `CONFIG_FILE_DATA` environment variable in the `server-config-init` container) and manipulates it to add additional CDC related fields to the `additional-jvm-opts` within the `jvm-options` key.

Integrates these functions into the reconciliation process, such that the Cassandra config, classpath and command line arguments all support the CDC agent to start.

Fixes #317 

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
I think this PR now fulfils the immediate requirements of the ticket. I'm not happy with the following areas:

1. The (un)marshalling logic in `pkg/cdc/serde.go` is verbose and seems like overkill to meet my needs. But I can't think of a better way to handle the fact that we only want to deal with a very particular path of the json (`jvm-options.additional-jvm-opts.${cdc-option}`) but want to preserve all other fields as well.
2. The way that the `pkg/cdc/` functions hook into `pkg/reconciliation` is untidy IMO. Ideally, this function would sit in `GetConfigAsJSON()`, but because `GetConfigAsJSON()` is a method of `CassandraDatacenter`, this is impossible as it would cause a circular import.
3. The tests could probably use a refactor to be more table-based, as is, there is a fair bit of duplicated code in them.

I'm going to think about how to resolve (1) and (2). Maybe it would be best to refactor `GetConfigAsJSON()` so that it isn't a method of the `CassandraDatacenter()` and sits in its own package under `pkg`.

It still needs an e2e test to ensure integration with Pulsar works. I feel like that might be best in a wholly separate PR as it'll be large.

Miles-Garnsey:
For the sake of keeping the ticket updated, I'm just going to summarise what's been going on for the past week. E2E testing brought up a few areas where changes had to be made.

1. I've added makefile targets for new test infra required (kuttl, Helm) and written a test which gets as far as installing Pulsar, cass-operator, cert-manager, configuring the Pulsar CDC source (that was tough) and spinning up a CassandraDatacenter with CDC enabled.
2. I've added logic to set `cdc_enabled` in the cassandra.yaml (this was previously missing).
3. It became clear that using the `*.options` fields to set `additional-jvm-opts` wasn't going to be a good approach because the key needs to change depending on Cassandra version. Updated this logic to use `cassandra-env-sh` instead which is consistent between versions (although this is not quite the intended usage of `cassandra-env.sh`).
4. It became clear that I'd stuffed up the image build and didn't have the CDC agent included. While trying to rebuild it became clear that our management API images fail to build unless they point at the most recent Cassandra version. This was fixed under mgmt API PR [#192](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/192) and we no longer fail if we aren't pointing at the most recent version.

With all of that completed, I'm now seeing an error out of the CDC agent as below:

```
org.apache.cassandra.exceptions.ConfigurationException: org.apache.cassandra.locator.K8SeedProvider
Fatal configuration error; unable to start server.  See log for stacktrace.
	at org.apache.cassandra.config.DatabaseDescriptor.applySeedProvider(DatabaseDescriptor.java:1030)
	at org.apache.cassandra.config.DatabaseDescriptor.applyAll(DatabaseDescriptor.java:360)
	at org.apache.cassandra.config.DatabaseDescriptor.daemonInitialization(DatabaseDescriptor.java:178)
	at org.apache.cassandra.config.DatabaseDescriptor.daemonInitialization(DatabaseDescriptor.java:162)
	at com.datastax.oss.cdc.agent.Agent.main(Agent.java:49)
	at com.datastax.oss.cdc.agent.Agent.premain(Agent.java:31)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(Unknown Source)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(Unknown Source)
```

Which appears to be a complaint about our custom seed provider. I am looking into this to see if it is a misconfiguration on my end (wouldn't be the first).

Additionally, I think I will need to make further changes to add `cdc_raw_directory` into `cassandra.yaml` as well, since it looks like the agent fails to start if it is not defined.

So still quite some work to be done here, and a new and unexpected risk cropping up in the form of a potential seed provider incompatibility.

Miles-Garnsey:
More updates:

* I found that adding `CLASSPATH=$CLASSPATH:/opt/management-api/datastax-mgmtapi-agent-0.1.0-SNAPSHOT.jar` to `cassandra-env.sh` does resolve the issues with the CDC agent not finding the SeedProvider.
* As a result I raised a [PR](https://github.com/datastax/cass-config-definitions/pull/80) on cass-config-builder to allow us to pass that parameter down from `spec.config.cassandra-env-sh.additional-env-opts` - a new field within the json.
* Cassandra now starts and so does the CDC agent. 
* We now see an issue presenting on stderr.log as below. I am not convinced this is a fatal problem given these classes appear to be identical. It isn't great though.

It is probably desirable to separate the SeedProvider in management API into a jar separate from the agent to avoid these kinds of issues.

We also need to now add the entry in `additional-env-opts` as part of the reconciliation.

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/opt/cassandra/lib/logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/opt/management-api/datastax-mgmtapi-agent-0.1.0-SNAPSHOT.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
```

Miles-Garnsey:
I've confirmed that the SLF4J error is not a problem with for either the management API (via @emerkle826) or Cassandra (via folks internally).

However, there is now a new problem where - although Cassandra starts without error - management API doesn't ever return a ready status on `/api/v0/probes/readiness`.

**Management API log**

```
INFO  [nioEventLoopGroup-2-2] 2022-05-20 00:23:35,786 Cli.java:617 - address=/10.244.1.1:39918 url=/api/v0/probes/readiness status=500 Internal Server Error
```

**Cassandra log excerpt**
```
INFO  [main] 2022-05-20 00:10:34,782 StorageService.java:2769 - Node /10.244.1.27:7000 state jump to NORMAL
...
INFO  [main] 2022-05-20 00:10:42,969 CassandraDaemon.java:780 - Startup complete
```

Miles-Garnsey:
Updates:

* We determined that starting the CDC agent before management API was blocking management API from starting, hence the 500 errors.
* We're now starting the management API first, and this appears to lead to management API coming up successfully.
* I now have the connector coming up, and everything appears OK from the logs I can see. However I still see no output in the Pulsar topic. I'm asking the Pulsar team for some more assistance to help diagnose.
* There is no logging from the CDC agent after the message `Agent.java:92 - CDC agent started`, which makes me wonder if the MAAC agent is now stopping the CDC agent.

Miles-Garnsey:
Update:

We encountered a number of problems trying to get the CDC agent working but have now demonstrated that it is definitely compatible with the management API.

I am now writing a test utility to coordinate the addition of data to Cassandra, configuring the connector etc. checking the data arrives on the Pulsar topic etc. The bash script we have does demonstrate this functionality, but a separate test container is preferable because:

* The shell based test relies on a custom Cassandra image which has Pulsar built in.
* There is a lot of backwards and forwards into different Pulsar/Cassandra containers to make this work on the shell.
* I also need to gain familiarity with the Java/Scala libraries in the ecosystem and demonstrate that end of the equation works in order to write docs and blogs etc.

Miles-Garnsey:
This PR is ready for review. The new E2E kuttl test will not pass until PRs from the following repos are merged:

1. [cass-config-definitions](https://github.com/datastax/cass-config-definitions/pull/80).
2. [management api](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/185)

emerkle826:
> 1. [cass-config-definitions](https://github.com/datastax/cass-config-definitions/pull/80).

Are the cass-config-definitions changes still necessary? Those changes are what adds extra jarfiles to the classpath of the JVM that runs Cassandra, correct?

Miles-Garnsey:
Updates:

* I've made a few changes to bump the CDC agent version, fix the testutils arguments (which were based on another cluster I was working on).
* I've squashed a bunch of commits.
* Tests are passing locally and remotely when I test with my local management API and config builder images.

Based on feedback from Erik we determined that we may not need the CLASSPATH changes, so I'm now:
* Updating again to remove the CLASSPATH logic and will re-test to see if everything still works. **(Done, tests successful, committed.)**
* Merging Erik's management API changes into my branch and re-testing again. **(Done, tests successful, committed.)**

Miles-Garnsey:
@burmanm I think we're about to release management API  v0.1.42. Are there any fields I need to update in this repo to use that image by default? Or does it always just default to the `latest` based on Cassandra version?

Miles-Garnsey:
I think I've encountered a problem here while manually testing my blog post steps. It looks like the pods don't bounce when the CassDC resource is edited, which means CDC remains disabled/enabled in the cass-config-builder's CONFIG_DATA env variable despite having been changed on the resource.

I'll do some more investigation to figure out how we can force the cluster to bounce when this field is updated.

Miles-Garnsey:
This is all working now and the test utility has been overhauled to accurately report passes/fails in all circumstances. It does still require multiple retries to return a success at the moment, which is down to some sort of race condition/timing error in the Pulsar connector. 

Either a delay is needed between schema creation and connector creation, or a delay is needed between connector creation and data insertion (into Cassandra).

I'm passing that back over to the Pulsar team to see if we can eliminate the need for retries, but the current state is as good as we'll get right now.

apis/cassandra/v1beta1/cdc_types.go
```diff
@@ -0,0 +1,50 @@
+package v1beta1
+
+// CDCConfiguration holds CDC config for the CassandraDatacenter. Note that it cannot contain arrays, channels, maps etc. because of the way the
+// reflection logic works which marshalls it into a string for the purposes of passing on the command line.
+type CDCConfiguration struct {
+	Enabled          bool    `json:"enabled"`
```
burmanm: Why do we need separate enabled flag when CDCConfiguration is a pointer?

apis/cassandra/v1beta1/cdc_types.go
```diff
@@ -0,0 +1,50 @@
...
+	// +optional
+	PulsarMaxPendingMessages *int `json:"pulsarMaxPendingMessages,omitempty"`
+	// +optional
+	PulsarMaxPendingMessagesAcrossPartitions *int `json:"pulsarMaxPendingMessagesAcrossPartitions,omitempty"`
+	// +optional
+	PulsarAuthPluginClassName *string `json:"pulsarAuthPluginClassName,omitempty"`
+	// +optional
+	PulsarAuthParams *string `json:"pulsarAuthParams,omitempty"`
+	// +optional
+	SSLProvider *string `json:"sslProvider,omitempty"`
```
burmanm, line 31: Instead of flat structure, would it be easier for the user if there was some sort of tree?

config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml
```diff
@@ -140,6 +140,61 @@ spec:
                   size, then all nodes in the rack will get updated.
                 format: int32
                 type: integer
+              cdc:
+                description: CDC allows configuration of the change data capture agent
```
burmanm, line 144: We should probably mention here that this isn't usable for generic CDC usage, only Datastax Pulsar integration.

pkg/cdc/const.go
```diff
@@ -0,0 +1,3 @@
+package cdc
+
+var CDCAgentVer = "2.0.0"
```
burmanm: Lets not put versions to .go files, it requires the user to deploy a version of cass-operator whenever we need to change this version.

pkg/cdc/generate_config.go
```diff
@@ -0,0 +1,143 @@
...
+	"fmt"
+	"reflect"
+	"strings"
+
+	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
+)
+
+// UpdateConfig updates the json formatted Cassandra config which incorporates the JVM options (under key additional-jvm-opts) passed into the launch scripts
+// for Cassandra via the config builder.
+func UpdateConfig(config json.RawMessage, cassDC cassdcapi.CassandraDatacenter) (json.RawMessage, error) {
```
burmanm, line 15: I can't follow the logic of doing it this way. What was the main reason to not do these in the existing methods such as GetConfigASJson() ?

pkg/cdc/validation.go
```diff
@@ -0,0 +1,19 @@
...
+
+// Validate validates that the cdcConfig is valid.
+func Validate(cdcConfig *cassdcapi.CDCConfiguration) bool {
+	if cdcConfig == nil {
+		return true
+	}
+	switch {
+	case !cdcConfig.Enabled:
+		return true
+	case cdcConfig.Enabled && cdcConfig.PulsarServiceUrl == nil:
```
burmanm: Shouldn't OpenAPI validation take care of this? PulsarServiceURL minLength > 0

pkg/test/testobjects.go
```diff
@@ -0,0 +1,21 @@
+package test
```
burmanm: Unnecessary package, these are only used in the cdc package.

burmanm (changes requested):
This is missing the e2e tests from the tests directory and the kuttl stuff should be removed as kuttl is not used in this project.

apis/cassandra/v1beta1/cdc_types.go
Miles-Garnsey: That's a valid point. In k8ssandra-operator, I'd allow setting this `false` because a cluster level setting might need to be overridden for individual DCs. 

But in this case, cass-operator operates only at DC level, so I'm happy to make this change if you'd like?

apis/cassandra/v1beta1/cdc_types.go
Miles-Garnsey: Can you expand on this? What struct do you envisage to encapsulate `PulsarAuthParams` and `SSLProvider`?

config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml
Miles-Garnsey: Pulsar isn't limited a DataStax proprietary product... I imagine in future we'll see connectors which might allow output to Kafka et al. via some "Pulsar on Kafka" type technology.

In saying that, if you provide me your preferred phrasing I'm probably happy to document as you like.

pkg/cdc/const.go
Miles-Garnsey: This affects the path to the change agent .jar file so it is kind of required... What's your alternative? Would you change management API image builds to use the same path for CDC agent regardless of version?

I could be persuaded, but given that I think CDC agent is still a bit unstable, there are some benefits to tightly coupling the CDC agent version to cass-operator. My concern is in the following case:

* Change agent exposes an "API" via Java CLI arguments. 
* cass-operator embeds that API [here](https://github.com/Miles-Garnsey/cass-operator/blob/feature/cdc-integration/apis/cassandra/v1beta1/cdc_types.go#L5).
* If the change agent's API evolves, cass-operator will pass the wrong arguments via the reflection process [here](https://github.com/Miles-Garnsey/cass-operator/blob/feature/cdc-integration/pkg/cdc/generate_config.go#L57).

I think given where the maturity level of the technology is at, and given we don't understand the "API" evolution strategy for change agent (i.e. what arguments it will accept), I'd rather play safe and couple tightly.

pkg/cdc/generate_config.go
Miles-Garnsey: I don't quite understand the question here. `GetConfigAsJson()` looks like it would be a very appropriate place to hook this logic in, but we wouldn't want to put all of the reconciliation logic in that function would we? It seems nicer to have it segregated off in `pkg`.

Can you clarify what you're suggesting we change?

pkg/cdc/validation.go
Miles-Garnsey: Now that is a very good idea! Thanks.

pkg/test/testobjects.go
Miles-Garnsey: I can move this into the CDC package if you prefer. My logic in separating it is that hopefully it will be re-used elsewhere. I think all of our projects should have a `pkg/test` package which holds functions that can produce minimal examples of the top level (CR-linked) objects they produce.

Otherwise we end up in a situation where there are many `GetCassandraDatacenter` functions all over the place, and they produce subtly different `CassandraDatacenter` objects and are harder to maintain.

apis/cassandra/v1beta1/cdc_types.go
burmanm: For example, all SSL properties (I assume they're CDC -> target) to one struct, all Pulsar properties to one struct. If you envision that there's going to more targets than Pulsar, then next there could be "kafka" part and so on. Instead of one flat structure with hundreds of options.

pkg/cdc/const.go
burmanm: There's multiple issues in that approach though. Lets say there's a bug in 2.0.0. What will we ship in the management-api images? Are all those previous Cassandra versions now unable to update management-api because we can't anymore process them? Lets say 4.0.4 mgmt-api CDC has a bug. We can't update that 4.0.4 image, it has to ship with 2.0.0 forever - otherwise you will kill existing installations if they restart a pod. 

We can't ship 4.0.5 with 2.0.1 either, since now their cass-operator installation can no longer update their Cassandra to another minor version, so all of 4.0.* has to stay in 2.0.0 forever. And probably 4.1 also at this point. 

And if we update here to 2.0.1, that version of cass-operator can no longer create any new 4.0.x or 4.1.x installations.

And then it gets worse if we actually have to ship multiple Cassandra images. :4.0.4-2.0.1 tag and manually forcing the image for every CDC? User upgrades cass-operator, but that cass-operator might require an update to Kubernetes also. Thus, their 2.0.0 -> 2.0.1 might require both Kubernetes as well as Cassandra major version changes.

pkg/test/testobjects.go
burmanm: The problem is that there's already such mess in cass-operator, with each package having its own generator methods for unit tests (usually for a reason since each one creates a different kind and Golang is pretty crappy for these types of things where simple function overload would look nicer). If we want to later refactor these, then I wish that would happen in a separate PR.

pkg/cdc/generate_config.go
burmanm: Just creating the JSON config. Why would it do reconcile logic otherwise? It's just taking whatever has been set in the CassandraDatacenter and creating a JSON output of configuration - it does not care about any logic otherwise. 

It would reduce quite a lot of the logic, since there's no need to read any JSON - everything is still in a struct form there.

pkg/cdc/const.go
emerkle826: > This affects the path to the change agent .jar file so it is kind of required... What's your alternative? Would you change management API image builds to use the same path for CDC agent regardless of version?

@Miles-Garnsey We may want to alter the Management API image so that it removes the version number from the CDC agent jarfile so that the version can change and cass-operator doesn't have to know or care about specific versions. If you decide you want to do this, we can cut a new Management API release pretty quickly. Talking about [this line](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/Dockerfile-4_0#L40) in the Dockerfile.

apis/cassandra/v1beta1/cdc_types.go
Miles-Garnsey: There aren't currently Kafka output capabilities in the ecosystem. I'd be open to a second nested struct for the Pulsar agent as below which would enable later additions with fewer API changes, if you wanted. 

```
type CDCConfiguration struct {
        pulsar: PulsarCDCConfiguration `...`
        ...
}

type PulsarCDCConfiguration struct {
	Enabled          bool    `json:"enabled"`
	PulsarServiceUrl *string `json:"pulsarServiceUrl"`
	// +optional
        ...
}
```

If we wanted this kind of extensibility I suspect we'd be better off with a separate CRD. However, this is planning a long way forward given there are no currently planned Kafka connectors available.

The SSL properties **are Pulsar connector properties**, so I think these would always lie in the Pulsar struct in any event.

pkg/cdc/const.go
Miles-Garnsey: I can see your point here. It does seem that tightly coupling in the way I'm proposing will force a (possibly unnecessary) operator version upgrade whenever we bump a CDC agent version.

The flip side is that, if we make this change, the API we are using may change which could cause hard-to-detect failures.

My proposal would still lead to a failure in management API (at Cassandra start) which still isn't where this error is actually coming from (it comes from cass-operator starting Cassandra with the wrong Java parms).

On balance, I'm actually going to support the proposal that we DO change management API's image builds to always place CDC agent in the same location regardless of version. The implications are that management API and cass-operator become agnostic as to the CDC agent version in use.

If we subsequently find that the CDC agent API does change, we can build additional validation to catch and communicate this to the user in a clearer way.

pkg/test/testobjects.go
Miles-Garnsey: OK, will do.

pkg/cdc/generate_config.go
Miles-Garnsey: I'm still not understanding, sorry. When the CDC config changes, reconciliation of the CONFIG_DATA environment variable is required to bring it into the state declared in the CR.

The JSON still needs to be read because (in the case CDC is disabled) the CDC related configs (in cassandra-env.sh) need to be removed.

I don't understand what you're looking to change here?

burmanm: Why would you need to remove them? If there's no CDC definition set -> ConfigASJSON does not generate anything to the JSON about CDC. If it is set, then it is generated there. If the newly generated value does not match the existing one -> StatefulSet is reconciled and modified to get the new config.

In which case would this not be true? Why would ConfigASJSON generate a config with something that's no longer set and it needs to be removed?

pkg/cdc/const.go
Miles-Garnsey: Completed.

apis/cassandra/v1beta1/cdc_types.go
Miles-Garnsey: @burmanm - waiting on a response for this one.

apis/cassandra/v1beta1/cdc_types.go
Miles-Garnsey: I have made this change as requested.

config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml
Miles-Garnsey: @burmanm - waiting on a response on this one too.

pkg/test/testobjects.go
Miles-Garnsey: Done.

pkg/cdc/generate_config.go
Miles-Garnsey: I've been considering this one. The problem is that the function `serverconfig.GetModelValues` which generates the JSON is not very well structured. 

It is in a strange place, has far too many arguments, and produces an untyped `type NodeConfig map[string]interface{}`.

I followed this path for a little while regardless. I thought I could add an argument `CDCConfig cassdcapi.CDCConfiguration` to `GetModelValues` and generate the json from there. However, when I do this, I get a circular import whereby `serverconfig` imports `cassdcapi` and vice-versa.

The root of the problem is that the API package `apis/cassandra/v1beta1` is polluted with excessive numbers of functions (especially `ConfigASJSON`, which should probably be segregated from the APIs). This makes it hard to use `ConfigASJSON` to assemble the CONFIG_DATA env variable because there is no clean way to pass in all of the required fields (in the pulsar case there is a substantial number). You can never use the CR-linked struct itself if you want to call anything to do with `serverconfig.GetModelValues`.

The fix for this is to move `ConfigASJSON` into its own package, and address the questionable number of parameters and loosely typed return value of `serverconfig.GetModelValues`. I think we should actually build some sort of pipeline construct, whereby each phase performs some transformation on the config and passes it to the next.

All of this is major refactoring and I don't think it belongs in this ticket.

Makefile
```diff
@@ -126,7 +126,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test -v ./apis/... ./controllers/... -coverprofile cover.out
 
 .PHONY: integ-test
-integ-test: kustomize cert-manager ## Run integration tests from directory M_INTEG_DIR or set M_INTEG_DIR=all to run all the integration tests.
+integ-test: kustomize cert-manager ./bin/helm ## Run integration tests from directory M_INTEG_DIR or set M_INTEG_DIR=all to run all the integration tests.
```
burmanm, line 129: Just do helm here, which would lead to ``helm:`` block. When calling Helm, we should probably call it using $(HELM) instead of directly with /bin/helm. Or even check if the system has it with ``which`` command. We do that for some of the tools. Also, all these parts should be prefixed with ``.PHONY`` to avoid looking a file instead of Makefile target. 

Lets not mix up Makefile targets with filenames.

config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml
burmanm: This only configures the Datastax CDC client (not Cassandra's internal CDC - these features do not allow setting any properties to support that feature). And as you said in earlier comments on structure, this is just for the Datastax agent which only supports Pulsar.

pkg/cdc/generate_config.go
```diff
@@ -0,0 +1,124 @@
...
+	err := json.Unmarshal(config, &c)
+	if err != nil {
+		return nil, err
+	}
+	// If CassEnvSh.AddtnlJVMOptions exists, populate a string slice with it.
+	var additionalJVMOpts []string
+	if c.CassEnvSh != nil && c.CassEnvSh.AddtnlJVMOptions != nil {
+		additionalJVMOpts = *c.CassEnvSh.AddtnlJVMOptions
+	}
+	// Deal with the possibility that cassdcapi.CDCConfiguration is nil.
```
burmanm: Wouldn't it still make more sense here to just have as the first line: ``if cassDC.Spec.CDC == nil { return }`` or never go to this Update part is CDC isn't set? All Unmarshaling and other work seems unnecessary most of the time and done in every reconcile.

pkg/cdc/generate_config.go
```diff
@@ -0,0 +1,124 @@
...
+				return nil, fmt.Errorf("invalid type in CDC struct, cannot reflect %s", fieldName)
+			default: // no need to call .Elem() when we have a value type.
+				reflectedValue = reflectedCDCConfig.Field(i).Interface()
+				optsSlice = append(optsSlice, nameTag+"="+fmt.Sprintf("%s", reflectedValue))
+			}
+		}
+		CDCOpt := fmt.Sprintf("-javaagent:%s=%s", "/opt/cdc_agent/cdc-agent.jar", strings.Join(optsSlice, ","))
+		out = append(out, // We need to add these two elements to the slice first, because the management agent must start before the CDC agent.
+			"-javaagent:/opt/metrics-collector/lib/datastax-mcac-agent.jar",
+			"-javaagent:/opt/management-api/datastax-mgmtapi-agent-0.1.0-SNAPSHOT.jar",
```
burmanm: If https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/94 is ever fixed (and it was intended), then this will blow up. If we ever remove MCAC, then that would cause this to blow up also.

burmanm: Shouldn't this kind of logic live in the management-api .sh scripts to ensure things work? Instead of modifying it here, modify it there based on some ENV variable? ``ENABLE_CDC=true`` ?

Miles-Garnsey: That may be true, and I agree that the `datastax-mgmtapi-agent-0.1.0-SNAPSHOT` is unfortunate, but I can see no alternative? The logical way to solve this is to always place MAAC in the same location irrespective of version (as we are doing with the change agent) but that is out of scope for this ticket.

We have no plans to remove MCAC, so I am less worried there.

Makefile
Miles-Garnsey: This is deliberately a file target, as I quite like having it this way. I run into frequent issues with tool installations where (even where the tool isn't present) they often don't run the installation step as (for some reason) make thinks the tool is already installed.

Am I using make wrong perhaps?

pkg/cdc/generate_config.go
```diff
@@ -0,0 +1,124 @@
...
+func updateCassandraYaml(cassConfig *configData, cdcConfig *cassdcapi.CDCConfiguration) {
+	if cdcConfig != nil {
+		if cassConfig.CassandraYaml == nil {
+			cassConfig.CassandraYaml = make(map[string]interface{})
+		}
+		cassConfig.CassandraYaml["cdc_enabled"] = true
+	}
+}
+
+// removeEntryFromSlice removes all an entry from a slice if it contains a substring.
```
burmanm: Note, this does not actually remove it from the slice, this only creates a new slice without the substring string. If you use the original input optsSlice, there's still the original unwanted item in it.

pkg/cdc/generate_config.go
Miles-Garnsey: > Shouldn't this kind of logic live in the management-api .sh scripts to ensure things work? Instead of modifying it here, modify it there based on some ENV variable? ENABLE_CDC=true ?

I'd say absolutely not actually. I'd be more in favour of not configuring anything through the management API .sh scripts and putting MAAC, MCAC and change agent all under control of the operator while otherwise starting vanilla cassandra via management API.

pkg/cdc/generate_config.go
Miles-Garnsey: That's true. I should make this comment more precise.

pkg/cdc/generate_config.go
Miles-Garnsey: I think you may be right. Let me look further tomorrow.

pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -358,11 +360,14 @@ func getConfigDataEnVars(dc *api.CassandraDatacenter) ([]corev1.EnvVar, error) {
 	}
 
 	configData, err := dc.GetConfigAsJSON(dc.Spec.Config)
-
 	if err != nil {
 		return envVars, err
 	}
-	envVars = append(envVars, corev1.EnvVar{Name: "CONFIG_FILE_DATA", Value: configData})
+	cdcAdded, err := cdc.UpdateConfig(json.RawMessage(configData), *dc)
```
burmanm, line 366: As said above, would be nice either here check that if CDC isn't used and don't call the UpdateConfig, or in the function exit quickly.

tests/cdc_successful/cdc_successful_test.go
```diff
@@ -0,0 +1,121 @@
...
+				Fail(err.Error())
+			}
+			By("Waiting for all components ready")
+			readyGetter := kubectl.Get("pods").
+				WithFlag("selector", "app=cdc-testutil").
+				WithFlag("selector", "component=proxy").
+				WithFlag("namespace", "pulsar").
+				FormatOutput("jsonpath={.items[0].status.conditions[?(@.type=='Ready')].status}")
+			if err := kubectl.WaitForOutputContains(readyGetter, "True", 1800); err != nil {
+				Fail(err.Error())
```
burmanm: Why so complicated here instead of Expect(err).To(Not(HaveOccurred())) ?

tests/cdc_successful/cdc_successful_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+			By("Creating a testutils deployment")
+			step = "Deploying TestUtils"
+			k = kubectl.ApplyFiles(testUtilsDeployment)
+			ns.ExecAndLog(step, k)
+
+			By("Confirming testutils ready")
+			readyGetter = kubectl.Get("pods").
+				WithFlag("selector", "app=cdc-testutil").
+				WithFlag("namespace", namespace).
```
burmanm: You can use the ns to simplify and write less (and make it faster to read). Unnecessary components are namespace flag, By and error check. You can simply replace lines 97-104 with:

```go
readyGetter = kubectl.Get("pods").WithFlag("selector", "app=cdc-testutil").FormatOutput("jsonpath={.items[0].status.conditions[?(@.type=='Ready')].status}")
ns.WaitForOutputContains("Confirming testutils ready", readyGetter, "True", 1800)
```

tests/testdata/test-cdc/cassandra-datacenter.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc1
+spec:
+  clusterName: test-cluster
+  # The number of server nodes.
+  size: 1
+  serverImage: docker.io/k8ssandra/cass-management-api:4.0.1-v0.1.43
```
burmanm: Why serverImage? We don't usually force these in the integration tests (other than smoke tests to ensure older versions work). Also, serverVersion: 4.0.4 is probably more up-to-date.

config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml
Miles-Garnsey: I don't know about this. Cassandra doesn't really have internal CDC. It has a hook which juggles some files containing the mutations into a different folder. The more I think about it, the less I agree with this suggestion...

pkg/cdc/const.go
```diff
@@ -0,0 +1,3 @@
+package cdc
```
burmanm: I think this file is now redundant?

Makefile
burmanm: I've never seen make not run the target. I think there's probably been some other case that causes it.

If the match is a file, make will not execute that block in your Makefile. It will also always call the file first before the Makefile target.

pkg/cdc/const.go
Miles-Garnsey: It is. Good catch, thanks.

tests/testdata/test-cdc/cassandra-datacenter.yaml
Miles-Garnsey: Let me try this. It wasn't working last I tried, and kept pulling image v0.1.41 which was why I asked this question way back [here](https://github.com/k8ssandra/cass-operator/pull/328#issuecomment-1158533140).

tests/cdc_successful/cdc_successful_test.go
Miles-Garnsey: Why is the `By` unnecessary? It seems to make the tests more self-documenting.

I'll make the other two changes you suggest. It might be good to document the e2e test functions if you plan to ask people to use them.

tests/cdc_successful/cdc_successful_test.go
Miles-Garnsey: Probably because I don't really have any familiarity with ginkgo - it isn't used in any of our other projects.

I've made the requested change.

Makefile
Miles-Garnsey: Mm, it was strange behaviour. But in either event, I think that (because this target does create a file) we're best off with a file-based makefile target. If we make it a .PHONY target then won't it run all the time, even if helm has already been downloaded?

pkg/cdc/generate_config.go
Miles-Garnsey: You were right, this needed a slightly closer look. I think it makes more sense now.

pkg/reconciliation/construct_podtemplatespec.go
Miles-Garnsey: We now exit quickly. I think this is addressed.

tests/testdata/test-cdc/cassandra-datacenter.yaml
```diff
@@ -0,0 +1,24 @@
...
+  name: dc1
+spec:
+  clusterName: test-cluster
+  # The number of server nodes.
+  size: 1
+  cdc:
+    enabled: true
+    pulsarServiceUrl: pulsar://pulsar-proxy.pulsar.svc.cluster.local:6650
+    topicPrefix: persistent://public/default/events-
+    cdcWorkingDir: /var/lib/cassandra/cdc
```
burmanm, line 13: This is something I forgot to ask.. is this something that can be created with additionalVolumes and make it work?

Miles-Garnsey: You mean create the additional volume automatically depending on what you specify for `cdcWorkingDir`? You probably could, but I don't think it is idiomatic.

A directory and a volume mount are separate things, you need to be able to control both, and this setting only controls the directory side of things.

burmanm (approved):

PR #339, opened 2022-06-03
Fix: label values not validated (#355)

### LINKED ISSUE #335: K8SSAND-1523 ⁃ Label values are not validated when created
**What happened?**
I was migrating an existing cluster to the cass-operator and it had a Cluster name of `Test Cluster`. This caused cass-operator to try to label some resources with incorrect labels:

```
  Warning  ReconcileFailed  72s (x8 over 83s)  cass-operator  Service "testcluster-seed-service" is invalid: [metadata.labels: Invalid value: "cassandra-Test Cluster": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), metadata.labels: Invalid value: "Test Cluster": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), spec.selector: Invalid value: "Test Cluster": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]
```

We validate clusterName in multiple places and strip incorrect characters. We should be doing the same in these labels also (and ensure that the queries still work).

Create CassDc with `Cluster: Test Cluster`.

**Environment**

* Cass Operator version:

    `kubectl version`
    * Kubernetes cluster kind:

```
```

* Manifests:

```
```

* Cass Operator Logs:

```
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1523) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1523
┆priority: Medium

jeffbanks:
**What this PR does**:
Provides cleaning of the cluster name being used for part of overall label value.  The cluster name can contain illegal characters not allowed for label values.
 
Fixes #335 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 ```
 
 ## unreleased
+* [BUGFIX] [#355](https://github.com/k8ssandra/cass-operator/issues/335) Cleanse label values derived from cluster name, which can contain illegal chars. Include app.kubernetes.io/created-by label. 
 
 ## v1.11.0
 
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -545,7 +545,7 @@ const dns1035LabelFmt string = "[a-z]([-a-z0-9]*[a-z0-9])?"
 
 var whitelistRegex = regexp.MustCompile(`(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`)
 
-// CleanLabelValue CleanLabelName a valid label must be an empty string or consist of alphanumeric characters,
+// CleanLabelValue a valid label must be an empty string or consist of alphanumeric characters,
 // '-', '_' or '.', and must start and end with an alphanumeric.
 // Note: we apply a prefix of "cassandra-" to the cluster name value used as label name.
 // As such, empty string isn't a valid case.
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -32,7 +32,6 @@ const (
 	// RackLabel is the operator's label for the rack name
 	RackLabel = "cassandra.datastax.com/rack"
 
-	// RackLabel is the operator's label for the rack name
 	CassOperatorProgressLabel = "cassandra.datastax.com/operator-progress"
 
 	// PromMetricsLabel is a service label that can be selected for prometheus metrics scraping
@@ -53,19 +52,16 @@ const (
 	// CassandraDatacenter is deleted.
 	DecommissionOnDeleteAnnotation = "cassandra.datastax.com/decommission-on-delete"
 
-	// CassNodeState
 	CassNodeState = "cassandra.datastax.com/node-state"
 
-	// Progress states for status
 	ProgressUpdating ProgressState = "Updating"
 	ProgressReady    ProgressState = "Ready"
 
-	// Default port numbers
 	DefaultNativePort    = 9042
 	DefaultInternodePort = 7000
 )
 
-// This type exists so there's no chance of pushing random strings to our progress status
+// ProgressState - this type exists so there's no chance of pushing random strings to our progress status
 type ProgressState string
 
 type CassandraUser struct {
@@ -238,7 +234,7 @@ type NodePortConfig struct {
 	InternodeSSL int `json:"internodeSSL,omitempty"`
 }
 
-// Is the NodePort service enabled?
+// IsNodePortEnabled is the NodePort service enabled?
 func (dc *CassandraDatacenter) IsNodePortEnabled() bool {
 	return dc.Spec.Networking != nil && dc.Spec.Networking.NodePort != nil
 }
@@ -254,7 +250,7 @@ type DseWorkloads struct {
 	SearchEnabled    bool `json:"searchEnabled,omitempty"`
 }
 
-// StorageConfig defines additional storage configurations
+// AdditionalVolumes StorageConfig defines additional storage configurations
 type AdditionalVolumes struct {
 	// Mount path into cassandra container
 	MountPath string `json:"mountPath"`
@@ -533,19 +529,30 @@ func (dc *CassandraDatacenter) SetCondition(condition DatacenterCondition) {
 // GetDatacenterLabels ...
 func (dc *CassandraDatacenter) GetDatacenterLabels() map[string]string {
 	labels := dc.GetClusterLabels()
-	labels[DatacenterLabel] = dc.Name
+	labels[DatacenterLabel] = CleanLabelValue(dc.Name)
 	return labels
 }
 
 // GetClusterLabels returns a new map with the cluster label key and cluster name value
 func (dc *CassandraDatacenter) GetClusterLabels() map[string]string {
 	return map[string]string{
-		ClusterLabel: dc.Spec.ClusterName,
+		ClusterLabel: CleanLabelValue(dc.Spec.ClusterName),
 	}
 }
 
 // apimachinery validation does not expose this, copied here
 const dns1035LabelFmt string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+
+var whitelistRegex = regexp.MustCompile(`(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`)
+
+// CleanLabelValue CleanLabelName a valid label must be an empty string or consist of alphanumeric characters,
+// '-', '_' or '.', and must start and end with an alphanumeric.
+// Note: we apply a prefix of "cassandra-" to the cluster name value used as label name.
+// As such, empty string isn't a valid case.
+func CleanLabelValue(value string) string {
+	regexpResult := whitelistRegex.FindAllString(strings.Replace(value, " ", "", -1), -1)
+	return strings.Join(regexpResult, "")
+}
 
 func CleanupForKubernetes(input string) string {
 	if len(validation.IsDNS1035Label(input)) > 0 {
@@ -653,8 +660,7 @@ func (dc *CassandraDatacenter) GetConfigAsJSON(config []byte) (string, error) {
 	}
 
 	// Combine the model values with the user-specified values
-
-	modelParsed, err := gabs.ParseJSON([]byte(modelBytes))
+	modelParsed, err := gabs.ParseJSON(modelBytes)
 	if err != nil {
 		return "", errors.Wrap(err, "Model information for CassandraDatacenter resource was not properly configured")
 	}
@@ -673,6 +679,7 @@ func (dc *CassandraDatacenter) GetConfigAsJSON(config []byte) (string, error) {
 	return modelParsed.String(), nil
 }
 
+// GetNodePortNativePort
 // Gets the defined CQL port for NodePort.
 // 0 will be returned if NodePort is not configured.
 // The SSL port will be returned if it is defined,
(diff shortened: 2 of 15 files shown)


pkg/oplabels/labels.go
```diff
@@ -17,16 +18,17 @@ const (
 	NameLabelValue             = "cassandra"
 	InstanceLabel              = "app.kubernetes.io/instance"
 	VersionLabel               = "app.kubernetes.io/version"
-	CreatedByLabel             = "app.kubernetes.io/created-by"
```
burmanm, line 20: This shouldn't be removed (we should use it if it's not at the moment..)

pkg/oplabels/labels.go
```diff
@@ -17,16 +18,17 @@ const (
...
+	emptyString                = ""
```
burmanm: This should be handled in the regexp itself.

pkg/oplabels/labels.go
```diff
@@ -43,3 +45,12 @@ func HasManagedByCassandraOperatorLabel(m map[string]string) bool {
 	v, ok := m[ManagedByLabel]
 	return ok && v == ManagedByLabelValue
 }
+
+// CleanLabelName a valid label must be an empty string or consist of alphanumeric characters,
+// '-', '_' or '.', and must start and end with an alphanumeric.
+// Note: we apply a prefix of "cassandra-" to the cluster name value used as label name.
+// As such, empty string isn't a valid case.
+func CleanLabelName(labelName string) string {
+	trimLeft := strings.TrimLeft(blacklistRegex.ReplaceAllString(labelName, emptyString), labelCut)
```
burmanm: We shouldn't do trimming as the regexp itself can do that. Also, instead of removing blacklisted characters, the regexp should take care of adding all accepted characters in it. So, find all accepted characters and then join them together.

So, regexp.FindAll(regexp) and then strings.Join(regexpResult)

pkg/oplabels/labels.go
```diff
@@ -17,16 +18,17 @@ const (
 	NameLabelValue             = "cassandra"
 	InstanceLabel              = "app.kubernetes.io/instance"
 	VersionLabel               = "app.kubernetes.io/version"
-	CreatedByLabel             = "app.kubernetes.io/created-by"
+	emptyString                = ""
+	labelCut                   = ".-_"
```
burmanm: In other parts of the code we replace "_" with "-" if possible. All of these characters are accepted values, they should be allowed (just not in the first character), in which case the regexp should probably just filter them and this field itself is not necessary.

pkg/oplabels/labels.go
```diff
@@ -17,16 +18,17 @@ const (
...
 )
 
+var blacklistRegex = regexp.MustCompile(`[^a-zA-Z\d_.-]`)
```
burmanm: See notes below, but instead of blacklisting, make this regexp the one that accepts everything - and if possible, use the Kubernetes' internal regexp that's used for validation.

tests/add_operator_labels/add_operator_labels_suite_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/k8ssandra/cass-operator/pkg/oplabels"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+/**
+A valid label must be an empty string or consist of alphanumeric characters,
+'-', '_' or '.', and must start and end with an alphanumeric.
+*/
+func TestLabelNameClean(t *testing.T) {
```
burmanm: tests directory is for e2e tests, unit tests should be in the pkg/labels. Also, could we have additionally some real ones here? Like

"TestCluster" should come out as "TestCluster", while "Test Cluster" should be without the whitelabel and so on. It makes it easier for someone to verify their result if it's not just theoretical instances here. 

And also we need to verify that the values that we currently accept (and that do work) are not changed. Otherwise all existing clusters would stop working.

burmanm (dismissed):

pkg/oplabels/labels.go
jeffbanks: All accepted characters are different when existing insides the context of starting and end alphanumeric accepted names.  

Example: usage of '-', '_', and '.' are not allowed on the outside of the 'cassandra-abc_123' name group, but can reside in the name itself.  

The regex applied in one pass to allow for those inside the name group, followed by prefix and suffix trim makes it easier to understand for the developer IMO.

burmanm: It doesn't matter if they're easier for developer to read or not, whatever anyone has currently deployed in any cluster must still work.

If I have `My_-.Cluster`, then that must work. I fail to see why we would trim if regexp trims them also (if they're not accepted characters, they're not copied). What makes it easier to understand that we do something in two places?

jeffbanks: I'll flip the regex to use of whitelist pattern instead of a blacklist in the next push.  

Using this approach, I believe it still requires removal of the spaces before applying the regex (or build into the regex pattern itself), as I noticed that in the following case, the accepted chars were ignored when spaces were present.  

Usage: 
(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?
```golang
var unclean = "+!*(-_)cor @#$%^&rect _ LABEL.name-1=<>_?,."
```

Results in:
`correctLABEL.name-1`, but expecting `correct_LABEL.name-1`

tests/add_operator_labels/add_operator_labels_suite_test.go
jeffbanks: > And also we need to verify that the values that we currently accept (and that do work) are not changed. Otherwise all existing clusters would stop working.

Do we have this list of accepted values, I'm not familiar with the set being used?

burmanm: The accepted values are all values that someone could have used. Cassandra itself limits everything that can be cast to UTF-16. So all values that are accepted by regexp `(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?` should be valid moving forward.

burmanm: That is, all those UTF-16 strings that fell in to that regexp previously, should remain the same and not be modified in any way.

jeffbanks: @burmanm since the latest push has this whitelist check `(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`, in addition to removing any `spaces` included, are you good with the constraints implemented or am I missing a detail here?  

Also, I've added test checks more direct checks of "Test Cluster", etc.

pkg/oplabels/labels.go
```diff
@@ -43,3 +44,12 @@ func HasManagedByCassandraOperatorLabel(m map[string]string) bool {
 	v, ok := m[ManagedByLabel]
 	return ok && v == ManagedByLabelValue
 }
+
+// CleanLabelName a valid label must be an empty string or consist of alphanumeric characters,
+// '-', '_' or '.', and must start and end with an alphanumeric.
+// Note: we apply a prefix of "cassandra-" to the cluster name value used as label name.
+// As such, empty string isn't a valid case.
+func CleanLabelName(labelName string) string {
+	var regexpResult = whitelistRegex.FindAllString(strings.Replace(labelName, " ", "", -1), -1)
```
burmanm: Just a note, the `var` keyword is not necessary when you can use assignment operator (`:=`). So `regexpResult := whitelist ..` is just as good (and more goish). Only the global-scope ones need the var ones since they're not inside a function.

pkg/oplabels/labels.go
```diff
@@ -43,3 +44,12 @@ func HasManagedByCassandraOperatorLabel(m map[string]string) bool {
 	v, ok := m[ManagedByLabel]
 	return ok && v == ManagedByLabelValue
 }
+
+// CleanLabelName a valid label must be an empty string or consist of alphanumeric characters,
+// '-', '_' or '.', and must start and end with an alphanumeric.
+// Note: we apply a prefix of "cassandra-" to the cluster name value used as label name.
+// As such, empty string isn't a valid case.
+func CleanLabelName(labelName string) string {
```
adejanovski:
```suggestion
func CleanLabelValue(value string) string {
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 ```
 
 ## unreleased
+* [BUGFIX] [#355](https://github.com/k8ssandra/cass-operator/issues/335) Cleanse label names derived from cluster name, which can contain illegal chars for label names.
```
adejanovski:
```suggestion
* [BUGFIX] [#355](https://github.com/k8ssandra/cass-operator/issues/335) Cleanse label values derived from cluster name, which can contain illegal chars. 
```

adejanovski (changes requested):
A few changes, and you'll also need to clean [this label](https://github.com/k8ssandra/cass-operator/pull/346/files#diff-9a6ece13700a2568ec9915ade5b094433a489a30dcdb4aa75d75db4847d92b8aR543).

And I have a few tests in [my PR](https://github.com/k8ssandra/cass-operator/pull/346) that you can cherry pick.

jeffbanks:
@burmanm @emerkle826 @adejanovski 

In the latest push: Added support for handling the label `cassandradatacenter_controller` as well as addition of label clean to be part of the api, which follows pattern of `CleanupForKubernetes` usage for other resources.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -533,20 +529,31 @@ func (dc *CassandraDatacenter) SetCondition(condition DatacenterCondition) {
...
+		ClusterLabel: CleanLabelValue(dc.Spec.ClusterName),
 	}
 }
 
 // apimachinery validation does not expose this, copied here
 const dns1035LabelFmt string = "[a-z]([-a-z0-9]*[a-z0-9])?"
 
+var whitelistRegex = regexp.MustCompile(`(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`)
+
+// CleanLabelValue CleanLabelName a valid label must be an empty string or consist of alphanumeric characters,
```
adejanovski:
```suggestion
// CleanLabelValue a valid label must be an empty string or consist of alphanumeric characters,
```

pkg/oplabels/labels_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"regexp"
+	"strings"
+	"testing"
+)
+
+/**
+A valid label must be an empty string or consist of alphanumeric characters,
+'-', '_' or '.', and must start and end with an alphanumeric.
+*/
+func TestLabelNameClean(t *testing.T) {
```
adejanovski:
```suggestion
func TestLabelValueClean(t *testing.T) {
```

adejanovski (approved):
LGTM, pending tests passing successfully.

jeffbanks:
@adejanovski thanks for the quick review and keeping the comments accurate to the `label values` change!

jeffbanks:
Label values validated in recent changes.

adejanovski:
@jeffbanks, can we move forward and merge this PR?

PR #344, opened 2022-06-09
Fix service label and annotation updates

### LINKED ISSUE #330: K8SSAND-1503 ⁃ Unable to remove the additional labels on the seed service through CR
**What happened?**
We found it impossible to remove the additional labels by unsetting the key-value pairs under the field `spec.additionalServiceConfig.seedService.additionalLabels`. We first added a label `labelKey: labelValue` under the `additionalLabels` field, and cass-operator correctly adds the label to the seed service. However, later we wanted to remove the `labelKey: labelValue` label from the seed service by deleting it from the CR. However, by doing so, cass-operator does not remove the additionalLabel from the seed service.

We also tried to change the `labelKey: labelVaule` to `labelKey: null`, but cass-operator still is not able to update the labels in seed service.

We expected that the CR represents the desired state of the application, based on the declarative model of Kubernetes. So we expected that when remove the label from the `spec.additionalServiceConfig.seedService.additionalLabels` field, cass-operator would automatically remove this label from the seed service.

1. Deploy cass-operator and the `server-storage` storage class
1. Deploy a basic CassandraDB using the example CR:
    `kubectl apply -f example.yaml`
    <details><summary>example.yaml</summary>

      ```yaml
      kind: CassandraDatacenter
      metadata:
        name: cassandra-datacenter
      spec:
        clusterName: cluster1
        config:
          cassandra-yaml:
            authenticator: org.apache.cassandra.auth.PasswordAuthenticator
            authorizer: org.apache.cassandra.auth.CassandraAuthorizer
            role_manager: org.apache.cassandra.auth.CassandraRoleManager
          jvm-options:
            initial_heap_size: 800M
            max_heap_size: 800M
        managementApiAuth:
          insecure: {}
        serverType: cassandra
        serverVersion: 3.11.7
        size: 1
        storageConfig:
          cassandraDataVolumeClaimSpec:
            accessModes:
            - ReadWriteOnce
            resources:
              requests:
                storage: 3Gi
            storageClassName: server-storage
      ``````
</details>
```

1. Add additional labels to seed service by applying the cr1.yaml
    <details><summary>cr1.yaml</summary>

      ```yaml
      kind: CassandraDatacenter
      metadata:
        name: cassandra-datacenter
      spec:
        additionalServiceConfig:
          seedService:
            additionalLabels:
              labelKey: labelValue
        clusterName: cluster1
        config:
          cassandra-yaml:
            authenticator: org.apache.cassandra.auth.PasswordAuthenticator
            authorizer: org.apache.cassandra.auth.CassandraAuthorizer
            role_manager: org.apache.cassandra.auth.CassandraRoleManager
          jvm-options:
            initial_heap_size: 800M
            max_heap_size: 800M
        managementApiAuth:
          insecure: {}
        serverType: cassandra
        serverVersion: 3.11.7
        size: 1
        storageConfig:
          cassandraDataVolumeClaimSpec:
            accessModes:
            - ReadWriteOnce
            resources:
              requests:
                storage: 3Gi
            storageClassName: server-storage
      ``````
</details>
```

1. Observe that the seed service now has the additional label `labelKey: labelValue`
1. Delete the additionalLabels by applying the example CR again
1. Observe that the seed service still has the label `labelKey: labelValue`

# Environment
* Cass Operator version:

    `docker.io/k8ssandra/cass-operator:v1.10.3`
    <!-- Try: kubectl describe deployment cass-operator -->
    <!-- Note: please provide operator version and not a helm chart version -->* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:11:29Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:25:06Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:

```
minikube start --vm-driver=docker --cpus 4 --memory 4096 --kubernetes-version v1.21.0
```

The root cause is inside the https://github.com/k8ssandra/cass-operator/blob/973cea89cad9d407fcac0f2ed388e98fdae3bc2a/pkg/reconciliation/reconcile_services.go#L103 where the labels in the desired service are merged into the existing service label map. The comment states that the reason of the merge is because cass-operator does not want to mess up the labels that are added by other sources.

A possible fix is to keep track of the labels that are added by the user using the `spec.additionalServiceConfig.seedService.additionalLabels` field, so that when users delete some labels from the field, cass-operator can delete these from the service.

The similar problem happens to the annotations too: https://github.com/k8ssandra/cass-operator/blob/973cea89cad9d407fcac0f2ed388e98fdae3bc2a/pkg/reconciliation/reconcile_services.go#L104



┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1503) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1503
┆priority: Medium

emerkle826:
**What this PR does**:
Correctly applies service label and annotation changes when AdditionalLabels and/or AdditionalAnnotations are changed.

Fixes #330 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 
-* [BUGFIX] [#330](https://github.com/k8ssandra/cass-operator/issues/330) Apply correct updates to Service labels and annotations through additionalServiceConfig
+* [BUGFIX] [#330](https://github.com/k8ssandra/cass-operator/issues/330) Apply correct updates to Service labels and annotations through additionalServiceConfig (they are now validated and don't allow reserved prefixes).
 
 ## v1.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 
+* [BUGFIX] [#330](https://github.com/k8ssandra/cass-operator/issues/330) Apply correct updates to Service labels and annotations through additionalServiceConfig
+
 ## v1.11.0
 
 * [CHANGE] [#183](https://github.com/k8ssandra/cass-operator/issues/183) Move from PodDisruptionBudget v1beta1 to v1 (changes min. required Kubernetes version to 1.21)
--- a/apis/cassandra/v1beta1/cassandradatacenter_webhook.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_webhook.go
@@ -31,6 +31,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
+const (
+	datastaxPrefix string = "cassandra.datastax.com"
+	k8ssandraPrefix string = "k8ssandra.io"
+)
+
 var log = logf.Log.WithName("api")
 
 func (r *CassandraDatacenter) SetupWebhookWithManager(mgr ctrl.Manager) error {
@@ -112,6 +117,11 @@ func ValidateSingleDatacenter(dc CassandraDatacenter) error {
 
 			return attemptedTo("use multiple nodes per worker without cpu and memory requests and limits")
 		}
+	}
+
+	err := validateServiceLabelsAndAnnotations(dc)
+	if err != nil {
+		return err
 	}
 
 	return ValidateFQLConfig(dc)
@@ -242,4 +252,52 @@ func ValidateFQLConfig(dc CassandraDatacenter) error {
 	}
 
 	return nil
+}
+
+func validateServiceLabelsAndAnnotations(dc CassandraDatacenter) error {
+	// check each service
+	addSeedSvc  := dc.Spec.AdditionalServiceConfig.AdditionalSeedService
+	allPodsSvc  := dc.Spec.AdditionalServiceConfig.AllPodsService
+	dcSvc       := dc.Spec.AdditionalServiceConfig.DatacenterService
+	nodePortSvc := dc.Spec.AdditionalServiceConfig.NodePortService
+	seedSvc     := dc.Spec.AdditionalServiceConfig.SeedService
+
+	services := map[string]ServiceConfigAdditions{
+		"AdditionalSeedService" : addSeedSvc,
+	    "AllPOdsService" : allPodsSvc,
+		"DatacenterService" : dcSvc,
+		"NodePOrtService" : nodePortSvc,
+		"SeedService" : seedSvc,
+	}
+
+	var msg string
+
+	for svcName, config := range services {
+		if containsReservedAnnotations(config) || containsReservedLabels(config) {
+			msg = string(append([]byte(msg), fmt.Sprintf("configure %s with reserved annotations and/or labels (prefixes %s and/or %s)", svcName, datastaxPrefix, k8ssandraPrefix)...))
+		}
+	}
+	msg = strings.Trim(msg, " ")
+	if msg != "" {
+		return attemptedTo(msg)
+	}
+	return nil
+}
+
+func containsReservedAnnotations(config ServiceConfigAdditions) bool {
+	return containsReservedPrefixes(config.Annotations)
+}
+
+func containsReservedLabels(config ServiceConfigAdditions) bool {
+	return containsReservedPrefixes(config.Labels)
+}
+
+func containsReservedPrefixes(config map[string]string) bool {
+	for k := range config {
+		if strings.HasPrefix(k, datastaxPrefix) || strings.HasPrefix(k, k8ssandraPrefix) {
+			// reserved prefix found
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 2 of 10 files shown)


pkg/reconciliation/construct_service.go
```diff
@@ -78,7 +79,13 @@ func addAdditionalOptions(service *corev1.Service, serviceConfig *api.ServiceCon
 			service.Labels = make(map[string]string, len(serviceConfig.Labels))
 		}
 		for k, v := range serviceConfig.Labels {
-			service.Labels[k] = v
+			// make sure user specified labels don't override reserved labels
+			if oplabels.IsReservedLabel(k) || api.IsReservedLabel(k) {
+				// don't allow reserved labels as AdditionalLabels
+				log.Log.Info("Skipping reserved label from AdditionalLabels", "reserved label", k)
```
emerkle826: @jsanda Not sure if this is the desired behavior, but I implemented a log and just ignoring any user-supplied labels/annotations that conflict with reserved ones

burmanm: Just don't let user to include anything with `cassandra.datastax.com` prefix should be enough. Or even `k8ssandra.io` for that matter.

emerkle826: For both labels and annotations I presume?

Miles-Garnsey: Pardon the intrusion... Should we raise a ticket to implement logic in a webhook to reject resources which use our labels? 

Labels can be critical to integrations in k8s, so if a label gets ignored without the user realising the blast radius for unexpected behaviour can be wide.

burmanm: > For both labels and annotations I presume?

Yes. I think we mentioned somewhere that those are reserved for internal use.

burmanm: > Pardon the intrusion... Should we raise a ticket to implement logic in a webhook to reject resources which use our labels? 

I don't think it needs another ticket necessarily if Erik has time to implement it here. Just add it to the bottom of the current webhook (put the validation logic in the api package).

emerkle826: I'll do that with this PR. I'm still waiting to confirm that there won't be other labels/annotations applied to services outside of the "AdditionalLabels"/"AdditionalAnnotations" in the service config, so I'll work on the webhook bit while waiting on that.

burmanm:
You can't move the validation to webhooks only as users might not have webhooks enabled. If you wish to validate it, there's also handler.go#isValid function which is run in the reconcile process so you can reject it here also.

burmanm:
That said, it doesn't have to be in the handler.go as there's no protocol that it couldn't be elsewhere also (I might need to refactor this later..).

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -114,6 +119,11 @@ func ValidateSingleDatacenter(dc CassandraDatacenter) error {
 		}
 	}
 
+	err := ValidateServiceLabelsAndAnnotations(dc)
```
burmanm: Nit, usually this is just written with `if err := ..; err != nil {` instead of two pieces.

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -243,3 +253,51 @@ func ValidateFQLConfig(dc CassandraDatacenter) error {
...
+		"DatacenterService":     dcSvc,
+		"NodePOrtService":       nodePortSvc,
+		"SeedService":           seedSvc,
+	}
+
+	var msg string
+
+	for svcName, config := range services {
+		if containsReservedAnnotations(config) || containsReservedLabels(config) {
+			msg = string(append([]byte(msg), fmt.Sprintf("configure %s with reserved annotations and/or labels (prefixes %s and/or %s)", svcName, datastaxPrefix, k8ssandraPrefix)...))
```
burmanm: Why is string casted to `[]byte(msg)` ? Were you looking for strings.Builder?

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -243,3 +253,51 @@ func ValidateFQLConfig(dc CassandraDatacenter) error {
...
+		}
+	}
+	msg = strings.Trim(msg, " ")
```
burmanm: This is probably due to the odd casting. I think len(msg) > 0 should be enough. Or just bail out after first error.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 
+* [BUGFIX] [#330](https://github.com/k8ssandra/cass-operator/issues/330) Apply correct updates to Service labels and annotations through additionalServiceConfig
```
burmanm: Could you add notice that these are also validated now for reserved prefixes?

burmanm (approved):

PR #359, opened 2022-06-30
feat: Added labels to easily find secrets what created for cluster

Tom910:
**What this PR does**:

Right now cass-operator create 3 secret per cluster
- ${nameCluster}-superuser
- ${Datacenter.Name}-keystore
- ${Datacenter.Name}-ca-keystore

and only one ${nameCluster}-superuser has base labels and so, I can't find all the secrets created for a cluster with a command like `kubectl get secret -l 'k8ssandra.io/cluster-name=my-cluster'`. My cases - I want to delete all secrets after deleting a cluster (right now I generate a name but using a label would be easy)

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/secrets.go
+++ b/pkg/reconciliation/secrets.go
@@ -142,6 +142,9 @@ func (rc *ReconciliationContext) retrieveSuperuserSecretOrCreateDefault() (*core
 }
 
 func (rc *ReconciliationContext) createInternodeCACredential() (*corev1.Secret, error) {
+	labels := make(map[string]string)
+	oplabels.AddOperatorLabels(labels, rc.Datacenter)
+
 	secret := &corev1.Secret{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Secret",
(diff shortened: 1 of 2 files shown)


pkg/reconciliation/secrets.go
```diff
@@ -142,6 +142,9 @@ func (rc *ReconciliationContext) retrieveSuperuserSecretOrCreateDefault() (*core
 }
 
 func (rc *ReconciliationContext) createInternodeCACredential() (*corev1.Secret, error) {
+	labels := make(map[string]string)
```
burmanm, line 145: Instead of creating new empty map, use ``dc.GetDatacenterLabels()``. That adds the necessary clusterLabels also.

burmanm: This is buggy in the ``buildDefaultSuperuserSecret`` function also. 

With that GetDatacenterLabels, one can use ``-l cassandra.datastax.com/datacenter=dc2`` for example. Or ``/cluster=``.

burmanm:
Hi, thanks for the PR. Some other things that could be nice would be adding a test in ``secrets_test.go`` to verify the labels are there. And of course, we could autodelete them by setting a controller reference to all generated secrets so that deletion of CassandraDatacenter deletes the secrets also without manual process.

adejanovski:
Hi @Tom910,

do you still plan to work on this PR?

Thanks

burmanm:
Closing this in favor of https://github.com/k8ssandra/cass-operator/pull/608

PR #360, opened 2022-07-05
Create health condition for degraded status / quorum issues

### LINKED ISSUE #376: K8SSAND-1666 ⁃ Add LOCAL_QUORUM failed condition to CassandraDatacenter
**What is missing?**
cass-operator checks that LOCAL_QUORUM calls to each Cassandra node succeeds before adding and starting new nodes. However, we don't actually tell this to the user in other ways than in the logs. We should add a state to the CassandraDatacenter/status that would be more visible to the user. 

**Why do we need it?**

**Environment**

* Cass Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1666) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1666
┆priority: Medium

burmanm:
**What this PR does**:
In case CassandraDatacenter is under some operation that degrades the cluster or QUORUM health check fails, update Condition on the Status to indicate this. This is to make user aware that certain other operations are not going to be successful.

Fixes #376

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -328,6 +328,7 @@ const (
 	DatacenterRollingRestart DatacenterConditionType = "RollingRestart"
 	DatacenterValid          DatacenterConditionType = "Valid"
 	DatacenterDecommission   DatacenterConditionType = "Decommission"
+	DatacenterHealthy        DatacenterConditionType = "Healthy"
 )
 
 type DatacenterCondition struct {
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -328,7 +328,10 @@ const (
 	DatacenterRollingRestart DatacenterConditionType = "RollingRestart"
 	DatacenterValid          DatacenterConditionType = "Valid"
 	DatacenterDecommission   DatacenterConditionType = "Decommission"
-	DatacenterHealthy        DatacenterConditionType = "Healthy"
+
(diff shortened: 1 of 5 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -1161,6 +1161,30 @@ func (rc *ReconciliationContext) UpdateStatus() result.ReconcileResult {
 	return result.Continue()
 }
 
+func (rc *ReconciliationContext) UpdateHealth() result.ReconcileResult {
+	updated := false
+	dcPatch := client.MergeFrom(rc.Datacenter.DeepCopy())
+
+	if rc.isClusterDegraded() || !rc.isClusterHealthy() {
+		updated = rc.setCondition(
```
jsanda, line 1173: If the cluster is degraded or unhealthy, should we just requeue?

burmanm: We already do that elsewhere, this is just to add the status (as I thought the DBPE-2283 wish was). If we requeue here, then we could never recover from the wrong status.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -328,6 +328,7 @@ const (
 	DatacenterRollingRestart DatacenterConditionType = "RollingRestart"
 	DatacenterValid          DatacenterConditionType = "Valid"
 	DatacenterDecommission   DatacenterConditionType = "Decommission"
+	DatacenterHealthy        DatacenterConditionType = "Healthy"
```
jsanda: Can you add some comments explaining this?

jsanda (changes requested):

jsanda (approved):

PR #363, opened 2022-07-07
Allow custom Cassandra DC name

### LINKED ISSUE #362: K8SSAND-1636 ⁃ Allow the Cassandra DC name to be different than the CassandraDatacenter name
## Depends on

- [x] https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/242

**What is missing?**
The DC name as stored in Cassandra is always configured to be the same as the CassandraDatacenter name. We need to be able to specify a different name for the DC as stored in Cassandra.

**Why do we need it?**
This is needed in k8ssandra-operator to avoid naming collisions. See https://github.com/k8ssandra/k8ssandra-operator/issues/615.

This can be supported pretty easily with a small change in the `GetConfigAsJSON` method (see [here](https://github.com/k8ssandra/cass-operator/blob/241e71cdd32bd9f8a7e5c00d5427cdcaf9f55497/apis/cassandra/v1beta1/cassandradatacenter_types.go#L609-L609)) by allowing properties under the `datacenter-info` key to be overridden. I will try to include a unit test for good measure.

**Environment**

* Cass Operator version:

    `v1.11.0`
    
## Update

After discussing this offline, we want to apply the same strategy as what's been done with cluster name overrides in k8ssandra-operator:

* The cassdc CRD will get a new `DcName` field which will allow name overrides that aren't compliant with Kubernetes object names
* This name will be passed as is to cass-config-builder
* It will be sanitized to be made compliant with kubernetes names and labels, and will be used to name and label all subresources. Selectors will be modified accordingly.

In a nutshell, a new method will be created in the cassdc api looking as follows:

```
...
type CassandraDatacenterSpec struct {
...
...
  DcName string `json:"dcName,omitempty"`
...
}

// SanitizedName returns a sanitized version of the name returned by CassDcName()
func (in *CassandraDatacenter) SanitizedName() string {
	return CleanupForKubernetes(in.CassDcName())
}

// CassDcName returns the Cassandra DC name override if it exists,
// otherwise the cassdc object name.
func (in *CassandraDatacenter) CassDcName() string {
	if in.Spec.Cassandra.DcName != "" {
		return in.Spec.DcName
	}
	return in.Name
}
```

Wherever the `cassdc.Name` was previously use, we'll have to use `cassdc.SanitizedName()` instead, unless we want to refer directly to the cassdc object.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1636) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1636
┆priority: Medium

jsanda:
**What this PR does**:
Allows custom Cassandra DC name. Needed to avoid naming collisions in k8ssandra-operator.

Fixes #362

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -671,7 +671,12 @@ func (dc *CassandraDatacenter) GetConfigAsJSON(config []byte) (string, error) {
 			return "", errors.Wrap(err, "Error parsing Spec.Config for CassandraDatacenter resource")
 		}
 
-		if err := modelParsed.Merge(configParsed); err != nil {
+		collisionFn := func(dest, src interface{}) interface{} {
+			// Allow overrides of modelValues, e.g., the dc name.
+			return src
+		}
+
+		if err := modelParsed.MergeFn(configParsed, collisionFn); err != nil {
 			return "", errors.Wrap(err, "Error merging Spec.Config for CassandraDatacenter resource")
 		}
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 * [BUGFIX] [#355](https://github.com/k8ssandra/cass-operator/issues/335) Cleanse label values derived from cluster name, which can contain illegal chars. Include app.kubernetes.io/created-by label. 
-
 * [BUGFIX] [#330](https://github.com/k8ssandra/cass-operator/issues/330) Apply correct updates to Service labels and annotations through additionalServiceConfig (they are now validated and don't allow reserved prefixes).
+* [ENHANCEMENT] [#362](https://github.com/k8ssandra/cass-operator/issues/362) Allow custom Casandra DC name instead of using the CassandraDatacenter name.
 
 ## v1.11.0
 
--- a/apis/cassandra/v1beta1/cassandradatacenter_types_test.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types_test.go
@@ -1,0 +1,90 @@
+package v1beta1
+
+import (
+	"encoding/json"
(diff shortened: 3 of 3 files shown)


apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -671,7 +671,12 @@ func (dc *CassandraDatacenter) GetConfigAsJSON(config []byte) (string, error) {
 			return "", errors.Wrap(err, "Error parsing Spec.Config for CassandraDatacenter resource")
 		}
 
-		if err := modelParsed.Merge(configParsed); err != nil {
+		collisionFn := func(dest, src interface{}) interface{} {
```
burmanm, line 674: This allows overwriting also the seeds. Are we sure we want that? cass-operator would be very confused probably.

burmanm: Also, even if dc is overwritten here, all the other instances will use the older Objectmeta name. That means, label queries, logs etc will not find this given name here. I guess that's no problem if user understands it's the CassandraDatacenter.Name, not Cassandra's Datacenter name.

jsanda: >This allows overwriting also the seeds

I could modify `collisionFn` to only allow the datacenter name change. This is really intended as use at your own risk and primarily for k8ssandra-operator.

>That means, label queries, logs etc will not find this given name here. I guess that's no problem if user understands it's the CassandraDatacenter.Name, not Cassandra's Datacenter name.

Good point. Here was my motivation. A user creates a K8ssandraCluster with two DCs - `dc1` and `dc2`. When https://github.com/k8ssandra/k8ssandra-operator/issues/615 is fixed, the actual CassandraDatacenter names will be something generated by k8ssandra-operator. When I run `nodetool status`, I would expect to see `dc1` and `dc2`. This might be confusing.

I guess we need to take a step back and think about this some more.

burmanm:
Closed in favour of another PR.

PR #373, opened 2022-07-14
Consolidate some probe code

### LINKED ISSUE #377: Probe wget calls are missing a timeout
**What happened?**
When management-api is configured to use mutual TLS authentication, we are unable to use Kubernetes' httpGet for liveness and readiness probes, since Kubernetes does not allow us to use mutual auth TLS. Thus, in this case we replace the calls by executing wget on the pod that then has access to both of these certificates (since they're mounted to the pods).

If the calls never reach the target for some reason, this wget call was never exiting itself. This caused multiple wget processes to live in the pod, as they would live longer than the interval for those calls.

Each wget call should have a timeout that's shorter than (or equal to) the probing interval.

**Environment**

* Cass Operator version:

    `kubectl version`
    * Kubernetes cluster kind:

```
```

* Manifests:

```
```

* Cass Operator Logs:

```
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1667) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1667
┆priority: Medium

burmanm:
**What this PR does**:
Adds back the terminate test for drain test and consolidates from probe and presto hook URL paths.

Fixes #377

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ test: manifests generate fmt vet lint envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test -v ./apis/... ./controllers/... -coverprofile cover.out
 
 .PHONY: integ-test
-integ-test: kustomize cert-manager ./bin/helm ## Run integration tests from directory M_INTEG_DIR or set M_INTEG_DIR=all to run all the integration tests.
+integ-test: kustomize cert-manager helm ## Run integration tests from directory M_INTEG_DIR or set M_INTEG_DIR=all to run all the integration tests.
 ifeq ($(M_INTEG_DIR), all)
 	# Run all the tests (exclude kustomize & testdata directories)
 	cd tests && go test -v ./... -timeout 300m --ginkgo.progress --ginkgo.v
@@ -214,7 +214,7 @@ endif
 
 .PHONY: cert-manager
 cert-manager: ## Install cert-manager to the cluster
-	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.0/cert-manager.yaml
+	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.2/cert-manager.yaml
 	kubectl wait --for=condition=Established crd certificates.cert-manager.io
 	kubectl rollout status deployment cert-manager-webhook -n cert-manager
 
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -214,7 +214,7 @@ endif
 
 .PHONY: cert-manager
 cert-manager: ## Install cert-manager to the cluster
-	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.0/cert-manager.yaml
+	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.2/cert-manager.yaml
```
jsanda, line 217: It would be good to keep common deps in sync between cass-operator and k8ssandra-operator. Can you upgrade cert-manager in k8ssandra-operator as well?

pkg/httphelper/security.go
```diff
@@ -171,7 +170,8 @@ func (provider *InsecureManagementApiSecurityProvider) BuildMgmtApiWgetAction(en
 			"wget",
 			"--output-document", "/dev/null",
 			"--no-check-certificate",
-			fmt.Sprintf("http://%s", endpoint),
+			"--timeout 20",
```
jsanda: Should the timeout be configurable?

pkg/httphelper/security.go
```diff
@@ -185,7 +185,8 @@ func (provider *ManualManagementApiSecurityProvider) BuildMgmtApiWgetAction(endp
 			"--certificate", tlsCrt,
 			"--private-key", tlsKey,
 			"--ca-certificate", caCertPath,
-			fmt.Sprintf("https://%s", endpoint),
+			"--timeout 20",
```
jsanda: Should the timeout be configurable?

pkg/httphelper/security.go
```diff
@@ -196,8 +197,9 @@ func (provider *InsecureManagementApiSecurityProvider) BuildMgmtApiWgetPostActio
 			"wget",
 			"--output-document", "/dev/null",
 			"--no-check-certificate",
+			"--timeout 20",
```
jsanda: Should the timeout be configurable?

pkg/httphelper/security.go
```diff
@@ -211,8 +213,9 @@ func (provider *ManualManagementApiSecurityProvider) BuildMgmtApiWgetPostAction(
 			"--certificate", tlsCrt,
 			"--private-key", tlsKey,
 			"--ca-certificate", caCertPath,
+			"--timeout 20",
```
jsanda: Should the timeout be configurable?

jsanda:
@burmanm  Can you also create an issue for this with some details on why the wget calls are made and why it's a good idea to add the timeouts?

pkg/httphelper/security.go
burmanm: Not really, we don't allow configuring other mgmt-api timeouts in the code either.

jsanda: A hard coded value seems less than ideal. If someone, even for dev/testings needs to adjust it is going require building a new image. What about making it configurable through the operator config map?

burmanm: That would make it inconsistent with the httpGet calls we have as well as all the client.go (management-api communications), they're not configurable either.

We can make a ticket to configure all those, but that will require quite a lot of work and break APIs for all users of management-api client.

burmanm: I added a commit that allows it. It will break backwards compatibility, so I wouldn't backport it.

jsanda: Before you pushed the commit I was about reply saying I am ok punting if you think it is unnecessary :)

jsanda:
Changes look good other than the failing tests.

PR #374, opened 2022-07-14
If Cassandra start call fails, delete the pod

### LINKED ISSUE #370: K8SSAND-1653 ⁃ If we can't start the pod, kill it
**What is missing?**
In case we're unable to call the /start endpoint in the pod to start the actual Cassandra node, we should simply give up and kill the pod. This will cause the pod to be restarted by the StatefulSet controller and we can then retry starting it.

Otherwise we can end up in the situation where the pod doesn't start, but it will never be killed and we can never proceed with starting the next nodes.

**Why do we need it?**

**Environment**

* Cass Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1653) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-1653
┆priority: Medium
┆Link To Issue: https://k8ssandra.atlassian.net/browse/K8SSAND-1653

burmanm:
**What this PR does**:
If pod fails to start (mgt-api can't run the start command or receive it), kill the pod. Create an event warning that this happened in case this is reoccurring event (it shouldn't be common process).

Fixes #370

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,10 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 ```
 
 ## unreleased
+* [CHANGE] [#370](https://github.com/k8ssandra/cass-operator/issues/370) If Cassandra start call fails, delete the pod
 * [ENHANCEMENT] [#366](https://github.com/k8ssandra/cass-operator/issues/366) If no finalizer is present, do not do the process deletion. To prevent re-adding the finalizer, add an annotation no-finalizer that prevents re-adding.
 * [ENHANCEMENT] [#360](https://github.com/k8ssandra/cass-operator/pull/360) If Datacenter quorum reports unhealthy state, change Status Condition DatacenterHealthy to False (DBPE-2283)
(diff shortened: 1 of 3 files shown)


jsanda (approved):

jsanda:
It would be nice to have a full e2e test for this. I suspect that would take some effort. What do you think?

burmanm:
Do you have a good idea how to make the /start call fail?

jsanda:
>Do you have a good idea how to make the /start call fail?

I have _an_ idea. You would need a custom image for management-api with some additional behavior to trigger failing of the start endpoint. Maybe add another endpoint that tells management-api it should fail the start endpoint. The test could call that endpoint via `kubectl exec`.

burmanm:
> Maybe add another endpoint that tells management-api it should fail the start endpoint. The test could call that endpoint via kubectl exec.

This sounds like quite a flaky test sadly. That would require us to call the endpoint between the time that the mgmt-api is up, but we still haven't called start. There's no guarantee the test would be able to reliably do that before cass-operator calls the start method. 

I guess we would have to patch the pod before we let mgmt start.. or add a firewall rule that prevents connections to the start method. All of these sound like a recipe for flakiness ;)

PR #383, opened 2022-07-27
CassandraTask updates

### LINKED ISSUE #327: K8SSAND-1475 ⁃ Handle replacement of nodes stuck in "Starting" mode
My cluster ended up in a state where one Cassandra pod wouldn't start because the data volume was corrupted (or something like that). In this case, the management API tries over and over again to restart Cassandra and the pod gets stuck with the "Starting" cass-operator label.
One "logical" thing to do then is to use the `replaceNodes` setting of cass-operator to replace the faulty pod with a new one (including a new PV), and bootstrap it safely by replacing the previous instance of that node. Sadly, cass-operator prevents that from happening and the node never gets replaced.

The manual fix wasn't really easy, and involved:

* removing the node from the cluster through `nodetool removenode`
* deleting the PV and PVC
* then deleting the pod

The additional streaming session and token movements triggered by the node removal phase could really be avoided, as well as the follow up cleanup operation if we could make it so that cass-operator allows such replacements.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1475) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1475
┆priority: Medium

### LINKED ISSUE #384: K8SSAND-1706 ⁃ Add task for replace node
**What is missing?**
Currently replacenode requires to modify the Spec, manually delete the PVCs and kill the pod. After this, cass-operator modifies the Spec.

**Why do we need it?**
As noted in issue https://github.com/k8ssandra/cass-operator/issues/102  we should not be modifying the spec for operations. This is one of operations that should be migrated outside.

**Environment**

* Cass Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1706) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1706
┆priority: Medium

### LINKED ISSUE #387: K8SSAND-1709 ⁃ Add task for upgrade SSTables
**What is missing?**
When a Cassandra cluster is upgraded to a newer version which modifies the SSTable versions, we need to issue UpgradeSSTables request to the nodes so that they update the SSTables on the disk. There's currently no way to do this from the cass-operator.

**Why do we need it?**

**Environment**

* Cass Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1709) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1709
┆priority: Medium

burmanm:
**What this PR does**:
* Adds 3 new client.go endpoints for async operations (scrub, upgradesstables, compaction)
* Adds 3 new client.go endpoints for sync operations (scrub, compaction, upgradesstables)
* Refactor cassandratask_controller to separate jobs, reduce boilerplate and add a new validation function
* Add new CassandraTask for upgradeSSTables

Fixes #387
Fixes #384
Fixes #327 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/control/v1alpha1/cassandratask_types.go
+++ b/apis/control/v1alpha1/cassandratask_types.go
@@ -60,8 +60,10 @@ type CassandraTaskSpec struct {
 type CassandraCommand string
 
 const (
-	CommandCleanup CassandraCommand = "cleanup"
-	CommandRebuild CassandraCommand = "rebuild"
+	CommandCleanup         CassandraCommand = "cleanup"
+	CommandRebuild         CassandraCommand = "rebuild"
+	CommandRestart         CassandraCommand = "restart"
+	CommandUpgradeSSTables CassandraCommand = "upgradesstables"
 )
 
 type CassandraJob struct {
--- a/apis/control/v1alpha1/cassandratask_types.go
+++ b/apis/control/v1alpha1/cassandratask_types.go
@@ -64,6 +64,9 @@ const (
 	CommandRebuild         CassandraCommand = "rebuild"
 	CommandRestart         CassandraCommand = "restart"
 	CommandUpgradeSSTables CassandraCommand = "upgradesstables"
+	CommandReplaceNode     CassandraCommand = "replacenode"
+	CommandCompaction      CassandraCommand = "compact"
+	CommandScrub           CassandraCommand = "scrub"
 )
 
 type CassandraJob struct {
--- a/controllers/control/cassandratask_controller.go
+++ b/controllers/control/cassandratask_controller.go
@@ -30,9 +30,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	cassapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	api "github.com/k8ssandra/cass-operator/apis/control/v1alpha1"
@@ -349,7 +351,7 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 // SetupWithManager sets up the controller with the Manager.
 func (r *CassandraTaskReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&api.CassandraTask{}).
+		For(&api.CassandraTask{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Complete(r)
 }
 
--- a/controllers/control/cassandratask_controller.go
+++ b/controllers/control/cassandratask_controller.go
@@ -60,25 +60,56 @@ type AsyncTaskExecutorFunc func(httphelper.NodeMgmtClient, *corev1.Pod, *TaskCon
 // SyncTaskExecutorFunc is called as a backup if async one isn't supported
 type SyncTaskExecutorFunc func(httphelper.NodeMgmtClient, *corev1.Pod, *TaskConfiguration) error
 
+// ValidatorFunc validates that necessary parameters are set for the task
+type ValidatorFunc func(*TaskConfiguration) error
+
 // ValidatorExecutorFunc validates that necessary parameters are set for the task
-type ValidatorExecutorFunc func(*TaskConfiguration) error
+type ProcessFunc func(*TaskConfiguration) error
+
+// PodFilter approves or rejects the target pod for processing purposes.
+type PodFilterFunc func(*corev1.Pod, *TaskConfiguration) bool
 
 // TaskConfiguration sets the command's functions to execute
 type TaskConfiguration struct {
+	// Meta / status
 	Id            string
-	AsyncFunc     AsyncTaskExecutorFunc
-	SyncFunc      SyncTaskExecutorFunc
-	ValidateFunc  ValidatorExecutorFunc
-	AsyncFeature  httphelper.Feature
-	RestartPolicy corev1.RestartPolicy
-	Context       context.Context
-	Arguments     map[string]string
 	TaskStartTime *metav1.Time
+	Datacenter    *cassapi.CassandraDatacenter
+	Context       context.Context
+
+	// Input parameters
+	RestartPolicy corev1.RestartPolicy
+	Arguments     map[string]string
+
+	// Execution functionality per pod
+	AsyncFeature httphelper.Feature
+	AsyncFunc    AsyncTaskExecutorFunc
+	SyncFunc     SyncTaskExecutorFunc
+	PodFilter    PodFilterFunc
+
+	// Functions not targeting the pod
+	ValidateFunc   ValidatorFunc
+	PreProcessFunc ProcessFunc
 }
 
 func (t *TaskConfiguration) Validate() error {
 	if t.ValidateFunc != nil {
 		return t.ValidateFunc(t)
+	}
+	return nil
+}
+
+func (t *TaskConfiguration) Filter(pod *corev1.Pod) bool {
+	if t.PodFilter != nil {
+		return t.PodFilter(pod, t)
+	}
+
+	return true
+}
+
+func (t *TaskConfiguration) Process() error {
+	if t.PreProcessFunc != nil {
+		return t.PreProcessFunc(t)
 	}
 	return nil
 }
@@ -144,12 +175,12 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	// Get the CassandraDatacenter
-	dc := cassapi.CassandraDatacenter{}
+	dc := &cassapi.CassandraDatacenter{}
 	dcNamespacedName := types.NamespacedName{
 		Namespace: cassTask.Spec.Datacenter.Namespace,
 		Name:      cassTask.Spec.Datacenter.Name,
 	}
-	if err := r.Get(ctx, dcNamespacedName, &dc); err != nil {
+	if err := r.Get(ctx, dcNamespacedName, dc); err != nil {
 		logger.Error(err, "unable to fetch CassandraDatacenter", "Datacenter", cassTask.Spec.Datacenter)
 		// This is unrecoverable error at this point, do not requeue
 		return ctrl.Result{}, err
@@ -161,7 +192,7 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	// If we're active, we can proceed - otherwise verify if we're allowed to run
 	if !found {
 		// Does our concurrencypolicy allow the task to run? Are there any other active ones?
-		activeTasks, err := r.activeTasks(ctx, &dc)
+		activeTasks, err := r.activeTasks(ctx, dc)
 		if err != nil {
 			logger.Error(err, "unable to fetch active CassandraTasks", "Request", req.NamespacedName)
 			return ctrl.Result{}, client.IgnoreNotFound(err)
@@ -189,13 +220,13 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 
 		// Link this resource to the Datacenter and copy labels from it
-		if err := controllerutil.SetControllerReference(&dc, &cassTask, r.Scheme); err != nil {
+		if err := controllerutil.SetControllerReference(dc, &cassTask, r.Scheme); err != nil {
 			logger.Error(err, "unable to set ownerReference to the task", "Datacenter", cassTask.Spec.Datacenter, "CassandraTask", req.NamespacedName)
 			return ctrl.Result{}, err
 		}
 
 		utils.MergeMap(cassTask.Labels, dc.GetDatacenterLabels())
-		oplabels.AddOperatorLabels(cassTask.GetLabels(), &dc)
+		oplabels.AddOperatorLabels(cassTask.GetLabels(), dc)
 
 		// Starting the run, set the Active label so we can quickly fetch the active ones
 		cassTask.GetLabels()[taskStatusLabel] = activeTaskLabelValue
@@ -225,6 +256,7 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		taskConfig := &TaskConfiguration{
 			RestartPolicy: cassTask.Spec.RestartPolicy,
 			Id:            taskId,
+			Datacenter:    dc,
 			TaskStartTime: cassTask.Status.StartTime,
 			Context:       ctx,
 			Arguments:     job.Arguments,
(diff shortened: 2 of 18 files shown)


controllers/control/cassandratask_controller.go
```diff
@@ -55,19 +55,63 @@ type CassandraTaskReconciler struct {
...
+type AsyncTaskExecutorFunc func(httphelper.NodeMgmtClient, *corev1.Pod, *TaskConfiguration) (string, error)
 
 // SyncTaskExecutorFunc is called as a backup if async one isn't supported
-type SyncTaskExecutorFunc func(httphelper.NodeMgmtClient, *corev1.Pod, map[string]string) error
+type SyncTaskExecutorFunc func(httphelper.NodeMgmtClient, *corev1.Pod, *TaskConfiguration) error
+
+// ValidatorFunc validates that necessary parameters are set for the task
+type ValidatorFunc func(*TaskConfiguration) error
+
+// ValidatorExecutorFunc validates that necessary parameters are set for the task
```
adutra: Explanation doesn't seem right. Also it seems this type of function is used for preprocessing? In this case can we rename to `PreProcessFunc`?

controllers/control/cassandratask_controller.go
```diff
@@ -55,19 +55,63 @@ type CassandraTaskReconciler struct {
...
+
+func (t *TaskConfiguration) Filter(pod *corev1.Pod) bool {
+	if t.PodFilter != nil {
+		return t.PodFilter(pod, t)
+	}
+
+	return true
+}
+
+func (t *TaskConfiguration) Process() error {
```
adutra: Method is called `Process` but calls `PreProcessFunc`. Maybe rename method to `PreProcess`?

controllers/control/cassandratask_controller.go
```diff
@@ -451,7 +502,16 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc
 	}
 
 	failed, completed := 0, 0
+
 	for idx, pod := range dcPods {
+		// TODO Do we need post-pod processing functionality also? In case we need to wait for some other event to happen (processed by cass-operator).
+		//		or could we requeue with the filter instead of only bypassing? Or could we requeue in the Async / SyncFunc?
```
adutra, line 553: This sentence is not clear to me.

controllers/control/cassandratask_controller.go
```diff
@@ -209,22 +255,54 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		case api.CommandRestart:
+			r.restart(taskConfig)
+		case api.CommandReplaceNode:
+			/*
+				TODO The death of those pods should cause a restart.. what happens if someone wants to replace all the pods? Will we kill
+				  the datacenter? Do we need to wait for the new pods to be alive before we continue? Can we requeue from the process
+				  to make sure we retry the process (it isn't finished yet..) or just make it stuck?
+			*/
+			r.replace(taskConfig)
+		case "forceupgraderacks":
```
adutra, line 279: Do we still need this command? Or is it something to do later?

controllers/control/cassandratask_controller.go
```diff
@@ -209,22 +255,54 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			// res, failed, completed, err = r.reconcileEveryPodTask(ctx, &dc, scrub(taskConfigProto))
+		case api.CommandCompaction:
+			// res, failed, completed, err = r.reconcileEveryPodTask(ctx, &dc, compact(taskConfigProto, job.Arguments))
 		default:
 			err = fmt.Errorf("unknown job command: %s", job.Command)
 			return ctrl.Result{}, err
 		}
 
+		if err := taskConfig.Validate(); err != nil {
+			// TODO Or should we mark this task as failed and simply stop processing?
```
adutra: The validation could fail e.g. if the pod name does not exist. If the user is able to rectify a wrong task argument, I think it makes sense to requeue, no?

adutra (commented):
Just a very superficial review for now, I will continue tomorrow and try to dig deeper.

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -60,8 +60,13 @@ type CassandraTaskSpec struct {
 type CassandraCommand string
 
 const (
-	CommandCleanup CassandraCommand = "cleanup"
-	CommandRebuild CassandraCommand = "rebuild"
+	CommandCleanup         CassandraCommand = "cleanup"
```
adutra, line 63: In pkg/reconciliation/reconcile_racks.go:2107 there is one occurrence of `"cleanup"` as a string instead of `CommandCleanup`.

Speaking of which, the method `ReconciliationContext.createTask` looks a bit suspicious:

In pkg/reconciliation/reconcile_racks.go:2202 the command is always `CommandCleanup` regardless of the value of the `command` argument. I would rewrite the method as follows:

```go
func (rc *ReconciliationContext) createTask(command taskapi.CassandraCommand) error {
	generatedName := fmt.Sprintf("%s-%d", command, time.Now().Unix())
	dc := rc.Datacenter

	task := &taskapi.CassandraTask{
...
			Jobs: []taskapi.CassandraJob{
				{
					Name:    fmt.Sprintf("%s-%s", command, rc.Datacenter.Name),
					Command: command,
				},
			},
...
	}
...
	return rc.Client.Status().Patch(rc.Ctx, dc, dcPatch)
}
```

controllers/control/cassandratask_controller.go
```diff
@@ -48,26 +49,76 @@ const (
...
-type SyncTaskExecutorFunc func(httphelper.NodeMgmtClient, *corev1.Pod, map[string]string) error
+type SyncTaskExecutorFunc func(httphelper.NodeMgmtClient, *corev1.Pod, *TaskConfiguration) error
+
+// ValidatorFunc validates that necessary parameters are set for the task
+type ValidatorFunc func(*TaskConfiguration) error
+
+// ValidatorExecutorFunc validates that necessary parameters are set for the task
+type ProcessFunc func(*TaskConfiguration) error
+
+// PodFilter approves or rejects the target pod for processing purposes.
```
adutra:
```suggestion
// PodFilterFunc approves or rejects the target pod for processing purposes.
```

controllers/control/jobs.go
```diff
@@ -0,0 +1,239 @@
...
+	// TODO Add more arguments configurations
+	keyspaceName := ""
+	if keyspace, found := taskConfig.Arguments["keyspace_name"]; found {
+		keyspaceName = keyspace
+	}
+	return nodeMgmtClient.CallKeyspaceCleanupEndpoint(pod, -1, keyspaceName, nil)
+}
+
+func cleanup(taskConfig *TaskConfiguration) {
+	taskConfig.AsyncFeature = httphelper.AsyncSSTableTasks
```
adutra, line 37: Trying to understand this.

The goal is to determine whether whether we should invoke `/api/v1/ops/keyspace/cleanup` or `/api/v0/ops/keyspace/cleanup`. 

The feature tested is `AsyncSSTableTasks`.

But the only other place where we use `AsyncSSTableTasks` is in pkg/reconciliation/decommission_node.go:170. It's used there to determine whether to invoke `/api/v1/ops/node/decommission` or `/api/v0/ops/node/decommission`.

So my question is: what is the `AsyncSSTableTasks` feature exactly? It seems like it corresponds to several v1 endpoints.

controllers/control/jobs.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *CassandraTaskReconciler) restart(taskConfig *TaskConfiguration) {
+	// As a short version, what if we just upgraded the Datacenter.Status fields and let dc reconciler take care of the function still?
+	// We only need to get the dc reconciler to refresh itself when we do that (by default it ignores status updates)
+	taskConfig.AsyncFeature = httphelper.Feature("") // This should never match
```
adutra: This line is redundant since the zero value of `httphelper.Feature` is `httphelper.Feature("")`.

controllers/control/jobs.go
```diff
@@ -0,0 +1,239 @@
...
+func rebuild(taskConfig *TaskConfiguration) {
+	taskConfig.AsyncFeature = httphelper.Rebuild
+	taskConfig.AsyncFunc = callRebuild
+}
+
+// Rolling restart functionality
+
+func (r *CassandraTaskReconciler) callRestartSync(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) error {
+	podStartTime := pod.GetCreationTimestamp()
+	// TODO How do we verify the previous pod has actually restarted before we move on to restart the next one?
```
adutra, line 57: How are you going to handle these TODOs, are you going to create follow-up issues?

controllers/control/jobs.go
```diff
@@ -0,0 +1,239 @@
...
+	return fmt.Errorf("valid pod_name to replace is required")
+}
+
+func replaceFilter(pod *corev1.Pod, taskConfig *TaskConfiguration) bool {
+	// If pod isn't in the to be replaced pods, return false
+	podName := taskConfig.Arguments[replacePodNameArgument]
+	return pod.Name == podName
+}
+
+// replaceProcess adds enough information to CassandraDatacenter to ensure cass-operator knows this pod is being replaced
```
adutra:
```suggestion
// replacePreProcess adds enough information to CassandraDatacenter to ensure cass-operator knows this pod is being replaced
```

controllers/control/jobs.go
```diff
@@ -0,0 +1,239 @@
...
+func (r *CassandraTaskReconciler) setDatacenterCondition(dc *cassapi.CassandraDatacenter, condition *cassapi.DatacenterCondition) {
+	if dc.GetConditionStatus(condition.Type) != condition.Status {
+		// We are changing the status, so record the transition time
+		condition.LastTransitionTime = metav1.Now()
+		dc.SetCondition(*condition)
+	}
+}
+
+func (r *CassandraTaskReconciler) replace(taskConfig *TaskConfiguration) {
+	taskConfig.AsyncFeature = httphelper.Feature("") // This should never match
```
adutra: Same here.

controllers/control/cassandratask_controller_test.go
```diff
@@ -156,6 +160,7 @@ func waitForTaskCompletion(taskKey types.NamespacedName) *api.CassandraTask {
 
 var _ = Describe("Execute jobs against all pods", func() {
 	jobRunningRequeue = time.Duration(1 * time.Millisecond)
```
adutra, line 162: Nit: these casts are redundant.

tests/util/ginkgo/lib.go
```diff
@@ -514,12 +514,29 @@ func (ns NsWrapper) CqlExecute(podName string, stepDesc string, cql string, user
...
+func (ns *NsWrapper) CheckForCassandraTasks(dcName, command string, completed bool, count int) {
 	step := fmt.Sprintf("checking that cassandratask command %s has succeeded", command)
 	json := "jsonpath={.items[*].spec.jobs[0].command}"
 	k := kubectl.Get("cassandratask").
-		WithLabel(fmt.Sprintf("cassandra.datastax.com/datacenter=%s", dcName)).
-		WithLabel("control.k8ssandra.io/status=completed").
-		FormatOutput(json)
+		WithLabel(fmt.Sprintf("cassandra.datastax.com/datacenter=%s", dcName))
+
+	if completed {
```
adutra, line 523: Nit: when `completed` is false, I would have expected the `control.k8ssandra.io/status` to NOT be equal to `completed`. Here instead, we retrieve all tasks, both complete and incomplete.

adutra (changes requested):

controllers/control/cassandratask_controller.go
burmanm: Well, the intention was at some point to use the same function type for post-processing also. There's just no jobs yet in this release that require it. That's why it wasn't called PreProcessFunc.

controllers/control/cassandratask_controller.go
burmanm: Yeah, here it makes more sense.

controllers/control/cassandratask_controller.go
burmanm: Currently, since we don't have post-processing (that would wait for resources to get back), the task could actually restart or replace all the nodes in the Datacenter before the Datacenter can rebootstrap any of the nodes (thus, taking the cluster down).

controllers/control/cassandratask_controller.go
burmanm: It's something that should come later (it currently requires cass-operator to modify the Spec)

controllers/control/cassandratask_controller.go
burmanm: Yes, that's what it does now (allows fixing). But on the other hand, the "Job" type in Kubernetes just fails and refuses to rerun. So semantics are a bit different. That's why the question.

apis/control/v1alpha1/cassandratask_types.go
burmanm: reconcile_racks.go only makes cleanup tasks in any case at the moment. But sure, that's not a big thing to fix (it doesn't necessarily work with other commands though, since it doesn't accept args for example).

controllers/control/jobs.go
burmanm: It's a "generational" thing. Some async tasks were created at the same time when AsyncSSTableTasks feature was created. So if AsyncSSTableTasks is true, then those features will also work. Each command doesn't have it's own feature in that version.

tests/util/ginkgo/lib.go
burmanm: That was the intention. The idea was to fetch all possible tasks to ensure no extra ones are created.

controllers/control/jobs.go
burmanm: This feature will not close in this PR, so yes - it has its own ticket already.

controllers/control/jobs.go
jsanda: I can see this being a source of confusion for others as well. @burmanm can you please put your explanation in a comment to provide context for others.

adutra: Thanks for the explanation, makes total sense now. Yes a little comment on the `AsyncSSTableTasks` declaration would definitely help.

tests/util/ginkgo/lib.go
adutra: Maybe rename to `onlyComplete` for clarity?

adutra (approved):

burmanm:
The test failure looks to be failure in test design or potentially a bug in the cass-operator (but not from this PR really). It seems the node is replaced correctly, but cass-operator puts itself in the "quiet period" and does not change the status of replace to complete. Maybe I'm replacing the same node twice too fast in the test and causing this issue.

burmanm:
Some relevant parts from the log:

Start of the process:
```1.6601977672816293e+09	DEBUG	controller.cassandratask	starting execution of sync blocking job	{"reconciler group": "control.k8ssandra.io", "reconciler kind": "CassandraTask", "name": "replace-node", "namespace": "test-node-replace", "Pod": {"apiVersion": "v1", "kind": "Pod", "namespace": "test-node-replace", "name": "cluster1-dc1-r3-sts-0"}}
1.660197767282e+09	INFO	controller.cassandratask	calling Management API drain node - POST /api/v0/ops/node/drain	{"reconciler group": "control.k8ssandra.io", "reconciler kind": "CassandraTask", "name": "replace-node", "namespace": "test-node-replace", "pod": "cluster1-dc1-r3-sts-0"}
```

Then cass-operator notices it pretty quickly that pod has been deleted:

```
1.6601977791980631e+09	INFO	controllers.CassandraDatacenter	======== handler::Reconcile has been called	{"cassandradatacenter": "test-node-replace/dc1", "requestNamespace": "test-node-replace", "requestName": "dc1", "loopID": "62fce0a3-2cd0-4825-b148-f320a50476fd"}
```

Eventually we'll notice:

```
1.6601979557445126e+09	INFO	controllers.CassandraDatacenter.cassandradatacenter_controller.controller.cassandradatacenter-controller	Finished replacing pod	{"reconciler group": "cassandra.datastax.com", "reconciler kind": "CassandraDatacenter", "name": "dc1", "namespace": "test-node-replace", "namespace": "test-node-replace", "datacenterName": "dc1", "clusterName": "cluster1", "pod": "cluster1-dc1-r3-sts-0"}
1.6601979557446992e+09	INFO	controllers.CassandraDatacenter.cassandradatacenter_controller.controller.cassandradatacenter-controller	Finished replacing pod cluster1-dc1-r3-sts-0	{"reconciler group": "cassandra.datastax.com", "reconciler kind": "CassandraDatacenter", "name": "dc1", "namespace": "test-node-replace", "reason": "FinishedReplaceNode", "eventType": "Normal"}
```

That should update the status in cass-operator, but for some reason there's no error or update.

burmanm:
That said, I think there's a bug. It comes in the form of "PreProcess" being run too many times and that could cause this to happen also.

PR #394, opened 2022-08-19
Add Restart task to CassandraTasks

### LINKED ISSUE #385: K8SSAND-1707 ⁃ Add task for rolling restart
**What is missing?**
At the moment, to do a rolling restart, we must modify the CassandraDatacenter spec. We want a task that would do the same without requiring the modification of Spec.

**Why do we need it?**
As noted in issue https://github.com/k8ssandra/cass-operator/issues/102 we should not be modifying the spec for operations. This is one of operations that should be migrated outside.

**Environment**

* Cass Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1707) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1707
┆priority: Medium

burmanm:
**What this PR does**:
Creates a new option that does rolling restart outside cads-operator's controller. 

Fixes #385

**Checklist**
- [x] Changes manually tested


bradfordcp:
I see there is a `keyspaceArgument` in the changes, does this relate to the restart?

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -69,6 +73,12 @@ const (
 	CommandScrub           CassandraCommand = "scrub"
 )
 
+const (
+	KeyspaceArgument         string = "keyspace_name"
```
bradfordcp, line 77: How is this related to a restart?

controllers/control/jobs.go
```diff
@@ -3,22 +3,27 @@ package control
...
+	api "github.com/k8ssandra/cass-operator/apis/control/v1alpha1"
 )
 
 // Cleanup functionality
 
 func callCleanup(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) (string, error) {
 	// TODO Add more arguments configurations
 	keyspaceName := ""
-	if keyspace, found := taskConfig.Arguments["keyspace_name"]; found {
+	if keyspace, found := taskConfig.Arguments[api.KeyspaceArgument]; found {
```
bradfordcp, line 26: How is this related to a restart?

controllers/control/jobs.go
```diff
@@ -27,7 +32,7 @@ func callCleanup(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, task
 func callCleanupSync(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) error {
 	// TODO Add more arguments configurations
 	keyspaceName := ""
-	if keyspace, found := taskConfig.Arguments["keyspace_name"]; found {
+	if keyspace, found := taskConfig.Arguments[api.KeyspaceArgument]; found {
```
bradfordcp, line 35: How is this related to a restart?

controllers/control/jobs.go
burmanm: This is mostly superceded by #395 , just to ensure we don't have magic strings around the code. Same for previous comment.

controllers/control/jobs.go
burmanm: And same here

controllers/control/cassandratask_controller.go
```diff
@@ -273,7 +275,17 @@ func (r *CassandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 			cleanup(taskConfig)
 		case api.CommandRestart:
-			r.restart(taskConfig)
+			// This job is targeting StatefulSets and not Pods
+			sts, err := r.getDatacenterStatefulSets(ctx, dc)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+
+			res, err = r.restartSts(ctx, sts, taskConfig)
```
adejanovski, line 284: The `res` value doesn't seem to be used at all. Maybe that's what could be causing prematurely setting the completion time?

controllers/control/cassandratask_controller_test.go
```diff
@@ -158,254 +223,361 @@ func waitForTaskCompletion(taskKey types.NamespacedName) *api.CassandraTask {
...
+				Expect(k8sClient.Create(context.Background(), task)).Should(Succeed())
 
-			// Recreate it so the process "finishes"
-			createPod(testNamespaceName, fmt.Sprintf("test-%s", testDatacenterName), testDatacenterName, 2)
+				Eventually(func() bool {
+					Expect(k8sClient.Get(context.TODO(), stsKey, &sts)).Should(Succeed())
+					_, found := sts.Spec.Template.ObjectMeta.Annotations[api.RestartedAtAnnotation]
+					return found
+				}).Should(BeTrue())
+
```
adejanovski, line 508: You could maybe add a check here to see if the completion time is still `nil`.

adejanovski (changes requested):
The code mostly looks good to me, aside from a Result return that isn't used when invoking the rolling restart on the STS, unless I'm missing something.

It could be causing the only thing that looks problematic when testing this feature, which is that the completion time gets set right away when the CassandraTask gets created.
We'd need the completion time to be set only when the rolling restart is completed.

controllers/control/cassandratask_controller.go
burmanm: Isn't that res used on the line 333? I don't see it being overwritten anywhere.

controllers/control/cassandratask_controller_test.go
burmanm: Done.

controllers/control/cassandratask_controller.go
adejanovski: Isn't it overwritten on line 320?

```
res, failed, completed, err = r.reconcileEveryPodTask(ctx, dc, taskConfig)
```

adejanovski: Or is `break JobDefinition` sending us straight to line 333?

burmanm: break JobDefinition should break us from the for() clause and thus line 320 should never be run.

adejanovski (approved):
Looks good now 👍

PR #403, opened 2022-09-07
Make bootstrap operations deterministic

### LINKED ISSUE #381: K8SSAND-1697 ⁃ Make Bootstrap Operations Deterministic
**What is missing?**
When adding nodes to a cluster or creating a new cluster we can end up in a state where we do not bootstrap nodes in a balanced manner. This is possible when the number of nodes bootstrapping is greater than the number of racks.

Ideally we would sort the set of nodes to bootstrap by rack. Then the rack with the most nodes to bootstrap would have a node bootstrap. If there are any ties for the rack with the most nodes to bootstrap we should sort the racks by name.

**Why do we need it?**

Data can be skewed if we don't maintain balance during bootstrap operations. This is less of a concern with vnode clusters. When utilizing single-token we may end up with multiple replicas on a single node

**Environment**

* Cass Operator version: v1.12.0

Y'all are awesome, thanks for the great work :star2:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1697) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1697
┆priority: Medium

adutra:
**What this PR does**:
Makes bootstrap order 100% balanced and deterministic.

Fixes #381 .

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -5,7 +5,6 @@ package reconciliation
 
 import (
 	"fmt"
-	"math"
 	"reflect"
 	"sort"
 	"strconv"
@@ -618,6 +617,8 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
 		)
 		return result.RequeueSoon(5)
 	}
+
+	// step 4 - make sure all nodes are ready
 
 	nodeNotReady, err = rc.startAllNodes(endpointData)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


burmanm:
If you wish to write a test for this (using fakeclient), then it's possible with some helpers already in cass-operator. For example, https://github.com/k8ssandra/cass-operator/blob/master/pkg/reconciliation/reconcile_racks_test.go#L1625 calls startCassandra, but the more interesting for you is of course https://github.com/k8ssandra/cass-operator/blob/master/pkg/reconciliation/reconcile_racks_test.go#L1406

It shows how to mock the pod endpoint that the fakeclients then call, without doing too much coding (it's not intuitive at all, but copy-pasting parts of those existing tests usually ends up making one understand how to achieve something). And you can call a single test from the editor which makes iterating quite quick.

adutra:
@burmanm thank you. I managed to create a unit test that is not too lame.

Would you mind reviewing this? It should be a quick review.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -617,11 +618,13 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
 		return result.RequeueSoon(5)
 	}
 
-	needsMoreNodes, err := rc.startAllNodes(endpointData)
+	// step 4 - make sure all nodes are ready
```
burmanm, line 620: Now the step 3 comment is wrong as it used to be this step.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -617,11 +618,13 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
...
+
+	nodeNotReady, err = rc.startAllNodes(endpointData)
```
burmanm: Amm, why the same boolean for special process of restarting one node per rack and for rest of the starts? I don't see the point of this refactor. Yet most of the other parts still have their own booleans.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -1806,97 +1809,99 @@ func (rc *ReconciliationContext) startCassandra(endpointData httphelper.CassMeta
...
+		notReady, err = rc.startNode(pod, labelSeedBeforeStart, endpointData)
+		if notReady {
+			break
+		}
 
-	externalSeedPoints := 0
-	if existingEndpoints, err := rc.GetAdditionalSeedEndpoint(); err == nil {
-		externalSeedPoints = len(existingEndpoints.Subsets[0].Addresses)
 	}
+	return
```
burmanm: The named returns serve no purpose here, they only make reading more difficult. They're not used in cass-operator source code anywhere else.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -1806,97 +1809,99 @@ func (rc *ReconciliationContext) startCassandra(endpointData httphelper.CassMeta
...
+				if notReady {
+					return
 				}
-				return rackName, nil
+
+				done = done && podRankWithinRack == maxPodRankInThisRack
 			}
 		}
+		if done {
+			return
```
burmanm: Same as above.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -1806,97 +1809,99 @@ func (rc *ReconciliationContext) startCassandra(endpointData httphelper.CassMeta
...
+// startAllNodes starts all nodes in the datacenter in a deterministic order, dictated by the order
+// in which the racks were declared in the datacenter spec, and in such a way that at any given
+// time, the number of started nodes in a rack cannot be greater or lesser than the number of
+// started nodes in all other racks by more than 1. It returns true if a Cassandra node is not ready
+// yet, and an error if that node also failed to start.
+func (rc *ReconciliationContext) startAllNodes(endpointData httphelper.CassMetadataEndpoints) (notReady bool, err error) {
+	rc.ReqLogger.Info("reconcile_racks::startAllNodes")
 
-	labelSeedBeforeStart := readySeeds == 0 && len(rc.Datacenter.Spec.AdditionalSeeds) == 0 && externalSeedPoints == 0
+	for podRankWithinRack := 0; ; podRankWithinRack++ {
```
burmanm: Out of interest, would you explain the benefit vs, lets say following algorithm (the problem here looks very much like a dynamic programming)

```go
smallestRack := 0
for i := 0; i < len(rc.statefulSets); i++ {
    if rc.statefulSets[i].Status.ReadyReplicas < rc.statefulSets[smallestRack].Status.ReadyReplicas {
        smallestRack = i
    }
}

fmt.Printf("Next rack to start is: %d\n", smallestRack)
```

Also, is .Replicas really the number you want to compare to vs. ReadyReplicas?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -1806,97 +1809,99 @@ func (rc *ReconciliationContext) startCassandra(endpointData httphelper.CassMeta
...
-						"Labeled pod a seed node %s", pod.Name)
+			maxPodRankInThisRack := int(*statefulSet.Spec.Replicas) - 1
+			if podRankWithinRack <= maxPodRankInThisRack {
 
-					// sleeping five seconds for DNS paranoia
-					time.Sleep(5 * time.Second)
-				}
-				if err := rc.startCassandra(endpointData, pod); err != nil {
-					return "", err
+				pod := rc.getDCPodByName(statefulSet.Name + "-" + strconv.Itoa(podRankWithinRack))
```
burmanm: getStatefulSetPodNameForIdx(sts, podRankWithinRack) would have probably worked fine, no need to replicate the naming structure here. Doesn't really matter though.

pkg/reconciliation/reconcile_racks.go
adutra: OK will fix comment for step 3

pkg/reconciliation/reconcile_racks.go
adutra: I reused the same boolean because they really mean exactly the same thing: there is at least one pod not ready. But ok, if you want each step to have its own boolean, let's do it.

pkg/reconciliation/reconcile_racks.go
adutra: I obviously think the opposite, otherwise I wouldn't have used named return parameters in the first place :-)

But if you don't like them, let's remove them.

pkg/reconciliation/reconcile_racks.go
adutra: First, my understanding is that we want here to use the *desired* number of pods for each sts, thus we should use `.Replicas`.

Second, switching to `.ReadyReplicas` could indeed simplify the code, but can we guarantee that if a pod is counted as a ready replica, it is not only ready, but also started and has joined the ring? Because if that's not the case, we should NOT keep bootstrapping the next pods.

pkg/reconciliation/reconcile_racks.go
adutra: Ah nice, I missed that function, sorry. Will use it.

pkg/reconciliation/reconcile_racks.go
burmanm: How would Replicas help with that behavior? Isn't .Spec.Replicas updated when we add the pods? They don't have to be ready at all. Our startup process should take care of preventing next startup (there's one pod that's Started-but-not-Ready) before starting next ones. Though for that reason, perhaps it doesn't matter which one we use, since when we get to this point, we should have no starting pods anymore.

burmanm: Other than of course for the algorithm I mentioned, there the ReadyReplicas does matter since Replicas has wrong calculation.

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -1671,3 +1673,121 @@ func TestFailedStart(t *testing.T) {
 
 	mockClient.AssertExpectations(t)
 }
+
+func TestReconciliationContext_startAllNodes(t *testing.T) {
+
+	tests := []struct {
```
jsanda, line 1684: A comment that briefly explains the fixture would be helpful :) I had to read through the test setup, which is rather complex, before understanding `racks`.

pkg/reconciliation/reconcile_racks.go
adutra: So, what should I do here? Should I change as you suggested above?

burmanm: No, not necessarily. I only want to ensure that whoever reads it next can easily see what it's doing (and so avoid anything that looks pretty and complex). Would you think from that perspective the code is self-explanatory enough (the comment on the method only describes what we want to achieve, not how we achieve it) ?

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -1676,15 +1675,21 @@ func TestFailedStart(t *testing.T) {
 
 func TestReconciliationContext_startAllNodes(t *testing.T) {
 
+	// A boolean representing the state of a pod (started or not).
```
jsanda, line 1678: Very helpful. Thank you!

pkg/reconciliation/reconcile_racks.go
adutra: To me, it is self-explanatory as it is now, however that's a subjective notion. You are never going to get everybody on the same page on such matters.

burmanm (approved):

PR #405, opened 2022-09-08
Replace all invalid characters from rackNames when used in Kubernetes

### LINKED ISSUE #404: K8SSAND-1772 ⁃ User rackName should not be trusted
**What happened?**
If user enters a rackName that can't be used as is on the Kubernetes side, we should be filtering all the non-accepted characters from it.

See https://github.com/k8ssandra/k8ssandra-operator/issues/673 for more details.

**Environment**

* Cass Operator version:

    `kubectl version`
    * Kubernetes cluster kind:

```
```

* Manifests:

```
```

* Cass Operator Logs:

```
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1772) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1772
┆priority: Medium

burmanm:
… labels or resource names.

**What this PR does**:
If rackName contains characters or structure that's not accepted by the Kubernetes (in label values or in constructed resource names), remove those invalid characters.

While for new clusters it makes more sense to reject them, we have to reserve them for older ones migrating to Kubernetes.

Fixes #404 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -559,12 +559,12 @@ func (dc *CassandraDatacenter) GetClusterLabels() map[string]string {
 const (
 	dns1035LabelFmt     string = "[a-z]([-a-z0-9]*[a-z0-9])?"
 	dns1123SubdomainFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
-	labelFmt            string = "(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?"
+	dns1123LabelFmt     string = "(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?"
 )
 
 var dns1035LabelRegexp = regexp.MustCompile(dns1035LabelFmt)
 var dns1123SubdomainRegexp = regexp.MustCompile(dns1123SubdomainFmt)
-var dns1123LabelRegexp = regexp.MustCompile(labelFmt)
+var dns1123LabelRegexp = regexp.MustCompile(dns1123LabelFmt)
 
 // CleanLabelValue a valid label must be an empty string or consist of alphanumeric characters,
 // '-', '_' or '.', and must start and end with an alphanumeric.
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -474,7 +474,7 @@ func (dc *CassandraDatacenter) GetServerImage() string {
 // GetRackLabels ...
 func (dc *CassandraDatacenter) GetRackLabels(rackName string) map[string]string {
 	labels := dc.GetDatacenterLabels()
-	labels[RackLabel] = rackName
+	labels[RackLabel] = CleanLabelValue(rackName)
 	return labels
 }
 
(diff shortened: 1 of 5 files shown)


apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -555,23 +555,44 @@ func (dc *CassandraDatacenter) GetClusterLabels() map[string]string {
 	}
 }
 
-// apimachinery validation does not expose this, copied here
-const dns1035LabelFmt string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+// apimachinery validation does not expose these, copied here
+const (
+	dns1035LabelFmt     string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+	dns1123SubdomainFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+	labelFmt            string = "(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?"
```
adutra: Nit: rename to `dns1123LabelFmt` like the others above.

pkg/reconciliation/construct_statefulset.go
```diff
@@ -27,7 +27,7 @@ func newNamespacedNameForStatefulSet(
 	dc *api.CassandraDatacenter,
 	rackName string) types.NamespacedName {
 
-	name := api.CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.Name + "-" + rackName + "-sts"
+	name := api.CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.Name + "-" + api.CleanupSubdomain(rackName) + "-sts"
```
adutra, line 30: I think this could be simplified to:

```suggestion
	name := api.CleanupForKubernetes(dc.Spec.ClusterName + "-" + dc.Name + "-" + rackName + "-sts")
```

Because the real requirement is that `name` must conform to RFC 1035. IOW: if `dc.Spec.ClusterName` conforms to RFC 1035 and `rackName` conforms to RFC 1123, then the concatenation of both with a dash also conforms to RFC 1035.

pkg/reconciliation/construct_statefulset_test.go
```diff
@@ -450,6 +451,44 @@ func Test_newStatefulSetForCassandraPodSecurityContext(t *testing.T) {
 	}
 }
 
+func TestValidName(t *testing.T) {
```
adutra: These tests are good but if we have time to do it, it would be even better to have individual tests for `CleanupForKubernetes`, `CleanupSubdomain` and `CleanLabelValue` :-)

burmanm: I'll rename this to indicate it tests subdomain, but we do have those other tests. labels_test.go has one for CleanLabelValue and CleanupForKubernetes is tested in some other test as part of the process (but of course now I can't remember where - but at least Test_buildDefaultSuperuserSecret has a small test).

pkg/reconciliation/construct_statefulset.go
burmanm: Actually, this resulted in incorrect result:

        	Error:      	Not equal:
        	            	expected: "cluster1-dc1-invalid-pod-name-sts"
        	            	actual  : "cluster1-dc1-invalid-pod-namests"

adutra: I don't see why one dash would be removed and not the others, but OK.

adutra (approved):

pkg/reconciliation/construct_statefulset.go
burmanm: The 1035 does not allow dashes in every position (and not just the beginning). Not that I could directly say why that forth position from the end is not allowing it..

PR #413, opened 2022-09-22
Update Docker registry to new default

### LINKED ISSUE #414: Update default container registry to something other than Docker Hub (keep publishing images to Docker Hub)
**What is missing?**
Docker Hub has instituted pull limits for authenticated and anonymous users. This can lead to issues if a pull secret is not configured and used within Kubernetes.

**Why do we need it?**
Users may run into the issue of images not being pulled due to hitting their rate limit

* Cass Operator version:

Any

We should make the change to the major / minor branches in addition to master.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-35) by [Unito](https://www.unito.io)
┆Issue Number: CASS-35

bradfordcp:
**What this PR does**:

Updates the registry coordinates to refer to the new Google Artifact Registry
GHA writes images to us-docker.pkg.dev/k8ssandra/images
Tests and defaults refer to us-docker.pkg.dev/k8ssandra/images
Makefile updates for new repository locations

Fixes #414 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,9 +39,9 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # This variable is used to construct full image tags for bundle and catalog images.
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
-# cassandra.datastax.com/cass-oper-bundle:$VERSION and cassandra.datastax.com/cass-oper-catalog:$VERSION.
+# k8ssandra/cass-operator-bundle:$VERSION and k8ssandra/cass-operator-catalog:$VERSION.
 
-ORG ?= k8ssandra
+ORG ?= artifacts.k8ssandra.io/k8ssandra/images
 IMAGE_TAG_BASE ?= $(ORG)/cass-operator
 
 M_INTEG_DIR ?= all
@@ -161,7 +161,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	docker buildx build --build-arg VERSION=${VERSION} -t ${IMG} -t ${IMG_LATEST} . --load
+	docker buildx build --build-arg VERSION=${VERSION} -t ${IMG} -t k8ssandra/cass-operator:v${VERSION} -t ${IMG_LATEST} -t k8ssandra/cass-operator:latest . --load
 
 .PHONY: docker-kind
 docker-kind: docker-build ## Build docker image and load to kind cluster
@@ -172,15 +172,19 @@ docker-kind: docker-build ## Build docker image and load to kind cluster
 docker-push: ## Build and push docker image with the manager.
 	docker push ${IMG}
 	docker push ${IMG_LATEST}
+	docker push k8ssandra/cass-operator:v${VERSION}
+	docker push k8ssandra/cass-operator:latest
 
 .PHONY: docker-logger-build
 docker-logger-build: ## Build system-logger image.
-	docker buildx build -t ${LOG_IMG} -t ${LOG_IMG_LATEST} --build-arg VERSION=${VERSION} -f logger.Dockerfile . --load
+	docker buildx build -t ${LOG_IMG} -t ${LOG_IMG_LATEST} --build-arg VERSION=${VERSION} -t k8ssandra/system-logger:v${VERSION} -t k8ssandra/system-logger:latest -f logger.Dockerfile . --load
 
 .PHONY: docker-logger-push
 docker-logger-push: ## Push system-logger-image
 	docker push ${LOG_IMG}
 	docker push ${LOG_IMG_LATEST}
+	docker push k8ssandra/system-logger:v${VERSION}
+	docker push k8ssandra/system-logger:latest
 
 .PHONY: docker-logger-kind
 docker-logger-kind: docker-logger-build ## Build system-logger image and load to kind cluster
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -39,9 +39,9 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # This variable is used to construct full image tags for bundle and catalog images.
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
-# cassandra.datastax.com/cass-oper-bundle:$VERSION and cassandra.datastax.com/cass-oper-catalog:$VERSION.
+# k8ssandra/cass-operator-bundle:$VERSION and k8ssandra/cass-operator-catalog:$VERSION.
 
-ORG ?= k8ssandra
+ORG ?= us-docker.pkg.dev/k8ssandra/images
```
burmanm: This will break all local builds.

Makefile
```diff
@@ -161,7 +161,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	docker buildx build --build-arg VERSION=${VERSION} -t ${IMG} -t ${IMG_LATEST} . --load
+	docker buildx build --build-arg VERSION=${VERSION} -t ${IMG} -t k8ssandra/cass-operator:v${VERSION} -t ${IMG_LATEST} -t k8ssandra/cass-operator:latest . --load
```
burmanm, line 164: No hardcoding, can't override in scripts.

config/manager/image_config.yaml
```diff
@@ -3,7 +3,7 @@ kind: ImageConfig
 metadata:
   name: image-config
 images:
-  system-logger: "k8ssandra/system-logger:latest"
+  system-logger: "artifacts.k8ssandra.io/k8ssandra/images/system-logger:latest"
```
burmanm, line 6: Same here, won't work in local builds correctly.

Makefile
bradfordcp: Why? It is working for me locally. The presence of a domain only affects pulling and pushing. I can import this package into Kind without issue.

Makefile
bradfordcp: I'm hardcoding the docker hub images. My first attempt had two sets of variables, one for Docker Hub and the other for GCP AR, it looked cluttered. I can move back to that model if that's what's preferred, but given we hardcoded the bits for the RH repos I thought this would be ok.

What's your preference?

burmanm: We don't use this command for publishing to any repository (the GHA has different builder properties). This is just for local building. For that reason, the IMG_LATEST should be fine (since its values can be overridden)

Makefile
burmanm: It's only working for certain use-cases. See, now you're overriding the default repository - not just the org. If I want to use a localhost:5000 repository (local registry) - which I've setup for example, this won't work. The images now become: localhost:5000/us-docker.pkg.dev/k8ssandra/images - and this is invalid.

Same applies to the image_config.yaml. If you wish to override the default repository, use the ``imageRegistry`` field, otherwise everyone who uses their own local registry (not just localhost, but say a proxy also) will not be able to use that feature anymore. Our parsing does not understand that there's repository applied in the wrong field - it assumes the format for certain image structures is ``organization/image`` and the additional repository comes from a different property.

config/manager/image_config.yaml
burmanm: Also, if you wonder why to my previous question - the repository is different here than anything I would build. In the previous Makefile we build:

us-docker.pkg.dev/k8ssandra/images/system-logger:latest (which itself is problematic, but see another comment), but here we're calling ``artifacts.k8ssandra.io`` as prefix. So while your development env will popup, it's not actually loading the correct image, it's loading some historic one from the official repo, which is not what I'm actually developing.

bradfordcp: To be clear, `us-docker.pkg.dev` and `artifacts.k8ssandra.io` are the **same** repository. We want to _push_ all images to `us-docker.pkg.dev` and _pull_ from `artifacts.k8ssandra.io`.

Makefile
bradfordcp: Alright, I'm understanding the relationships in `image_config.yaml` but there's a bunch of overlap. Why are we baking in the `ORG` as well, shouldn't that be part of the registry definition?

So if I want to use `docker.io` why not just specify `imageRegistry: docker.io/k8ssandra` with the image name `system-logger:latest`?

There's another gotcha in that we don't host the DSE images within the `k8ssandra` org, there isn't a way for me to specify the `imageRegistry` for particular images and not others. Should we mirror the the images with management API to our org?

bradfordcp: For the `Makefile`, should we add a `REGISTRY` parameter that can be overridden?

config/manager/image_config.yaml
burmanm: But that's not the same string - thus if you push there and load from another, you will not load from say local cache (like kind).

burmanm:
I'm still confused with why we want to change images.go and other processing, or set the default image names to be different. Why won't adding them to GHA and just setting a new imageRegistry to image_config.yaml be enough? Isn't the point that users just download from artifacts.k8ssandra.io? All the other changes seem unnecessary.

bradfordcp:
> I'm still confused with why we want to change images.go...

At this point the only change in images.go is to set `DefaultCassandraRepository`. I agree we should set the `imageRegistry`, but there are some gotchas as we apply the registry to _all_ images. This would be a problem for users trying to use `datastax/dse-server` as we don't currently mirror that repository.

> All the other changes seem unnecessary

Does this statement include the changes to push packages to `us-docker.pkg.dev`?

burmanm:
> Does this statement include the changes to push packages to us-docker.pkg.dev?

I mean, only GHA needs to push there (that's the GHA part). But since Makefile's pushes are used only for local development work, there's no need to push with longer names there. 

One thing I obviously forgot is that the pre-release.sh needs the change to modify the cass-operator's deployment image to use the artifacts.k8ssandra.io/ .. path for the cass-operator image. 

The DSE image is more interesting - I didn't think about that. Are we intending to keep them on different repos?

PR #424, opened 2022-10-07
Allow modification of affinities

### LINKED ISSUE #351: K8SSAND-1578 ⁃ Make affinity fully configurable for each rack
**What is missing?**
cass-operator configures affinity, which includes node affinity and pod antiAffinity, for each statefulset based on the `allowMultipleNodesPerWorker` field and the `nodeAffinityLabels` for each rack. We should expose the fully Affinity property for each rack to give the user full control.

**Why do we need it?**
The existing behavior in cass-operator doesn't cover all uses cases. https://github.com/k8ssandra/k8ssandra-operator/issues/582 was created because someone wants to allow Cassandra pods from different C* clusters to be co-located on the same worker nodes. That is a perfectly reasonable request, but it isn't possible to do with cass-operator today.

**Environment**

* Cass Operator version:

    `v1.11.0`
    
┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1578) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1578
┆priority: Medium

burmanm:
**What this PR does**:
Allows defining custom affinity rules. Previously, all values set in Affinity in the PodTemplateSpec were ignored, now we merge them with our existing behavior. If allowMultipleWorkers is set, it's still possible to set custom rules now, if set to false, then our existing restrictions to anti-affinity are kept and custom ones are merged (and Kubernetes will ensure that all rules are adhered).

Also, remove rack.Zone, it's been deprecated for a very long time and the label it uses was deprecated in Kubernetes 1.15 with a replacement.

https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone

Fixes #351 

**Checklist**
- [ ] Changes manually tested


apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -312,10 +312,7 @@ type Rack struct {
 	// +kubebuilder:validation:MinLength=2
 	Name string `json:"name"`
 
-	// Deprecated. Use nodeAffinityLabels instead. Zone name to pin the rack, using node affinity
-	Zone string `json:"zone,omitempty"`
```
jsanda, line 316: We should a note in the release notes or change log explaining to users how to move away from `Zone` before upgrading.

pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -34,9 +34,9 @@ const (
 )
 
 // calculateNodeAffinity provides a way to decide where to schedule pods within a statefulset based on labels
-func calculateNodeAffinity(labels map[string]string) *corev1.NodeAffinity {
+func calculateNodeAffinity(labels map[string]string, existingNodeAffinity *corev1.NodeAffinity) *corev1.NodeAffinity {
```
jsanda: I don't think this will allow us to do things like:

```yaml
affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchExpressions:
          - key: cassandra.datastax.com/cluster
            operator: Exists
          - key: cassandra.datastax.com/datacenter
            operator: Exists
          - key: cassandra.datastax.com/rack
            operator: In
            values:
            - rack1
        topologyKey: kubernetes.io/hostname
```
This would allow us to pin all nodes from a rack to a single worker node. Additionally I think it would be confusing to configure affinity via PodTemplateSpec and with rack labels. I think it is easier for the user to be able to specify affinity per rack and also necessary to be able support use cases like the above.

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -192,9 +192,22 @@ func ValidateDatacenterFieldChanges(oldDc CassandraDatacenter, newDc CassandraDa
 				newRack.Name)
 		}
 		if oldRack.Zone != newRack.Zone {
-			return attemptedTo("change rack zone from '%s' to '%s'",
-				oldRack.Zone,
-				newRack.Zone)
+			if newRack.Zone != "" {
```
adejanovski, line 195: Shouldn't this be allowed only if affinity labels/rules have been defined to replicate the same behavior?
At least we may want to check that affinity was set for the rack, right?

pkg/reconciliation/construct_podtemplatespec_test.go
```diff
@@ -29,15 +29,100 @@ func init() {
...
+		if paa == nil ||
+			len(paa.RequiredDuringSchedulingIgnoredDuringExecution) != 1 {
+			t.Errorf("calculatePodAntiAffinity() = %v, and we want one element in RequiredDuringSchedulingIgnoredDuringExecution", paa)
+		}
+		assert.Equal(api.RackLabel, paa.RequiredDuringSchedulingIgnoredDuringExecution[0].LabelSelector.MatchExpressions[2].Key)
+		assert.Equal("rack1", paa.RequiredDuringSchedulingIgnoredDuringExecution[0].LabelSelector.MatchExpressions[2].Values[0])
+		assert.Equal("kubernetes.io/hostname", paa.RequiredDuringSchedulingIgnoredDuringExecution[0].TopologyKey)
+	})
+
 	t.Run("check when we do not allow more than one server pod per node", func(t *testing.T) {
```
adejanovski, line 103: That description doesn't seem accurate anymore.

adejanovski (commented):
Just a couple comments for now.
I'll run some tests to see if I understand correctly how these changes behave and if they work as expected.

adejanovski (approved):
It would be nice to be able to set the anti affinity rules at the dc level so that we don't have to repeat them in each rack when we want to allow multiple DC pods to be scheduled concurrently (same as `NodeAffinityLabels` which exist both at the DC and rack level).
This can be done in a follow up ticket though and shouldn't block this PR from being merged.

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
burmanm: There are multiple ways to achieve the same Zone behavior, so validating it is a bit complex.

burmanm:
Isn't that doable in the PodTemplateSpec? That should DC level property (and they're merged).

adejanovski:
> Isn't that doable in the PodTemplateSpec? That should DC level property (and they're merged).

Probably, I reckon I haven't tried.
Wouldn't it be cleaner to eventually take this out of the PodTemplateSpec?

burmanm:
Perhaps. Create a ticket? ;)

adejanovski:
> Perhaps. Create a ticket? ;)

will do 👍

PR #440, opened 2022-11-07
Restart task status.Succeeded fixes

### LINKED ISSUE #431: K8SSAND-1837 ⁃ Succeeded field in CassandraTask status does not get updated for restart task
**What happened?**
I created a CassandraTask to perform a rolling restart on a 3 node DC. The `.status.succeeded` field of the CassandraTask was never updated. I assume the same problem exists for `.status.failed` but the restarts were all successful so I can't say for certain.

`.status.succeeded` should get updated as per the API docs. I do see this behavior for other CasssandraTask types.

Create a 3-node CassandraDatacenter in a namespace named `test`:

```yaml
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: test
  serverType: cassandra
  serverVersion: 4.0.3
  size: 3
  storageConfig:
    cassandraDataVolumeClaimSpec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
      storageClassName: standard
```

Once the CassandraDatacenter is ready create a CassandraTask:

```yaml
kind: CassandraTask
metadata:
  name: restart-dc
spec:
  datacenter:
    name: dc1
    namespace: test
  jobs:
    - name: restart-dc1
      command: restart
```

**Environment**

* Cass Operator version:

    `v1.13.0`
    
    

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1837) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1837
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Ensure that the number of pods successfully restarted is tracked via the `UpdatedReplicas` status field.

Fixes #431 

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
One note, in our original discussion @burmanm suggested that we need to ensure that the revision has updated before setting `restartedPods = status.UpdatedReplicas`. 

I think this is actually already done [here](https://github.com/Miles-Garnsey/cass-operator/blob/bugfix/restart-task-status-fix/controllers/control/jobs.go#L81) so I haven't done anything further with that.

I also wonder if there is a chance that `restartedPods += int(status.UpdatedReplicas)` will overcount the number of pods restarting if that code runs twice while the restart is taking place (there is a requeue [here](https://github.com/Miles-Garnsey/cass-operator/blob/bugfix/restart-task-status-fix/controllers/control/jobs.go#L95)). 

We could feasibly use a map of sts->UpdatedReplicas to keep better track. I might experiment with implementing that and see if I can make it work.

controllers/control/jobs.go
```diff
@@ -69,7 +69,7 @@ func (r *CassandraTaskReconciler) restartSts(ctx context.Context, sts []appsv1.S
 			}
 		}
 	}
-
+	restartedPods := make(map[string]int)
```
burmanm: Why do you need map[string]int for calculating int?

controllers/control/jobs.go
```diff
@@ -84,9 +84,16 @@ func (r *CassandraTaskReconciler) restartSts(ctx context.Context, sts []appsv1.S
...
 				status.ObservedGeneration == st.GetObjectMeta().GetGeneration() {
 				// This one has been updated, move on to the next one
+
+				restartedPods[st.Name] = int(status.UpdatedReplicas)
 				continue
 			}
-
+			restartedPods[st.Name] = int(status.UpdatedReplicas)
+			totalRestarted := 0
+			for _, v := range restartedPods {
```
burmanm: This code is never called with finished StSes as they jump out of the loop before this is reached.

Miles-Garnsey: I know, but we want unfinished ones to be tracked too, don't we?

controllers/control/jobs.go
Miles-Garnsey: You're right, the way you originally suggested doing it is correct I think. I've changed it back, no need for the map.

controllers/control/jobs.go
burmanm: We do, but now we won't track the finished ones at all. We should update the TaskConfig.Completed inside that part of the loop also. Or use a pointer and update that one in TaskConfig..

Miles-Garnsey: Can you check my latest commit and confirm that's what you're after?

burmanm (approved):

burmanm:
.. just add missing changelog ;)

PR #458, opened 2022-11-21
Extract task execution attributes into CassandraTaskTemplate

### LINKED ISSUE #457: Extract task execution attributes into an inlined struct
**What is missing?**

In k8ssandra/k8ssandra-operator#739 we will introduce a `K8ssandraTask` resource, that triggers a `CassandraTask` for each datacenter in the cluster:
```yaml
kind: K8ssandraClusterTask
metadata:
  name: upgradesstables
spec:
  cluster:
    name: cluster1
  datacenters:
    - dc3
    - dc1
  spec:
    jobs:
      - name: upgradesstables-dc1
        command: upgradesstables
```
The `spec` part is almost identical to `CassandraTask`, except that it won't have a `datacenter` field. It would be nice to have a struct that we can reuse, instead of duplicating all the required fields.

**Why do we need it?**

Code reuse, maintainability.

**Environment**

Any.

* Cass Operator version: > 1.13.1

olim7t:
**What this PR does**:
Extract a structure that we can reuse in K8ssandra's upcoming `K8ssandraTask` API. It contains everything from `CassandraTask` except the datacenter.

This is almost transparent in cass-operator, the CRD doesn't change.

Fixes #457

**Checklist**
- [x] Changes manually tested


pkg/reconciliation/reconcile_racks.go
```diff
@@ -5,11 +5,11 @@ package reconciliation
 
 import (
 	"fmt"
+	"net"
 	"reflect"
 	"sort"
 	"strings"
 	"time"
-	"net"
```
olim7t, line 12: Unrelated, not sure why this happened (Go version maybe?)
I can revert if necessary.

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -1958,18 +1958,18 @@ func TestStartOneNodePerRack(t *testing.T) {
...
-					RpcAddress: "127.0.0.1",
-			},
-			{
-					HostID:     "2",
-					RpcAddress: "::1",
-			},
-			{
-					HostID:     "3",
-					RpcAddress: "2001:0DB8:0:0:8:800:200C:417A",
-			},
```
olim7t, line 1972: Same as above, unrelated.

pkg/reconciliation/reconcile_racks.go
burmanm: Yeah, go fmt in 1.19 has some changes.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 * [CHANGE] [#447](https://github.com/k8ssandra/cass-operator/issues/447) Update Github actions to remove all deprecated features (set-outputs, node v12 actions)
 * [CHANGE] [#448](https://github.com/k8ssandra/cass-operator/issues/448) Update to operator-sdk 1.25.1, update to go 1.19, update to Kubernetes 1.25, remove amd64 restriction on local builds (cass-operator and system-logger will be built for aarch64 also)
 * [FEATURE] [#441](https://github.com/k8ssandra/cass-operator/issues/441) Implement a CassandraTask for moving single-token nodes
+* [ENHANCEMENT] [#457](https://github.com/k8ssandra/cass-operator/issues/457) Extract task execution attributes into CassandraTaskTemplate
```
burmanm: Since this is a API change, I think "CHANGE" would be better in this case. It's not really improving stuff for end users, if someone is writing their own client for example (other than us).

burmanm:
Other than that, this is fine - as the YAML interface stays intact. However, I doubt this is enough in the long run. Won't the k8ssandra-operator need all the execution logic also? That is, the TTL handling, job scheduling (without the pod/StS etc targetting of course) to make it stay equivalent to cass-operator's behavior?

olim7t:
> Won't the k8ssandra-operator need all the execution logic also? That is, the TTL handling, job scheduling (without the pod/StS etc targetting of course) to make it stay equivalent to cass-operator's behavior?

My assumption is that a `K8ssandraTask` could simply create a `CassandraTask` in each of its target datacenters, and then monitor their status. Similar to the relationship between `K8ssandraCluster` and `CassandraDatacenter`.
In that case we wouldn't need most of the task execution logic, because the `K8ssandraTask` would only react to the `CassandraTask`s' state changes.

burmanm:
> My assumption is that a K8ssandraTask could simply create a CassandraTask in each of its target datacenters, and then monitor their status. Similar to the relationship between K8ssandraCluster and CassandraDatacenter.
In that case we wouldn't need most of the task execution logic, because the K8ssandraTask would only react to the CassandraTasks' state changes.

That's only true if the execution can happen in parallel - and even then you need to track (which means keeping up state machine). But as far as I understood, it should happen in serial, in which case you will need reconciliation logic to watch and only then move things forward. Also, if you intend to implement stuff like TTL or retries, it's again reconcile logic, similar to what cass-operator already does.

burmanm (approved):

burmanm (approved):

PR #473, opened 2023-01-04
Add POD_NAME and NODE_NAME env variables to cassandra container

### LINKED ISSUE #472: Add pod's name and Kubernetes node name as env variables
**What is missing?**
Like shown in the page https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/, we might want to have the metadata.name (which could differ from hostname in case the resolve options are changed for pods) as well as the physical Kubernetes node name the pod is running in.

** Why do we need it?**
These might be needed in certain queries with metrics.

**Environment**

* Cass Operator version:

burmanm:
**What this PR does**:
Adds POD_NAME and NODE_NAME env variables to cassandra container

Fixes #472 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/kindIntegTest.yml
+++ b/.github/workflows/kindIntegTest.yml
@@ -97,7 +97,7 @@ jobs:
         - webhook_validation
         # Three worker tests:
         - canary_upgrade
-        - config_change_condition
+        # - config_change_condition # config_change takes care of testing the same
         - cdc_successful # OSS only
         # - delete_node_lost_readiness # DSE specific behavior
         - host_network
--- a/.github/workflows/workflow-integration-tests.yaml
+++ b/.github/workflows/workflow-integration-tests.yaml
@@ -185,7 +185,7 @@ jobs:
         - webhook_validation
         # Three worker tests:
         # - canary_upgrade # See kind_40_tests job
-        - config_change_condition
+        # - config_change_condition # config_change takes care of the same testing
         # - cdc_successful # CDC is OSS only , see kind_311_tests and kind_40_tests jobs
         # - delete_node_lost_readiness # DSE specific behavior see kind_dse_tests job
         - host_network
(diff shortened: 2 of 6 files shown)


adejanovski (commented):
Looks good, but maybe a tiny test could secure the feature?
Could one of the existing tests be modified to check the presence of the variables and their values?

burmanm:
I added a unit test, the same way that initContainer EnvVars are checked.

.github/workflows/kindIntegTest.yml
```diff
@@ -97,7 +97,7 @@ jobs:
         - webhook_validation
         # Three worker tests:
         - canary_upgrade
-        - config_change_condition
+        # - config_change_condition # config_change takes care of testing the same
```
adejanovski, line 100: Maybe you could remove the line altogether and the corresponding test file?
Is there a reason to keep them?

.github/workflows/workflow-integration-tests.yaml
```diff
@@ -185,7 +185,7 @@ jobs:
         - webhook_validation
         # Three worker tests:
         # - canary_upgrade # See kind_40_tests job
-        - config_change_condition
+        # - config_change_condition # config_change takes care of the same testing
```
adejanovski, line 188: Same question here.

adejanovski (approved):
Looks good, pending passing tests.

I'd additionally recommend to clean up the config_change_condition test and remove traces of it in the codebase.

.github/workflows/kindIntegTest.yml
burmanm: I think that should be done in a separate PR, since at the moment there's lots of those tests we've merged together. I think all of those should be cleaned up in a single PR and that way we would have a single commit which could then be used to track "old tests" if ever necessary.

PR #475, opened 2023-01-11
Allow DC names override

### LINKED ISSUE #362: K8SSAND-1636 ⁃ Allow the Cassandra DC name to be different than the CassandraDatacenter name

adejanovski:
**What this PR does**:
This PR introduces a new field in the CassandraDatacenter CRD which allows to set a datacenter name override.
This override will be used to name kubernetes objects that cass-operator will create based on the cassdc CR, and it will also be used as datacenter name in Cassandra, replacing the cassdc CR `.metadata.name`.

Fixes #362 

**Checklist**
- [x] Changes manually tested


apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -544,7 +547,7 @@ func (dc *CassandraDatacenter) SetCondition(condition DatacenterCondition) {
 // GetDatacenterLabels ...
 func (dc *CassandraDatacenter) GetDatacenterLabels() map[string]string {
 	labels := dc.GetClusterLabels()
-	labels[DatacenterLabel] = CleanLabelValue(dc.Name)
+	labels[DatacenterLabel] = dc.SanitizedName()
```
burmanm, line 559: The SanitizedName() returns the wrong name. In the description it says:

``Kubernetes objects will be named after metadata.name but in Cassandra the name will be overridden by this value.``

Yet, dc.SanitizedName() returns the overridden value by default (since it calls in.CassDcName() while it should do dc.Name), not the metadata.name and thus all the objects are created with the override value.

config/crd/bases/control.k8ssandra.io_cassandratasks.yaml
```diff
@@ -1,9 +1,10 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.6.1
```
burmanm: You need to regenerate all these, as they're using the wrong controller-gen version.

docs/user/README.md
```diff
@@ -200,6 +200,25 @@ cass-operator. Within a single Kubernetes namespace:
...
+  name: cluster1-dc1
+spec:
+  clusterName: cluster1
+  datacenterName: dc1
+  serverType: cassandra
+  serverVersion: 3.11.7
+  ...
+```
+
+Objects such as statefulsets, pods, secrets will be named and labeled after the `spec.datacenterName` value. This value can contain characters which aren't compliant with Kubernetes naming rules (such as capital letters or underscores) and will be sanitized when used in object names/labels. For example, `Datacenter_1` will be sanitized to `datacenter-1`.
```
burmanm, line 220: We should not do this. The name of the Kubernetes objects and all the labels should follow the name of the ``CassandraDatacenter``.

For a very simple reason also, imagine I create two CassandraDatacenters:

```yaml
kind: CassandraDatacenter
metadata:
  name: dc1
```

```yaml
kind: CassandraDatacenter
metadata:
  name: cluster1-dc1
spec:
  datacenterName: dc1
```

Kaboom! Now these two CassandraDatacenters can't suddenly be queried and they will overlap since both have ``dc1`` everywhere. They need to follow the name of the CassDc to avoid any conflicts.

burmanm (changes requested):

apis/cassandra/v1beta1/cassandradatacenter_types.go
adejanovski: My comment on ` DatacenterName` is wrong actually, I need to update it.
I'd want the behavior of the DC name override to be consistent with the cluster name override in k8ssandra-operator, and it is how what we decided to do there. All objects are named according to the override.

config/crd/bases/control.k8ssandra.io_cassandratasks.yaml
adejanovski: sounds good 👍

adejanovski:
@burmanm, I've addressed the points we discussed together:

- Add a check preventing the modification of the dc name override in the webhook (and the associated test)
- Add a test for a CassandraTask running with a dc name override
- Add checks in handler.go to detect conflicts with other cassdc names/overrides
- Add a check to prevent changing the dc name override after the dc was created. This one uses a new status field which allows to store the initial value of the dc name override and perform the comparison against in case of change.

We should be good for another review.

tests/util/ginkgo/lib.go
```diff
@@ -389,28 +390,32 @@ func (ns *NsWrapper) WaitForDatacenterReadyPodCountWithTimeout(dcName string, co
...
+func (ns *NsWrapper) WaitForDatacenterReadyWithOverride(dcName, dcNameOverride string) {
+	ns.WaitForDatacenterReadyWithTimeouts(dcName, dcNameOverride, 1200, 1200)
 }
 
 func (ns *NsWrapper) Log(step string) {
 	ginkgo.By(step)
 }
 
-func (ns *NsWrapper) WaitForDatacenterReadyWithTimeouts(dcName string, podCountTimeout int, dcReadyTimeout int) {
+func (ns *NsWrapper) WaitForDatacenterReadyWithTimeouts(dcName string, dcNameOverride string, podCountTimeout int, dcReadyTimeout int) {
```
burmanm: Instead of replicating that dcNameOverride parameter everywhere, wouldn't it have made more sense to get the overrideName from that object we just fetched?

adejanovski: Of course!

brilliant idea 👍

docs/user/README.md
adejanovski: I've added checks for this now. We can detect any conflict with another cassdc.

burmanm (approved):

PR #478, opened 2023-01-18
Enable CDC on DSE clusters

### LINKED ISSUE #476: Enable CDC agent support for DSE
**What happened?**
Attempting to enable CDC with DSE images fails. Currently, the code that enabled CDC assume that:
1) MCAC is always present and should be enabled via Java Agent (DSE has its own built-in MCAC and does not need the agent line in the startup config)
2) The Management API agent jarfile name is the same regardless of server version (it's currently different for DSE, but this should be resolved once this [PR](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/243) is merged and released)

The code that would need to change currently lives [here](https://github.com/k8ssandra/cass-operator/blob/master/pkg/cdc/generate_config.go#L94-L96).

**NOTE**: The fix for this issue should wait for the [Management API PR](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/243) to be merged and released.

Miles-Garnsey:
**What this PR does**:

Ensures MCAC doesn't get enabled for DSE clusters.
Ensures the management API images come from the management API docker repository. These contain the CDC agent.

Fixes #476 

**Checklist**
- [ ] Changes manually tested


pkg/cdc/generate_config.go
```diff
@@ -91,8 +91,11 @@ func updateAdditionalJVMOpts(optsSlice []string, CDCConfig *cassdcapi.CDCConfigu
 		}
 	}
 	CDCOpt := fmt.Sprintf("-javaagent:%s=%s", "/opt/cdc_agent/cdc-agent.jar", strings.Join(optsSlice, ","))
+	// We want to disable MCAC when the server being deployed is DSE
+	if cassDC.Spec.ServerType == "cassandra" {
+		out = append(out, "-javaagent:/opt/metrics-collector/lib/datastax-mcac-agent.jar")
+	}
 	out = append(out, // We need to add these two elements to the slice first, because the management agent must start before the CDC agent.
-		"-javaagent:/opt/metrics-collector/lib/datastax-mcac-agent.jar",
 		"-javaagent:/opt/management-api/datastax-mgmtapi-agent-0.1.0-SNAPSHOT.jar",
```
emerkle826: With the latest Management API images, the MAAC agent jarfile can be specified as simply:
```
/opt/management-api/datastax-mgmtapi-agent.jar
```

emerkle826 (commented):
I'll leave approval to @burmanm, but I did have the one comment.

pkg/cdc/generate_config.go
Miles-Garnsey: Oh sorry, I forgot that bit! Fixing.

README.md
```diff
@@ -158,7 +158,7 @@ defaults:
   cassandra:
     repository: "k8ssandra/cass-management-api"
   dse:
-    repository: "datastax/dse-server"
+    repository: "datastax/dse-mgmtapi-6_8"
```
burmanm: This requires upgrade test to verify that existing clusters do not break, lets say 6.8.9 installations.

docs/ingress/kong/sni-ingress/README.md
```diff
@@ -14,7 +14,7 @@ When leveraging a single endpoint ingress / load balancer we lose the ability to
     k3d load image --cluster k3s-default \
       datastax/cass-operator:1.3.0 \
       datastax/cass-config-builder:1.0-ubi7 \
-      datastax/dse-server:6.8.4-ubi7 \
+      datastax/dse-mgmtapi-6_8:6.8.4-ubi7 \
```
burmanm: This image does not exists.

burmanm:
cdc_successful_test must be modified to test with DSE also (and updated in the .github configs).

docs/ingress/kong/sni-ingress/README.md
Miles-Garnsey: I updated this so we're talking about a recent version. The UBI will exist tomorrow :)

burmanm:
And of course CHANGELOG ;)

Miles-Garnsey:
Thanks @burmanm, I've made the requested changes.

burmanm:
No, not yet. The same integration test must work in DSE & OSS (just like we do for most tests), not separate one. There's a method that takes input the version and that should be used. No separate testfiles needed or maintaining of multiple tests per version.

Miles-Garnsey:
> No, not yet. The same integration test must work in DSE & OSS (just like we do for most tests), not separate one. There's a method that takes input the version and that should be used. No separate testfiles needed or maintaining of multiple tests per version.

I don't know about that. The problem is, we need to delete Pulsar to ensure the two tests don't interfere with each other. Otherwise Pulsar will still have the data from the previous test which might result in inaccurate test results. 

Also, if we do the tests serially, they will be very slow due to Pulsar taking a while to roll out (although image pulls might be reduced for Pulsar components, which could make things faster.. IDK.)

burmanm:
> I don't know about that. The problem is, we need to delete Pulsar to ensure the two tests don't interfere with each other. Otherwise Pulsar will still have the data from the previous test which might result in inaccurate test results.

This is not what I meant. The test code must be the same (cdc_successful_test), the execution in different VM is done by the GHA. Just like all the other tests behave, there's a separate code that modifies the input file to match the correct serverVersion and serverType.

Miles-Garnsey:
> This is not what I meant. The test code must be the same (cdc_successful_test), the execution in different VM is done by the GHA. Just like all the other tests behave, there's a separate code that modifies the input file to match the correct serverVersion and serverType.

I'm not following what you mean sorry. The parallelisation happens at the level of the test name doesn't it? And that means that we need to have two test files? Or am I wrong?

Maybe you can show me an example of the way you'd like it structured from another test.

burmanm:
> I'm not following what you mean sorry. The parallelisation happens at the level of the test name doesn't it? And that means that we need to have two test files? Or am I wrong?

You are, the parallelisation between different settings happens in the GHA. Most of the other tests are all working this way, just look at for example https://github.com/k8ssandra/cass-operator/blob/master/tests/test_all_the_things/test_all_the_things_suite_test.go#L62

That method will return the deployment file as DSE / Cass (3.11/4.0.x/4.1.x ..). Those are then different tests to be run.

burmanm:
And separate note, this PR can't be merged as is, the image_config.yaml can't change in the master branch. Otherwise it will break clusters that are running older versions that our images do not support.

emerkle826:
@Miles-Garnsey I wanted to try to clarify a few things about the tests.

You added a new test, `cdc_successful_dse`, but if you look at the GitHub Actions test run [here](https://github.com/k8ssandra/cass-operator/pull/478/checks), you'll see that it actually ran using Cassandra 4.0.7, not DSE.

It looks like you cloned the existing `cdc_successful` test to make this new test, and in doing so, you copied the code that dynamically changes the test fixtures to specify the server version that is run in the GitHub Action matrix. The utility function is [CreateTestFile](https://github.com/k8ssandra/cass-operator/blob/a1b29e9fbc7a006bde0d22e7146472f82a67af17/tests/util/ginkgo/lib.go#L43), and it will modify the test fixture yaml file passed in based on environment variables set when the tests runs.

The environment variables that can be set are [M_SERVER_VERSION](https://github.com/k8ssandra/cass-operator/blob/a1b29e9fbc7a006bde0d22e7146472f82a67af17/tests/util/ginkgo/lib.go#L66-L69), [M_SERVER_TYPE](https://github.com/k8ssandra/cass-operator/blob/a1b29e9fbc7a006bde0d22e7146472f82a67af17/tests/util/ginkgo/lib.go#L71-L74) and [M_SERVER_IMAGE](https://github.com/k8ssandra/cass-operator/blob/a1b29e9fbc7a006bde0d22e7146472f82a67af17/tests/util/ginkgo/lib.go#L61-L64). If any of these are set when the test runs, calls to [CreateTestFile](https://github.com/k8ssandra/cass-operator/blob/a1b29e9fbc7a006bde0d22e7146472f82a67af17/tests/util/ginkgo/lib.go#L43) will replace the test fixture data with  those overrides. So you don't need to add that test, you need to put it in one of the GitHub Actions jobs that has DSE in the [test matrix](https://github.com/k8ssandra/cass-operator/blob/a1b29e9fbc7a006bde0d22e7146472f82a67af17/.github/workflows/kindIntegTest.yml#L158-L164). That said, there isn't a spot where you can add `cdc_successful` as it should only run against Cassandra 4 and DSE. So you might need to create a new job with a matrix that specifies only those two server versions and only the `cdc_successful` integration_test.

Also, to clarify a little of what @burmanm is talking about:
> the image_config.yaml can't change in the master branch. Otherwise it will break clusters that are running older versions that our images do not support.

The main issue here is that changing the default DSE image repo from `datastax/dse-server` to `datastax/dse-mgmtapi-6_8` will break for anyone currently using DSE versions older than 6.8.25. That is because the `datastax/dse-mgmtapi-6_8` repo does not have any DSE images for versions older than 6.8.25. If someone is currently using cass-operator with DSE 6.8.9, upgrading with this change will fail to find that image since it doesn't exist.

A possible solution to this last point would be to publish older DSE images (all that cass-operator supports) to the `datastax/dse-mgmtapi-6_8` repo. But I imagine that would be every DSE version from 6.8.25 back to 6.8.9. So a lot of images to publish. And I don't think it would be feasible to effectively test each of those versions to make sure things work as they currnetly do with the `datastax/dse-server` repo images.

Miles-Garnsey:
@emerkle826 and I have had a chat about this and he's walked me through some of your comments @burmanm.

**Test versions/serverTypes**

@emerkle826, thanks for explaining this, it looks like a bunch of changes were made a few months ago and the CDC test no longer applies the manifests verbatim. I'll do as you suggested and update the workflow instead of duplicating the tests.

> And separate note, this PR can't be merged as is, the image_config.yaml can't change in the master branch. Otherwise it will break clusters that are running older versions that our images do not support.

**Image registry change**

Erik has provided some explanation of this above. It seems that the issue is that users may not have the image set, and (given we only publish 6.8.25+ via the management api image registry) a customer who has no image set but has the serverVersion set at e.g. 6.8.5 would experience problems due to the change in repo that we're implementing.

There are two solutions to this problem in my view:

1. Make a note in the changelog that users on versions < 6.8.25 need to explicitly set their image coordinates (or ImageConfig) prior to upgrade.
2. Publish images all the way back to 6.8.9 (or whatever the earliest version cass-operator's next release supports, I can't even find this information) in the management API registry so that the move is transparent to users. We'd also likely need to develop tests to ensure that all of these images work as expected.

I think that there is a consensus that we need to be building our own Docker images for DSE if we're going to be able to implement features like this.

adejanovski:
> * Make a note in the changelog that users on versions < 6.8.25 need to explicitly set their image coordinates (or ImageConfig) prior to upgrade.
> * Publish images all the way back to 6.8.9 (or whatever the earliest version cass-operator's next release supports, I can't even find this information) in the management API registry so that the move is transparent to users. We'd also likely need to develop tests to ensure that all of these images work as expected.

I'm not crazy about the former, we can't assume that folks will read the release notes before upgrading.
The latter would have my preference, but don't we have alternative solutions?
If we managed to implement something like "before 6.8.25 use datastax/dse-server and after use k8ssandra/dse-mgmt...." it would allow a smooth transition without the need for users to do something specific or us having to rebuild a whole lot of images.
@burmanm, do you think that would be feasible without spending too much time on it?

burmanm:
> If we managed to implement something like "before 6.8.25 use datastax/dse-server and after use k8ssandra/dse-mgmt...." it would allow a smooth transition without the need for users to do something specific or us having to rebuild a whole lot of images.

That can be done in the images.go, but I'd say it should almost be 6.8.31 only automatic, etc. Whenever we do the decided cut. Until then users might have deployed clusters with older versions, which we would then have to test for compatibility.

In any case, not for next release (1.14.0).

emerkle826:
We do now have Management API images for DSE based on RedHat UBI images:
https://hub.docker.com/r/datastax/dse-mgmtapi-6_8/tags

burmanm:
As discussed today, the approach we will take is documented in https://github.com/k8ssandra/cass-operator/issues/479. This will be a separate PR (target release 1.15.0). From this PR, the image_config.yaml as well as image.go changes should be removed.

adejanovski:
> As discussed today, the approach we will take is documented in https://github.com/k8ssandra/cass-operator/issues/479. This will be a separate PR (target release 1.15.0). From this PR, the image_config.yaml as well as image.go changes should be removed.

+1, let's do this

config/crd/bases/control.k8ssandra.io_cassandratasks.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.8.0
```
burmanm: Wrong controller-gen

pkg/cdc/generate_config.go
```diff
@@ -91,9 +91,12 @@ func updateAdditionalJVMOpts(optsSlice []string, CDCConfig *cassdcapi.CDCConfigu
 		}
 	}
 	CDCOpt := fmt.Sprintf("-javaagent:%s=%s", "/opt/cdc_agent/cdc-agent.jar", strings.Join(optsSlice, ","))
+	// We want to disable MCAC when the server being deployed is DSE
```
burmanm, line 94: We also want to remove MCAC agent when MGMT_API_DISABLE_MCAC is set (like done in https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/scripts/docker-entrypoint.sh#L43)

Does this process mess up with that logic?

.github/workflows/kindIntegTest.yml
```diff
@@ -53,9 +53,9 @@ jobs:
     strategy:
       matrix:
         integration_test:
-        - upgrade_operator # Test is not setup to run against 4.0
-        - additional_seeds #TODO: Fails against C* 4.0, fix in https://github.com/k8ssandra/cass-operator/issues/459
-        - scale_down_unbalanced_racks  #TODO: Fails against C* 4.0 and DSE 6.8, fix in https://github.com/k8ssandra/cass-operator/issues/459
+          - upgrade_operator # Test is not setup to run against 4.0
```
burmanm, line 56: What's up with these mass-reindents?

Miles-Garnsey: Indentation has been pretty inconsistent so I've turned on `Trim Trailing Whitespace` and `Yaml › Format: Enable` in vs code, I think everyone should.

pkg/cdc/generate_config.go
Miles-Garnsey: Maybe, but that's a problem to solve under that PR, not this one IMO. That one is still under development anyway.

Miles-Garnsey: Sorry, I meant [this](https://github.com/k8ssandra/k8ssandra-operator/pull/812) PR.

config/crd/bases/control.k8ssandra.io_cassandratasks.yaml
Miles-Garnsey: Your PR updated this to v0.10.0 dude: https://github.com/k8ssandra/cass-operator/pull/452/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R247

pkg/cdc/generate_config.go
burmanm: I created #481 to track this. We need it also.

burmanm: (separate PR)

config/crd/bases/control.k8ssandra.io_cassandratasks.yaml
burmanm: Yes. It's supposed to be 0.10.0 (that PR was done 3 months ago). You're recreating these with 0.8.0.

Miles-Garnsey: Oh, right, sorry! Fixed.

Those files get generated via kubebuilder don't they? I ran make controller-gen because I know that is local to the repo, but we use a system level kubebuilder in this repo don't we? Maybe we should be using a local one to avoid this problem.

adejanovski: it's because the make targets use a local version of controller-gen which is under `./bin` and is only re-downloaded if it's not already in that folder. You may have had an older version there, which can be fixed by just deleting the binary and re-running the make target.

Miles-Garnsey: Yeah, it turns out I'd only updated the one in k8ssandra-operator, not cass-operator. My bad, apologies.

.github/workflows/workflow-integration-tests.yaml
```diff
@@ -299,7 +299,7 @@ jobs:
       - name: set environment for DSE
         if: ${{ startsWith( matrix.version, '6.8' ) }}
         run: |
-          echo "M_SERVER_IMAGE=datastax/dse-mgmtapi-6_8:${{ matrix.version }}-jdk8" >> $GITHUB_ENV
+          echo "M_SERVER_IMAGE=datastax/dse-server:${{ matrix.version }}-jdk8" >> $GITHUB_ENV
```
burmanm, line 302: .. this isn't right. It should be in the mgmtapi-6_8 still.

burmanm (approved):

.github/workflows/kindIntegTest.yml
burmanm: This also reindented all our CRDs. And those are now being regenerated by the make manifests. So we have 10k line commits doing reindents and I have no idea if anything was changed in them (I just assumed they were re-created after changing the controller-gen version). We can't reindent stuff with these tools if they don't understand what's generated.

PR #487, opened 2023-01-25
AdditionalVolumes to allow VolumeSource

### LINKED ISSUE #467: Allow configuration of metrics server in mgmt-api
**What is missing?**
Once the new /metrics work is done, we need the ability to allow easy configuration of filtering, ports and other properties of the /metrics server. This needs to be injected to the correct place in the deployment. 

Either we extend the configuration of the operator to allow setting these (like ImageConfig) or we add a ConfigMap that's mapped (if it exists).

**Why do we need it?**

**Environment**

* Cass Operator version:

### LINKED ISSUE #486: Add the ability to mount AdditionalVolumes using VolumeSource
**What is missing?**
Currently, our AdditionalVolumes allows mounting only PVCSpecs to the ``cassandra`` container. We should add the ability to mount other VolumeSources also, such as ConfigMaps, Secrets, etc

**Why do we need it?**
This makes configuring for example the metrics-server (#467) easier.

**Environment**

* Cass Operator version:

burmanm:
**What this PR does**:
Adds the abiliity to make AdditionalVolumes take VolumeSource as input instead of only PVCSpec.

Breaks the API by making PVCSpec a pointer to struct instead of struct..

And allows the modification of AdditionalVolumes in the webhook

Fixes #467 
Fixes #486 

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
> Breaks the API by making PVCSpec a pointer to struct instead of struct..

This shouldn't be a problem for the yaml API right, because it just means that the `PVCSpec` becomes optional in the yaml? And a required -> optional change is always backwards compatible..?

I am a little more concerned about the impact it might have on the k8ssandra-operator side of things, since the move from a value -> pointer type will indeed break things (e.g. [here](https://github.com/k8ssandra/k8ssandra-operator/blob/feature/metrics-config/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L453)). Do you have any suggestions on how to handle that? 

I don't get the sense that we have many users accessing the APIs via the Go types, and given this is a beta API I don't think we have any obligation to maintain exact backwards compatibility even if we did (based on [this](https://kubernetes.io/docs/reference/using-api/)). However, I think we should probably consider bumping the version to v1beta2 if we're going to make a breaking change, since this will signal the breaking change to users.

Finally, is there a reason that this field is better than accessing the podTemplateSpec directly? My understanding was that making these changes via PodTemplateSpec was becoming more accepted around the team.

burmanm:
> This shouldn't be a problem for the yaml API right, because it just means that the PVCSpec becomes optional in the yaml? And a required -> optional change is always backwards compatible..?

It does not break YAML, that's true.

> I am a little more concerned about the impact it might have on the k8ssandra-operator side of things, since the move from a value -> pointer type will indeed break things (e.g. [here](https://github.com/k8ssandra/k8ssandra-operator/blob/feature/metrics-config/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L453)). Do you have any suggestions on how to handle that?

I have no idea why AdditionalVolumes is called "PVCs" in the K8ssandraCluster. That makes no sense. But why would it break that? The AdditionalVolumes struct hasn't changed type. 

> However, I think we should probably consider bumping the version to v1beta2 if we're going to make a breaking change, since this will signal the breaking change to users.

We haven't traditionally bumped to v1beta2 with other breaking changes either. That would require modifying webhook because that would change the YAML format also. And since YAML isn't breaking in this case, I don't think there's a reason to bump it to v1beta2. It only affects the Go API, but changing its package is not probably a very meaningful thing to do.

> Finally, is there a reason that this field is better than accessing the podTemplateSpec directly? My understanding was that making these changes via PodTemplateSpec was becoming more accepted around the team.

cass-operator does not provide any PodTemplateSpec guarantees or backwards compatibility. We can and do break it (it's not passed as is and has never been) and have never recommended it. Configuring this type of Volume through PodTemplateSpec is cumbersome for users (requiring both volumes and volumesources). For YAML users it's often just a "hack" to get a missing feature.

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -150,9 +154,16 @@ func ValidateDatacenterFieldChanges(oldDc CassandraDatacenter, newDc CassandraDa
 		return attemptedTo("change serviceAccount")
 	}
 
-	// StorageConfig changes are disallowed
-	if !reflect.DeepEqual(oldDc.Spec.StorageConfig, newDc.Spec.StorageConfig) {
-		return attemptedTo("change storageConfig")
+	// CassandraDataVolumeClaimSpec changes are disallowed
+
+	if !reflect.DeepEqual(oldDc.Spec.StorageConfig.CassandraDataVolumeClaimSpec, newDc.Spec.StorageConfig.CassandraDataVolumeClaimSpec) {
```
adutra, line 159: There is a semantic change here, not sure if this is OK. I can now change `MountPath` or `Name`whereas this was previously forbidden. I can also change `VolumeSource`, again not sure if this is intended.

burmanm: Well, I did think about this, but opted not to prevent it. When AdditionalVolumes are changed, the StatefulSets are recreated and these are allowed values to be changed in the StS (unlike PVC modifications).

adejanovski: agreed, we should allow to remove such volumes 👍

adutra (approved):

PR #489, opened 2023-01-27
Expose the new metrics port in the all-pods service

### LINKED ISSUE #488: Expose the new metrics port in services
**What happened?**

In https://github.com/k8ssandra/cass-operator/issues/329 we exposed a new container port, 9000, for the new metrics endpoint. But we also need to expose the new port through the all-pods service in order for Prometheus to be able to scrape metrics when a ServiceMonitor is created for the DC.

A port 9000 should be exposed in all services that are already exposing the (legacy) `prometheus` port (9143).

adutra:
**What this PR does**:
Exposes the 9000 port in 2 services: DC and all-pods.

Fixes #488 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/construct_service_test.go
+++ b/pkg/reconciliation/construct_service_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	api "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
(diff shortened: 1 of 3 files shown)


pkg/reconciliation/construct_service_test.go
```diff
@@ -407,3 +408,99 @@ func TestAddingAdditionalLabels(t *testing.T) {
...
+			name: "Cassandra 3.11.14",
+			dc: &api.CassandraDatacenter{
+				Spec: api.CassandraDatacenterSpec{
+					ClusterName:   "bob",
+					ServerType:    "cassandra",
+					ServerVersion: "3.11.14",
+				},
+			},
+			dcServicePorts:      []int32{8080, 9000, 9042, 9103, 9142, 9160},
+			allPodsServicePorts: []int32{8080, 9000, 9042, 9103},
```
adutra, line 429: I wonder: shouldn't all-pods service expose the same ports as the dc service? But I didn't want to change that in this PR.

pkg/reconciliation/construct_service_test.go
```diff
@@ -407,3 +408,99 @@ func TestAddingAdditionalLabels(t *testing.T) {
...
+											ContainerPort: 9004,
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			// FIXME: 9004 should be in the list of open ports
```
adutra, line 479: This is interesting: I copied this test from `TestPorts`, but I realized that if the user changes a managed port at container level, the change is not reflected at service level.

pkg/reconciliation/construct_service_test.go
adejanovski: it depends what the all-ports service is used for.
If no one expressed any requirement around that, I agree we shouldn't change it.

adejanovski (approved):

pkg/reconciliation/construct_service_test.go
adutra: From what I understand the all-pods service is identical to the dc service, but also includes pods not ready. So it's not suitable to be used as a contact point for the drivers, but it is useful for metrics for instance. Maybe that's the reason why it doesn't expose so many ports. That said, following this logic it shouldn't expose 9042 either. Anyways 🤷‍♂️

burmanm: cass-operator sets allPods as the ServiceName for all pods to ensure that they get DNS working correctly. Using only the dcPods would make the DNS names fail in certain cases:

https://github.com/k8ssandra/cass-operator/pull/18

PR #491, opened 2023-02-08
Envtesting...

### LINKED ISSUE #563: Ability to test basic scenarios with envtests
### What is missing?

There's no way to run CassandraDatacenter controller in the cass-operator inside envtests.

This ticket probably existed inside some other ticket, but has since been closed. 

### Why is this needed?

Finding out if the operator works at all through the e2e tests is a long wait. We need to have some basic smoke tests available in the envtesting form to speed up the feedback loop when making changes.

burmanm:
**What this PR does**:
Adds the ability to run envtests against CassandraDatacenter controller.

Fixes #563 

**Checklist**
- [ ] Changes manually tested


internal/controllers/cassandra/cassandradatacenter_controller.go
```diff
@@ -117,7 +122,7 @@ func (r *CassandraDatacenterReconciler) Reconcile(ctx context.Context, request c
 
 		// Error reading the object
 		logger.Error(err, "Failed to get CassandraDatacenter.")
-		return ctrl.Result{RequeueAfter: 10 * time.Second}, err
+		return ctrl.Result{}, err
```
adejanovski, line 125: **Question:** why was this changed?

adejanovski: Oh, maybe to support shorter requeue times for the envtests?

adejanovski (approved):

internal/controllers/cassandra/cassandradatacenter_controller.go
burmanm: Kinda, but not really. It's because returning non-nil err will cause the requeue to happen in controller-runtime, so we should never set Requeue time in that case.

PR #497, opened 2023-02-20
Allow disable mcac even when CDC enabled

### LINKED ISSUE #481: Enabling CDC messes up mgmt-api configuration
**What happened?**
management-api uses properties to handle how the JVM_OPTS is modified. Enabling CDC messes up this behavior and causes things that are not supposed to be there to re-appear or go missing. For example, MGMT_API_DISABLE_MCAC should not cause MCAC to appear.

The entire generate_config.go should probably be history at some point and the configuration of the startup properties should happen in the mgmt-api repository. 

**Environment**

* Cass Operator version:

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* Cass Operator Logs:

```
```

Miles-Garnsey:
**What this PR does**:

Ensures that CDC classpath/agent manipulation logic doesn't erroneously re-enable MCAC when the MCAC environment variable is set to disable it.

Fixes #481 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/cdc/generate_config.go
+++ b/pkg/cdc/generate_config.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // UpdateConfig updates the json formatted Cassandra config which incorporates the JVM options (under key additional-jvm-opts) passed into the launch scripts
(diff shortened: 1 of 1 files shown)


Miles-Garnsey:
This appears to be good to go now, there was one mistake and some test flakes happening earlier, but I think it is all good now.

emerkle826 (approved):

pkg/cdc/generate_config.go
```diff
@@ -90,17 +91,46 @@ func updateAdditionalJVMOpts(optsSlice []string, CDCConfig *cassdcapi.CDCConfigu
 			optsSlice = append(optsSlice, nameTag+"="+fmt.Sprintf("%s", reflectedValue))
 		}
 	}
+	// We need to add these two elements to the slice first, because the management agent must start before the CDC agent.
```
olim7t: Is this comment misplaced? It looks like a duplicate of the one on line 103.

pkg/cdc/generate_config.go
```diff
@@ -90,17 +91,46 @@ func updateAdditionalJVMOpts(optsSlice []string, CDCConfig *cassdcapi.CDCConfigu
...
+	// If MCAC is disabled, we need to avoid adding it to startup.
```
olim7t: Same here, this refers to MCAC but the next line deals with CDC. Shouldn't this be merged with line 97 rather?

pkg/cdc/generate_config.go
Miles-Garnsey: Probably, this is why some people say don't write comments - they fall out of date :/ 

I'm not sure what my thinking was there so I'm getting rid of these.

pkg/cdc/generate_config.go
Miles-Garnsey: Ditto here.

burmanm:
I'll fix the missing CHANGELOG line before release.

PR #503, opened 2023-03-08
Run /start on the background for pods and up the timeout 

### LINKED ISSUE #500: configurable timeout for lifecycle/start
**What is missing?**

Make the timeout for startCassandra() REST call to management api lifecycle configurable.

**Why do we need it?**

In reconciling racks startCassandra() REST call has a very short time out of only 10 seconds before it kills the pods. The called lifecycle/start call was always synchronous, it has to wait until the start script fully executed and Cassandra/DSE starts can take much more time than that. 

While it would be desirable to have this rest call async and then monitor state of starting-not-ready - currently that is not the case.

**Environment**

* Cass Operator version:

Needed in 1.10.x branches to be used in Astra Classic.

burmanm:
**What this PR does**:
Increase the /start timeout from 10s to 10 mins, but run it on the background to prevent blocking other operations in cass-operator.

Fixes #500

**Checklist**
- [ ] Changes manually tested


emerkle826 (approved):

pkg/reconciliation/reconcile_racks.go
```diff
@@ -1783,33 +1782,34 @@ func (rc *ReconciliationContext) startCassandra(endpointData httphelper.CassMeta
...
-			"Starting Cassandra for pod %s to replace Cassandra node with address %s", pod.Name, replaceAddress)
-		err = mgmtClient.CallLifecycleStartEndpointWithReplaceIp(pod, replaceAddress)
-	} else {
-		// Either we are not replacing this pod or the relevant cassandra node
-		// never joined the ring in the first place and can be started normally
-		rc.Recorder.Eventf(rc.Datacenter, corev1.EventTypeNormal, events.StartingCassandra,
-			"Starting Cassandra for pod %s", pod.Name)
-		err = mgmtClient.CallLifecycleStartEndpoint(pod)
-	}
+	go func(pod *corev1.Pod) {
```
berndocklin, line 1785: I know, late comment, but it would have been great to see a comment here why this needs to run concurrently

burmanm: It's a background task to allow cass-operator to process next command. The Reconcile pattern should never block itself.

PR #540, opened 2023-06-20
Enable new config builder for 4.1 deployments

### LINKED ISSUE #512: Replace cass-config-builder with Go
### What is missing?

We currently run cass-config-builder, which is not compatible with the new configs in 4.1 and up. The steps that we will need to implement are created as subtickets (we don't need to replicate entire cass-config-builder as a lot of it isn't Kubernetes related).

- [x] #513 
- [x] #514

In addition, some smaller ones:

- [ ] Sanitize env variables (including cassandra-env.sh additionals) to prevent any injections
- [ ] Get rid of some env variables that are not relevant to Kubernetes installations
- [ ] Add ability to set Cassandra version to 5.x.x also
- [ ] Support DSE 6.8.x and OSS versions

### Why is this needed?

Refactoring cass-config-builder is time consuming, is built on language we don't use (Clojure) and it's a resource hog.

### Additional info

We need to decide which versions will use the newer builder. Do we want to keep the old one for the stuff where it has worked previously (3.11.x, 4.0.x and DSE 6.8.x) and only use it for the newer stuff - or do we want to replace and use a common path for all versions?

Restricting it to newer ones would reduce the amount of boilerplate and different paths, but of course would require different deployment model for different versions.

### LINKED ISSUE #513: Add defaults if values are missing
### What is missing?

In the config-builder, the first step is to add defaults which might be missing. If we parse the cassandra.yaml directly (instead of creating it from the scratch like config-builder) this step can be merged to the enrich-step.

### Why is this needed?

To provide adequate information for Cassandra to work in Kubernetes.

### LINKED ISSUE #514: Enrich configs
### What is missing?

Current cass-config-builder will enrich the following keys of the config that's given from cass-operator:

* cassandra-yaml
* cassandra-env-sh
* dse
* datastax-env-sh
* cassandra-rack-properties

In the cassandra-yaml, we need to take care of several properties that the user should not override and we should always. These include `seed_provider`, `cluster_name` as well as:

```clojure
(def node-private-props #{:listen_address
                          :broadcast_address
                          :native_transport_address
                          :native_transport_broadcast_address
                          :rpc_address
                          :broadcast_rpc_address
                          :seed_provider
                          :initial_token
                          :auto_bootstrap})
```

And from the definitions, we'll need `listen_address` and such. Some of these are defined in the mcctl already, some have different settings depending on the version we'll support.

### Why is this needed?

Otherwise the configuration that's applied might not work correctly in our environment.

burmanm:
**What this PR does**:
Replaces the usage of cass-config-builder with the new k8ssandra-client config build when a Cassandra 4.1 and newer OSS image is deployed.

The old behavior remains with DSE and older versions of Cassandra.

Fixes #512
Fixes #513
Fixes #514

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Jeffail/gabs/v2"
 	"github.com/k8ssandra/cass-operator/pkg/serverconfig"
 	"github.com/pkg/errors"
+	"golang.org/x/mod/semver"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -952,4 +953,8 @@ func (dc *CassandraDatacenter) DatacenterName() string {
 		return dc.Spec.DatacenterName
 	}
 	return dc.Name
+}
+
+func (dc *CassandraDatacenter) UseClientImage() bool {
+	return dc.Spec.ServerType == "cassandra" && semver.Compare(fmt.Sprintf("v%s", dc.Spec.ServerVersion), "v4.1.0") >= 0
 }
--- a/config/manager/image_config.yaml
+++ b/config/manager/image_config.yaml
@@ -3,9 +3,9 @@ kind: ImageConfig
 metadata:
   name: image-config
 images:
-  system-logger: "k8ssandra/system-logger:latest"
+  system-logger: "k8ssandra/system-logger:v1.17.0-dev.ada6cb3-20230727"
   config-builder: "datastax/cass-config-builder:1.0-ubi7"
-  k8ssandra-client: "k8ssandra/k8ssandra-client:bee01ccf"
+  k8ssandra-client: "michaelburman290/k8ssandra-client:v0.2.0-dev.bee01cc-20230727"
   # cassandra:
   #   "4.0.0": "k8ssandra/cassandra-ubi:latest"
   # dse:
--- a/config/manager/image_config.yaml
+++ b/config/manager/image_config.yaml
@@ -3,9 +3,9 @@ kind: ImageConfig
 metadata:
   name: image-config
 images:
-  system-logger: "k8ssandra/system-logger:v1.17.0-dev.ada6cb3-20230727"
+  system-logger: "k8ssandra/system-logger:v1.17.0"
   config-builder: "datastax/cass-config-builder:1.0-ubi7"
-  k8ssandra-client: "k8ssandra/k8ssandra-client:73bae732"
+  k8ssandra-client: "k8ssandra/k8ssandra-client:e5ae32c3"
   # cassandra:
   #   "4.0.0": "k8ssandra/cassandra-ubi:latest"
   # dse:
--- a/config/manager/image_config.yaml
+++ b/config/manager/image_config.yaml
@@ -5,7 +5,7 @@ metadata:
 images:
   system-logger: "k8ssandra/system-logger:v1.17.0-dev.ada6cb3-20230727"
   config-builder: "datastax/cass-config-builder:1.0-ubi7"
-  k8ssandra-client: "michaelburman290/k8ssandra-client:v0.2.0-dev.bee01cc-20230727"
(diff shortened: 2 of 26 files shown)


burmanm:
I just realized this isn't going to test the new stuff in the CI since it overrides the version in the GHA workflow. Dang.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -953,3 +954,7 @@ func (dc *CassandraDatacenter) DatacenterName() string {
 	}
 	return dc.Name
 }
+
+func (dc *CassandraDatacenter) UseClientImage() bool {
```
adejanovski, line 958: **suggestion:** Some comments or a different method name could help understanding the intent here. Currently I got confused by what this is supposed to support as feature (no mention of config building for example).

config/manager/image_config.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
 images:
   system-logger: "k8ssandra/system-logger:latest"
   config-builder: "datastax/cass-config-builder:1.0-ubi7"
+  k8ssandra-client: "michaelburman290/k8ssandra-client:v0.2.0-dev.cfda014-20230724"
```
adejanovski: **todo:** It would be better to get an actual release of the k8ssandra-client image with the new config builder before we merge this.

adejanovski (changes requested):
Seems like there are some errors in the new checks.
I just have one blocking request around the image coordinates, and the rest looks good.
Very cool to finally be replacing cass-config-builder!

config/manager/image_config.yaml
burmanm: Yep, but it wasn't available until now.

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: True, I just lacked imagination when creating the method. Perhaps something like "NewConfigSystem() bool" ?

burmanm: I mean the real idea is to check if the deployed Cassandra version uses old Config.java or the newer one.

adejanovski: `NewConfigSystem()` sounds better. Maybe `Builder` would be better than `System`?

The idea indeed is that with versions equal or higher than 4.1, we'll use the new config builder that sits in the k8ssandra-client project.
`New` is a little vague as well, because if we come up with another one later (god forbids) we'll end up with `NewNew` ? :D 
Is `UseK8ssandraConfigBuilder()` too long? You could also turn it into `K8ssandraConfigBuilder() bool`, which may work with Go idioms I guess?

burmanm: Ah, I didn't mean to reflect on the newer config-builder, I was thinking about from the perspective of Cassandra itself. That said, of course Cassandra's Config system can evolve as well later.

adejanovski (approved):
Approved, pending passing tests

burmanm:
There's something super fishy with the size = 2 and role creation in 4.1 settings. 

```
INFO  [epollEventLoopGroup-5-3] 2023-08-15 08:04:37,693 RpcExecutionException.java:34 - Failed to execute method NodeOps.createRole
org.apache.cassandra.exceptions.UnavailableException: Cannot achieve consistency level EACH_QUORUM in DC My_Super_Dc
```

emerkle826:
> There's something super fishy with the size = 2 and role creation in 4.1 settings.
> 
> ```
> INFO  [epollEventLoopGroup-5-3] 2023-08-15 08:04:37,693 RpcExecutionException.java:34 - Failed to execute method NodeOps.createRole
> org.apache.cassandra.exceptions.UnavailableException: Cannot achieve consistency level EACH_QUORUM in DC My_Super_Dc
> ```

Is this an issue where creating Roles use a special CL that is different than the effective CL for other queries? Or is it just a weird thing specifically with exactly 2 nodes (i,.e things work with 1 node, or 3+ nodes, but not 2)?

burmanm:
> Is this an issue where creating Roles use a special CL that is different than the effective CL for other queries? Or is it just a weird thing specifically with exactly 2 nodes (i,.e things work with 1 node, or 3+ nodes, but not 2)?

The number 2.. but hopefully that's solved now.

PR #555, opened 2023-07-18
Update server-logger to function with multiple securityContext settings

### LINKED ISSUE #553: Deprecate dockerImageRunsAsCassandra
### What is missing?

dockerImageRunsAsCassandra has been set to true as default for a long time. Setting it to false provided the ability to run the images as root, but none of our images do that.

The correct way to do this is to override the functionality with setting podTemplateSpec/spec/securityContext.

### Why is this needed?

Some users have set this to false due to some old examples, but that does not do what the user expects (nor does it fix anything).

Correct method that would work is to use securityContext which is similar to what we're using with OpenShift deployments.

### LINKED ISSUE #554: server-system-logger does not work with different securityContexts
### What happened?

When using SecurityContext other than 999:999, the server-system-logger will fail with the /var/lib/vector rights.

This requires us to use nonroot in OpenShift as well as restrict certain installation options.

_No response_

-

### cass-operator version

1.16.0

### Kubernetes version

1.27

### Method of installation

_No response_

_No response_

burmanm:
**What this PR does**:
Modifies the server-logger to use an EmptyVolume to ensure that it doesn't have securityContext restrictions.

* Update the CI tests to using a :version instead of :latest so that we don't accidentially test with incorrect version
* Deprecate runDockerAsCassandraUser since our current security changes have made it non-functional.

Fixes #553 
Fixes #554 

**Checklist**
- [x] Changes manually tested


pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -304,7 +304,14 @@ func addVolumes(dc *api.CassandraDatacenter, baseTemplate *corev1.PodTemplateSpe
...
 
-	volumeDefaults := []corev1.Volume{vServerConfig, vServerLogs}
+	vectorLib := corev1.Volume{
+		Name: "vector-lib",
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	}
+
+	volumeDefaults := []corev1.Volume{vServerConfig, vServerLogs, vectorLib}
```
emerkle826, line 314: So is this always adding a Vector volume mount now? Even if cass-operator is deployed without Vector?

burmanm: Yes, but since it's not mounted anywhere it's not really created.

burmanm:
That CI error is not relevant, it's really just CI issue with download issue (originally, later the date won't match so it won't work).

emerkle826 (approved):

PR #556, opened 2023-07-26
Revert commit that rolled back logger.Dockerfile, never push :latest

burmanm:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/logger.Dockerfile
+++ b/logger.Dockerfile
@@ -3,14 +3,12 @@ ARG VERSION
 ARG TARGETPLATFORM
 
 # Install Vector
-ENV VECTOR_VERSION 0.29.1
+ENV VECTOR_VERSION 0.31.0
 RUN case ${TARGETPLATFORM} in \
          "linux/amd64")  VECTOR_ARCH=x86_64  ;; \
          "linux/arm64")  VECTOR_ARCH=aarch64  ;; \
     esac \
  && rpm -i https://packages.timber.io/vector/${VECTOR_VERSION}/vector-${VECTOR_VERSION}-1.${VECTOR_ARCH}.rpm
-
-RUN mkdir -p /var/lib/vector
 
 FROM redhat/ubi8-micro:latest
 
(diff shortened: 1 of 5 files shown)


burmanm:
For context, this is where the version-tag is used for system-logger and cass-operator to ensure our tests don't just use :latest and be happy:

https://github.com/k8ssandra/cass-operator/blob/master/Makefile#L208

burmanm:
Ah.. need to fix the Kustomize version again:

/home/runner/work/cass-operator/cass-operator/bin/kustomize version is not expected v5.0.3. Removing it before installing.
test -s /home/runner/work/cass-operator/cass-operator/bin/kustomize || { curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash -s -- 5.0.3 /home/runner/work/cass-operator/cass-operator/bin; }
Github rate-limiter failed the request. Either authenticate or wait a couple of minutes.

logger.Dockerfile
```diff
@@ -29,11 +27,10 @@ LABEL description="Sidecar to output Cassandra system logs to stdout"
 COPY ./config/logger/vector_config.toml /etc/vector/vector.toml
 COPY --from=builder /usr/lib64/libstdc++.so.6 /usr/lib64/libstdc++.so.6
 COPY --from=builder /usr/bin/vector /usr/bin/vector
-COPY --from=builder --chown=999:999 /var/lib/vector /var/lib/vector
```
emerkle826, line 32: Question #1: Does removing the ownership change here cause issues in OpenShift when running as non-root user?

If you have tested this change in OpenShift and there is no issue when running with restricted security context, the rest of this is moot.

If we do use `chown` in the COPY here, I have found that for permissions to work correctly in OpenShift, you need to create the user with UID:GID in the `builder` stage, and then either do it again in this final stage, or copy the /etc/passwd, /etc/group and /etc/shadow files from the `builder` stage after creating the user in that stage.

 For some strange reason, it seems if you create files in a previous stage where the `cassandra` user does not exist, and then copy them in the final stage, even if you use a RUN line to change ownership and permissions on the files you copied, it still somehow does not have the correct permissions for running in OpenShift where the effective UID and GID are some random ids that are not `999`. Only when I created the `cassandra` user in the previous stage and then created that user again in the final stage did the permissions get set the way I expected/intended.

Question #2: All of that said, I'm guessing removing this line is fine because `var/lib/vector` will only be added to the image as a mount and won't have the OpenShift permissions issues we had with some of the Management API images?

emerkle826 (approved):
This looks fine, just a couple of questions regarding the image and use in OpenShift

logger.Dockerfile
burmanm: Yes, it's removed here because Kubernetes creates it. Adding chown here would make no difference.

PR #562, opened 2023-08-23
Add new Prometheus metrics, cass_operator_datacenter_pods_status

### LINKED ISSUE #184: K8SSAND-937 ⁃ Expose CassandraDatacenter status as metrics
What is missing?

Expose various status indicators from the CassandraDatacenter resources as metrics that could be scraped and aggregated by 3rd party tools.

Why do we need it?

So that 3rd party observability and alerting tools can monitor, visualize, and more easily create custom thresholds to establish the status of the cluster.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-937) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-937
┆priority: Medium

burmanm:
**What this PR does**:
Adds a new Prometheus metric cass_operator_datacenter_pods_status which has status label as the current status of the pod, with value of 1 indicating the current active status.

While this fixes #184, we might want to add more metrics indicating the reconciliation status also in the future.

Fixes #184

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/monitoring/metrics.go
+++ b/pkg/monitoring/metrics.go
@@ -1,0 +1,89 @@
+package monitoring
+
+import (
+	"strings"
+
+	api "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+type PodStatus string
+
+const (
+	PodStatusInitializing    PodStatus = "Initializing"
+	PodStatusReady           PodStatus = "Ready"
+	PodStatusPending         PodStatus = "Pending"
+	PodStatusError           PodStatus = "Error"
+	PodStatusDecommissioning PodStatus = "Decommissioning"
+)
+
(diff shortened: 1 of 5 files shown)


pkg/monitoring/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	return status
+}
+
+var (
+	PodStatusVec *prometheus.GaugeVec
+)
+
+func init() {
+	podVec := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "cass_operator",
```
adejanovski, line 61: **Issue**: cass-operator is hardcoded here but we will have DCs in different namespaces.
Even if this `Namespace` notion is not equivalent to the k8s namespace, we lack a label in the metrics to filter namespaces. We could two dcs with the same cluster and dc names but in different namespaces and I don't see how we could differentiate them.

adejanovski (changes requested):
Looks great!
I just have an issue with the lack of namespace label.

pkg/monitoring/metrics.go
burmanm: This "namespace" is actually just part of the metric name, the structure is ``Namespace_Subsystem_Name``. We can add namespace as a label or we can of course keep changing the metric name also, but I would assume label is easier to query and index in any Prometheus compatible system instead of changing the prefix of a metric name.

adejanovski (approved):
Looks good :shipit:

PR #564, opened 2023-08-25
Fix e2e tests, change from wget to curl

### LINKED ISSUE #565: Replace wget with curl
### What happened?

At some point, the DSE images built with -ubi8 / -ubi7 prefix lost their wget from default installation. All our drain endpoints and other mgmt-api calls (which happen directly from Kubernetes) depend on wget to exists. Suddenly, this no longer was true.

So to keep older images working correctly, we need to use curl instead. 

_No response_

Using any image without wget. It'll fail on smoke tests.

### cass-operator version

1.16.0

### Kubernetes version

1.27

### Method of installation

_No response_

_No response_

burmanm:
**What this PR does**:
This PR changes the default Kubernetes -> management-api communication from using wget to using curl. This allows better compatibility between different images.

* Fixes scale_down_unbalanced_test by setting all correct parameters to it (M_SERVER_TYPE & M_SERVER_VERSION). 
* Fixes DSE tests to use ``-Ddse`` prefix if ``-Dcassandra`` is used instead for replication factors.

Fixes #565 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/.github/workflows/kindIntegTest.yml
+++ b/.github/workflows/kindIntegTest.yml
@@ -59,8 +59,8 @@ jobs:
     strategy:
       matrix:
         integration_test:
-          - upgrade_operator # Test is not setup to run against 4.0
-          - additional_seeds #TODO: Fails against C* 4.0, fix in https://github.com/k8ssandra/cass-operator/issues/459
+          # - upgrade_operator # Test is not setup to run against 4.0
+          # - additional_seeds #TODO: Fails against C* 4.0, fix in https://github.com/k8ssandra/cass-operator/issues/459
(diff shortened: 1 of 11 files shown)


burmanm:
decommission dc is fine (well, as fine as it's always been, it's a bit flaky), it's just that our e2e tests can't be rerun on the next day at this point:

Run kind load docker-image --name=kind k8ssandra/cass-operator:v1.17.0-dev.2e60096-20230830
ERROR: image: "k8ssandra/cass-operator:v1.17.0-dev.2e60096-20230830" not present locally

I probably need to ditch the -date at the end of the version number for GHA runs.

adejanovski (approved):

PR #572, opened 2023-09-26
Ensure server-config-init and server-config-init-base are in initContainers

### LINKED ISSUE #571: More resilience to initContainers overrides
### What is missing?

Currently, overriding the "server-config-init" caused the "server-base-config" init container to be absent as well, making users unaware that the upgrade from 4.0 -> 4.1 would cause issues in starting the container. This caused also the k8ssandra-operator to fail when using 4.1.x and medusa.

We should allow automated correction to always include both server-config-init and server-base-config, but also allow user to add extra initContainers before and after these configs. "server-config-init" should be appended as last one, "server-base-config" before it.

### Why is this needed?

Users might not be aware that all the required initContainers must be defined if one is overridden.

burmanm:
**What this PR does**:
Changes the override function of initContainers. Previously, setting "server-config-init" would allow overriding all the initContainers, but that also meant removing "server-config-init-base" even if that was actually necessary and user did not set it.

Fixes #571

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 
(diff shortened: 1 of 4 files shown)


adejanovski (approved):
Looks good and works as expected 👍

PR #574, opened 2023-10-02
Add namespace label on metrics

### LINKED ISSUE #573: Metrics need the namespace as label
We currently don't label the metrics scraped through Vector with the namespace in which the cluster is deployed.
This doesn't allow distinguishing two clusters in different namespaces that would have the same cluster and dc names from each other.
We need to add the namespace as env variable in the server system logger container, and use it to label the metrics (like we do for cluster, dc and rack names).

adejanovski:
**What this PR does**:
Adds the namespace as env variable in the server-system-logger container.

Fixes #573 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pkg/reconciliation/construct_podtemplatespec.go
+++ b/pkg/reconciliation/construct_podtemplatespec.go
@@ -710,6 +710,7 @@ func buildContainers(dc *api.CassandraDatacenter, baseTemplate *corev1.PodTempla
 		{Name: "CLUSTER_NAME", Value: dc.Spec.ClusterName},
 		{Name: "DATACENTER_NAME", Value: dc.DatacenterName()},
 		{Name: "RACK_NAME", ValueFrom: selectorFromFieldPath("metadata.labels['cassandra.datastax.com/rack']")},
+		{Name: "NAMESPACE", Value: dc.Namespace},
 	}
 
 	loggerContainer.Env = combineEnvSlices(loggerEnvDefaults, loggerContainer.Env)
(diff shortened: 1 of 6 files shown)


pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -710,6 +710,7 @@ func buildContainers(dc *api.CassandraDatacenter, baseTemplate *corev1.PodTempla
 		{Name: "CLUSTER_NAME", Value: dc.Spec.ClusterName},
 		{Name: "DATACENTER_NAME", Value: dc.DatacenterName()},
 		{Name: "RACK_NAME", ValueFrom: selectorFromFieldPath("metadata.labels['cassandra.datastax.com/rack']")},
+		{Name: "NAMESPACE", Value: dc.Namespace},
```
burmanm: I don't think this is a correct approach in general sense or how other stuff is done. The pod itself knows the namespace already, it's in the file ``/var/run/secrets/kubernetes.io/serviceaccount/namespace`` so just catting that would be enough. This is also something that can be used in the upstream alone without needing a new version of cass-operator by simply modifying the vector configuration there.

The other "correct" method I think is to use selector, so:

```go
{Name: "NAMESPACE", ValueFrom: selectorFromFieldPath("metadata.namespace")},
```

adejanovski: Vector VRLs language doesn't have a function to read data from a file. All we can do is read the hostname or read an env variable (see [here](https://vector.dev/docs/reference/vrl/functions/#system-functions)).
So I think we need the additional variable to avoid building a convoluted solution.
I'm ok with using the "metadata.namespace" instead of the dc namespace, I'll push a change for this.

burmanm (approved):

PR #579, opened 2023-10-09
Fix serverImage example in README

apa64:
**What this PR does:**
Fixes serverImage example in README to have `@sha256` when referencing an image ID.

**Which issue(s) this PR fixes:**
N/A

**Checklist**

- [x] Changes manually tested
- [ ] CLA Signed: DataStax CLA

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -429,7 +429,7 @@ metadata:
 spec:
   serverType: cassandra
   serverVersion: 3.11.7
-  serverImage: private-docker-registry.example.com/cass-img/cassandra-with-mgmtapi:1a2b3c4d
+  serverImage: private-docker-registry.example.com/cass-img/cassandra-with-mgmtapi@sha256:1a2b3c4d
 ```
 
 DSE:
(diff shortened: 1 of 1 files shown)


burmanm:
We don't require SHA256 to be used in the ServerImage field, that would be up to the user. Only our OLM installation uses that method.

apa64:
Oh! I misunderstood how it works - thank you! I'll cancel this PR as invalid.

apa64:
Closed as invalid, do not merge.

burmanm:
To clarify, there's nothing wrong if you wish to use SHA256s, that's fine. But some users prefer to use the tagged versions, given that they could get updated base-images (such as CVE fixes like the upcoming curl one) that way automatically, while using SHA256 images would prevent that and require manually updating the manifests.

Both methods have their own pros/cons. But for examples, I'd prefer to keep it as simple tagged ones to ensure users always have the newest base image.

apa64:
We encountered an issue with our private registry (working as pull-through cache): the cluster account is not allowed to pull external images to the cache from docker.io automatically. The updated minor version tag made us hit this restriction. I see now that in our case I should specify the ID. Thank you, again :)

PR #584, opened 2023-10-11
Implement new tasks, fix task bugs and improve testing

### LINKED ISSUE #578: Add a flush CassandraTask
We would need to give the ability to perform flushes using a CassandraTask.
It would perform the equivalent of a "nodetool flush" command, but through the management api.
This task would need to be configurable so it can run either sequentially or in parallel.

```[tasklist]
### Definition of Done
- [ ] A CassandraTask command to support a flush operation is added
```

### LINKED ISSUE #580: Add a GC CassandraTask which can target specific nodes or a full DC
We'd need to have support for [garbagecollect](https://cassandra.apache.org/doc/latest/cassandra/tools/nodetool/garbagecollect.html) as CassandraTask job type.

This Cassandra task should support the granularity and jobs arguments.

```[tasklist]
### Definition of Done
- [ ] A new CassandraTask job type is added to support garbagecollect, with support for the granularity and jobs arguments
```

### LINKED ISSUE #582: Add a compaction CassandraTask
### What is missing?

We would need to give the ability to perform compactions using a CassandraTask.
It would perform the equivalent of a "nodetool compact" command, but through the management api.
This task would need to be configurable so it can run either sequentially or in parallel, and target specific nodes/pods only.

```[tasklist]
### Definition of Done
- [ ] A CassandraTask command to support a compaction operation is added
- [ ] That command can take a keyspace and a table as arguments
- [ ] This CassandraTask can optionnally target specific nodes
```

### Why is this needed?

Compaction is a common operation when managing Cassandra clusters, and we want to have a more Kubernetes native way of scheduling them, without requiring to run `nodetool` from within the containers.

### LINKED ISSUE #585: Task validation will never be visible in the task.Status
### What happened?

If task validation, we would only log it to the cass-operator logs and then requeue. This would happen forever and the only way user would know why the task is forever in the running state would be to check the cass-operator logs.

The task would have Failed condition set and it would have a message indicating what went wrong.

Create a replace job without target podName.

### cass-operator version

1.17.2

### Kubernetes version

1.27

### Method of installation

_No response_

_No response_

### LINKED ISSUE #586: Add CassandraTask scrub
### What is missing?

We don't have the ability to run scrub on a dataset as CassandraTask. 

### Why is this needed?

Sometimes data might need rebuilding, or recovering from errors.

burmanm:
**What this PR does**:
* Implements new task types, scrub, compaction, flush and garbagecollect
* Improves testing of tasks with ability to verify workload sent to the management-api
* Fix some bugs related to sync tasks
* Modify TaskConditions to be metav1.Conditions instead of JobConditions. This is backwards compatible in YAML format.

Fixes #578 
Fixes #580
Fixes #586 
Fixes #585 
Fixes #582 

**Checklist**
- [ ] Changes manually tested


burmanm:
Can't find the Draft button.. ouh well. This is still missing scrub & compact tasks + CHANGELOG updates.

adejanovski:
Draft button is at the top right, under reviewers ;)

I did turn it into a draft.

burmanm:
The test failures are expected until 0.6.70 from mgmt-api is released.

emerkle826 (approved):

PR #601, opened 2023-12-19
Add AdditionalAnnotations field and logic to reconcile it

### LINKED ISSUE #600: Add AdditionalAnnotations field
### What is missing?

We currently have an AdditionalLabels field in the CassDC CRD which propagates labels to the resources created by cass-operator. Per [this ticket](https://github.com/k8ssandra/k8ssandra-operator/issues/1094), there is a need in the community to be able to propagate annotations in addition to labels. 

### Why is this needed?

As above.

Miles-Garnsey:
**What this PR does**:

Fixes #600 

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/oplabels/labels.go
+++ b/pkg/oplabels/labels.go
@@ -20,7 +20,7 @@ const (
 	CreatedByLabelValue = ManagedByLabelValue
 )
 
-func AddOperatorLabels(m map[string]string, dc *api.CassandraDatacenter) {
+func AddOperatorTags(m map[string]string, dc *api.CassandraDatacenter) {
 	m[ManagedByLabel] = ManagedByLabelValue
 	m[NameLabel] = NameLabelValue
 	m[VersionLabel] = dc.Spec.ServerVersion
@@ -29,6 +29,12 @@ func AddOperatorLabels(m map[string]string, dc *api.CassandraDatacenter) {
 
 	if len(dc.Spec.AdditionalLabels) != 0 {
 		for key, value := range dc.Spec.AdditionalLabels {
+			m[key] = api.CleanLabelValue(value)
+		}
+	}
+
+	if len(dc.Spec.AdditionalAnnotations) != 0 {
+		for key, value := range dc.Spec.AdditionalAnnotations {
 			m[key] = api.CleanLabelValue(value)
 		}
 	}
--- a/pkg/oplabels/labels.go
+++ b/pkg/oplabels/labels.go
@@ -41,7 +41,7 @@ func AddOperatorAnnotations(m map[string]string, dc *api.CassandraDatacenter) {
 	}
 	if len(dc.Spec.AdditionalAnnotations) != 0 {
 		for key, value := range dc.Spec.AdditionalAnnotations {
-			m[key] = api.CleanLabelValue(value)
+			m[key] = value
 		}
 	}
 }
--- a/pkg/oplabels/labels.go
+++ b/pkg/oplabels/labels.go
@@ -20,7 +20,7 @@ const (
 	CreatedByLabelValue = ManagedByLabelValue
 )
 
-func AddOperatorTags(m map[string]string, dc *api.CassandraDatacenter) {
+func AddOperatorLabels(m map[string]string, dc *api.CassandraDatacenter) {
 	m[ManagedByLabel] = ManagedByLabelValue
 	m[NameLabel] = NameLabelValue
 	m[VersionLabel] = dc.Spec.ServerVersion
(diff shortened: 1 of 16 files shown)


Miles-Garnsey:
Manually testing:

```
$ kubectl get services -n cass-operator -o json | jq '.items[] | .metadata.name, .metadata.annotations."test-annotation"'
"cass-operator-webhook-service"
null
"cluster1-dc1-additional-seed-service"
"is-here"
"cluster1-dc1-all-pods-service"
"is-here"
"cluster1-dc1-service"
"is-here"
"cluster1-seed-service"
"is-here

$ kubectl get secrets -n cass-operator -o json | jq '.items[] | .metadata.name, .metadata.annotations."test-annotation"'
"cass-operator-controller-manager-token-t478h"
null
"cluster1-superuser"
"is-here"
"default-token-h6pfj"
null
"webhook-server-cert"
null

$kubectl get pods -n cass-operator -o json | jq '.items[] | .metadata.name, .metadata.annotations."test-annotation"'
"cass-operator-controller-manager-dc8d5f5df-jscb8"
null
"cluster1-dc1-default-sts-0"
"is-here"

kubectl get sts -n cass-operator -o json | jq '.items[] | .metadata.name, .metadata.annotations."test-annotation"'
"cluster1-dc1-default-sts"
"is-here"
```

Looks like all secrets, services, pods and statefulsets receieve the desired annotations.

Miles-Garnsey:
@burmanm can I get an approval on this when you get a sec?

pkg/oplabels/labels.go
```diff
@@ -32,6 +32,18 @@ func AddOperatorLabels(m map[string]string, dc *api.CassandraDatacenter) {
...
+
+}
+
+func AddOperatorAnnotations(m map[string]string, dc *api.CassandraDatacenter) {
+	if m == nil {
+		m = make(map[string]string)
+	}
+	if len(dc.Spec.AdditionalAnnotations) != 0 {
+		for key, value := range dc.Spec.AdditionalAnnotations {
+			m[key] = api.CleanLabelValue(value)
```
burmanm: issue: This isn't correct. Label values have different requirements in Kubernetes than Annotation values (label and annotations have similar key requirements, but not values).

https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

pkg/reconciliation/reconcile_racks.go
```diff
@@ -1616,12 +1650,32 @@ func (rc *ReconciliationContext) ReconcilePods(statefulSet *appsv1.StatefulSet)
...
+			rc.Recorder.Eventf(rc.Datacenter, corev1.EventTypeNormal, events.LabeledRackResource,
+				"Update rack annotations for pvc %s", pvc.Name)
+		}
 	}
 
 	return nil
 }
 
-func mergeInLabelsIfDifferent(existingLabels, newLabels map[string]string) (bool, map[string]string) {
+func mergeInTagsIfDifferent(existingLabels, newLabels map[string]string) (bool, map[string]string) {
```
burmanm: issue: What are tags? Why is this renamed to something weird?

pkg/oplabels/labels.go
Miles-Garnsey: That's interesting, I didn't realise, I'll correct this.

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: The 'tags' word comes from k8ssandra-operator, I wanted to re-use the same function for both annotations and labels, which means we need some other word that covers both.

So 
1. Is this a nit, issue, suggestion, or..?
2. Feel free to suggest a better word if you can think of one, and if you think it is worth moving away from the k8ssandra-operator way of naming things.

pkg/oplabels/labels.go
Miles-Garnsey: It looks like (although it isn't mentioned directly in those links) annotation values are basically a free-for-all. I guess that makes sense as they aren't indexed the same way. I'll remove all validation since it doesn't seem like it will be required.

pkg/reconciliation/reconcile_racks.go
burmanm: This isn't used for annotations, not anywhere in your PR (the only change is modifying name of this function). This function has had its name for 4 years now, k8ssandra-operator hasn't even existed that long.

This change shouldn't be in this PR. Is it a leftover of something that should be here, or something that's just dead code change that wasn't necessary?

Miles-Garnsey: Hmmm, I see your point having taken a closer look. Looks like (as you say) this is from a previous approach I tried rather than being strictly required. I will change it back.

burmanm (approved):

PR #631, opened 2024-04-08
Operator upgrade should not cause rolling restart

### LINKED ISSUE #566: Operator upgrade should not cause a rolling restart
### What is missing?

When updating the operator, there are two cases where we will actually cause a rolling restart of all Cassandra clusters managed by the same operator.

* imageConfig updates which will ship with updates to k8ssandra-client and server-system-logger
* Updates to PodTemplateSpec caused by something new we introduced / wanted to change

Now, while both are good updates usually, they shouldn't happen if user makes no changes. We would hope that user upgrades to newest builds of server-system-logger for example to get bugfixes / CVE free base builds. Or if we upgrade Cassandra to use better settings we usually do that for stability purposes. However, neither should happen without the knowledge of the user and shouldn't be an uncontrolled side-effect of updating the operator.

Instead, we should note in the release notes if we think user should do an rolling restart to get newer images or other improvements. 

I would propose we modify the processing slightly in cass-operator. If the Generation of CassandraDatacenter is equal to the ObservedGeneration, we would stop doing updates to any spec we deploy, but keep on doing the other parts of the processing (such as restarting nodes). Thus, this event processing can't happen in the Reconcile function's beginning, but instead at the points where something would be updated. All decommissions, status updates to CassandraDatacenter/status and other maintenance should continue to work as previously.

A rolling restart should cause the upgrade to the StS specs and others we detect. While this is a side-effect, there's an additional benefit to this and that's self healing. At the moment, modifying StS is a possibility which we will not revert back, but that could let users make the cluster unstable and operator would do nothing to fix that situation. Now, current rolling restart is done by modifying the PodTemplateSpec StS annotation that says restart time (like kubectl's rollout restart does), so we need to catch even this small change. What we have to do is prevent rolling restart twice, first for annotation change and then by cass-operator change. Not sure how to do that in every case, but that should be the goal.

We will need a test for all of this using the "upgrade_operator" test (and make it work with newer Cassandra version also) to detect cases where operator upgrade would do a rolling restart and fail the build in those cases.

Possible issues:

* An update to operator really requires an update to server-system-logger for example, such as that it can't function without those modifications. However, this same problem could arise with versions of management-api as well and we have a detection for that. So, similar approach should work and we could just make events + logging to indicate to users that they really should do a restart at some point.
* How many upgrades can we accept? What if users keep updating the operator, but the cluster hasn't been restarted in years? Rare issue probably, but not impossible. Can we still maintain a cluster that is based on some configuration a very old cass-operator version did? Should we add an annotation that indicates which version of cass-operator created the build so we could detect such cases? It could fit into the existing "created-by / managed-by" annotations without making a huge difference. We'll treat no-annotation as 1.16.0 for example.

### Why is this needed?

Updating the operator should be safe for the cluster - always. We should not make people run old versions of cass-operator because they're afraid of updating or do not have a possibility of restarting all clusters. Especially multiple clusters managed by the same cass-operator could be problematic as that would cause heavy load on the underlying Kubernetes cluster.

```[tasklist]
### Tasks
- [ ] cass-operator: If the Generation is the same as the ObservedGeneration we don't update the StatefulSet objects in Kubernetes (instead, we mark the CassandraDatacenter as requiring refresh). But we only do this if and only if we would update the StatefulSets in a way that would cause rolling restart. Metadata updates would still be applied.
- [ ] upstream operators: Same behavior where we prevent the updates of the cluster/dc custom resources if the generation wasn't updated
- [ ] A condition should be placed on the object to indicate that a restart is pending (something as "RestartRequired" condition)
- [ ] We would introduce a new annotation on the CassandraDatacenter objects allowing self healing of the StatefulSets, thus disabling the new behavior of not automatically fixing them
- [ ] In order to apply the upgrade changes, we can rely on a new annotation `cassandra.datastax.com/allow-sts-upgrade=true`. This annotation would be placed by either the users or the rolling restart K8ssandra/CassandraTask. If this annotation is detected, the Spec update will be unblocked, the "RestartRequired" condition will be set to false and the annotation will be removed at the end of the reconcile.
- [ ] If update is required and restart task is requested (for entire Datacenter), the restart of StS is not done in the task, but instead we only modify the CassandraDatacenter object. At that point, we should also track the restart from it and not from the status of StatefulSets. Alternatively (perhaps better), we add new task called "update/refresh" which would do this. Or we don't do this task at all and instead just let user modify the CassandraDatacenter object..
- [ ] We need to allow reconcile to trigger if this annotation is placed on the objects (cassdc, k8c). This requires an additional predicate for annotation listeners, but then one more filter. This could be useful outside of this feature also.
```

burmanm:
**What this PR does**:
Missing two features: new Condition and the annotation name might require changing. And also some unit tests and not just the e2e test. The RestartTask changes should come in separate PR (and will get its own ticket).

* Adds new reconcile event: annotation changes (and not just generation changes)
* Do not upgrade StatefulSets if Generation hasn't changed
* Modify upgrade_operator to run against 4.1 and verify in the operator upgrade test that the pods are not restarted on operator upgrade without added annotation approval. After the approval, ensure the pods were actually modified and have been restarted. Also, verify the "once" annotation was removed at the end.

Fixes #566 

**Checklist**
- [ ] Changes manually tested


pkg/reconciliation/constructor.go
```diff
@@ -68,5 +68,16 @@ func setOperatorProgressStatus(rc *ReconciliationContext, newState api.ProgressS
 		return err
 	}
 
+	// The allow-upgrade=once annotation is temporary and should be removed after first successful reconcile
+	if metav1.HasAnnotation(rc.Datacenter.ObjectMeta, "cassandra.datastax.com/allow-upgrade") && rc.Datacenter.Annotations["cassandra.datastax.com/allow-upgrade"] == "once" {
```
burmanm: Note to self: the annotation itself should be in a const.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -166,6 +166,10 @@ func (rc *ReconciliationContext) CheckRackCreation() result.ReconcileResult {
 	return result.Continue()
 }
 
+func (rc *ReconciliationContext) AllowdUpdate() bool {
```
burmanm: And I know there's a typo.

tests/upgrade_operator/upgrade_operator_suite_test.go
```diff
@@ -82,55 +82,78 @@ var _ = Describe(testName, func() {
...
 			// Verify PodDisruptionBudget is available (1.11 updates from v1beta1 -> v1)
-			json = "jsonpath={.items[].metadata.name}"
-			k = kubectl.Get("poddisruptionbudgets").WithLabel("cassandra.datastax.com/datacenter").FormatOutput(json)
-			err = ns.WaitForOutputContains(k, "dc1-pdb", 20)
-			Expect(err).ToNot(HaveOccurred())
-
-			// Update Cassandra version to ensure we can still do changes
-			step = "perform cassandra upgrade"
-			json = "{\"spec\": {\"serverVersion\": \"3.11.14\"}}"
+			// json = "jsonpath={.items[].metadata.name}"
```
burmanm, line 115: Remove this commented out check, it was checked in older versions and that's fine. It hasn't been relevant for a long time.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 
+* [CHANGE] [#566](https://github.com/k8ssandra/cass-operator/issues/566) BREAKING: StatefulSets will no longer be updated if CassandraDatacenter is not modified, unless an annotation "cassandra.datastax.com/allow-upgrade" is set to the CassandraDatacenter with value "always" or "once". This means users of config secret should set this variable to "always" to keep their existing behavior. For other users, this means that the upgrades of operator will no longer automatically apply updated settings or system-logger image. The benefit is that updating the operator no longer causes the cluster to have a rolling restart. A new condition to indicate such change could be necessary is called "RequiresUpdate" and it will be set to True until the next refresh of reconcile has happened.
```
olim7t: > users of config secret should set this variable to "always" to keep their existing behavior

[thought] This feels a bit inconsistent. Whether the config is directly in the CassandraDatacenter or externalized in a secret is just a detail, overall it's the same semantics (as far as I know).

Is there any way that we could detect when a reconcile comes from a config secret change, and treat that the same as a regular CassDC modification?

~~EDIT -- just saw your comment in `ReconciliationContext.UpdateAllowed()` which I think goes in that direction so +1~~ nm I misread that comment

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -65,6 +65,9 @@ const (
 	// cluster has gone through scale up operation.
 	NoAutomatedCleanupAnnotation = "cassandra.datastax.com/no-cleanup"
 
+	// UpdateAllowedAnnotation marks the Datacenter to allow upgrades to StatefulSets even if CassandraDatacenter object was not modified. Allowed values are "once" and "always"
+	UpdateAllowedAnnotation = "cassandra.datastax.com/allow-upgrade"
+
```
olim7t, line 73: [nitpick] Maybe extract constants for the two values?

pkg/reconciliation/constructor.go
```diff
@@ -62,11 +63,23 @@ func setOperatorProgressStatus(rc *ReconciliationContext, newState api.ProgressS
...
+	if metav1.HasAnnotation(rc.Datacenter.ObjectMeta, api.UpdateAllowedAnnotation) && rc.Datacenter.Annotations[api.UpdateAllowedAnnotation] == "once" {
+		// remove the annotation
+		patch = client.MergeFrom(rc.Datacenter.DeepCopy())
+		delete(rc.Datacenter.ObjectMeta.Annotations, api.UpdateAllowedAnnotation)
+		if err := rc.Client.Patch(rc.Ctx, rc.Datacenter, patch); err != nil {
+			rc.ReqLogger.Error(err, "error removing the allow-upgrade=once annotation")
+			return err
+		}
+	}
+
```
olim7t, line 83: [question] Why not apply the two changes in a single patch? Is it to make sure that the new state is still applied if removing the annotation fails?

olim7t (commented):
Looks great so far. I tested an operator upgrade in a local cluster and everything behaves as expected.
I'll hold off approval for now since it seems things are still in the air regarding the config secret.

pkg/reconciliation/constructor.go
burmanm: They target different objects. One is to the CassandraDatacenter, one is to the subresource, CassandraDatacenter/status. In the client, they're treated as two different objects.

CHANGELOG.md
burmanm: ReconciliationContext sadly doesn't know where the update came from as we don't pass the reconcile trigger inside the reconciliation process. That information is only available in the controller's process, where it gets Watch events and transforms those to Reconcile objects.

I don't think it's even easily possible to pass that information from there. For config secret we would need check some hashes to compare if it was changed from last time or something equivalent.

apis/cassandra/v1beta1/cassandradatacenter_types.go
olim7t, line 73: [suggestion] I also think we should pick a better name for this annotation. `allow-upgrade` is too generic, it doesn't capture the fact that we are only considering the case when the object was not modified.

Maybe something along the lines of `allow-changes-on-operator-upgrade`?

olim7t (approved):
Approving since the config secret issue can't be easily addressed.

Reminder to consider extracting constants for the annotation value, and one more comment below.

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: I renamed it to ``cassandra.datastax.com/autoupdate-spec`` which seemed to be close to what many Kubernetes "known annotations" seem to follow, such as ``rbac.authorization.kubernetes.io/autoupdate`` and ``apf.kubernetes.io/autoupdate-spec``, since their behavior is quite close to what we want.

I don't want this to be "upgrade of operator" only annotation, but instead dedicate the same behavior to all StatefulSet updates the operator might make automatically, such as removing user-made changes to the StatefulSets.

olim7t (approved):
The task looks good, I've verified that it sets the annotation to `once` on a DC that didn't have it, and completes immediately on a DC that has `always`.

PR #634, opened 2024-04-12
Allow replacenode task to replace nodes that have Cassandra crashed

### LINKED ISSUE #628: Allow replace nodes to replace pods in any state
### What is missing?

Currently our replace nodes task expects the pod to have started to correctly before it starts to replace it. That means ignoring the startTime and any async process verifications. Just simply replace it.

### Why is this needed?

Often this isn't the case, but the pod could enter infinite loop on starting and not stay up at all - or it could go up, but the FeatureSet calls fail with error codes that prevent replacing.

burmanm:
**What this PR does**:
Modifies the replace pod process to be able to replace a node that has no longer a running Cassandra pod. Also, add a test to verify the PVs that the pod uses are different from previous ones in the replace node e2e test.

Fixes #628

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/internal/controllers/control/cassandratask_controller.go
+++ b/internal/controllers/control/cassandratask_controller.go
@@ -296,8 +296,6 @@ JobDefinition:
 			completed = taskConfig.Completed
 			break JobDefinition
 		case api.CommandReplaceNode:
-			// Should we do it here? This is going to target just a single node in any case
-			// Validate the pod is a known one and get it
 			r.replace(taskConfig)
 		case api.CommandUpgradeSSTables:
 			upgradesstables(taskConfig)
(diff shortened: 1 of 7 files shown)


internal/controllers/control/cassandratask_controller.go
```diff
@@ -288,13 +289,15 @@ JobDefinition:
...
-			res, err = r.restartSts(ctx, sts, taskConfig)
+			res, err = r.restartSts(taskConfig, sts)
 			if err != nil {
 				return ctrl.Result{}, err
 			}
 			completed = taskConfig.Completed
 			break JobDefinition
 		case api.CommandReplaceNode:
+			// Should we do it here? This is going to target just a single node in any case
+			// Validate the pod is a known one and get it
```
adejanovski: **question:** not sure what you want to express here? Is there some check you want to do here that you're doing somewhere else atm?

adejanovski (approved):
Works great!

PR #673, opened 2024-07-08
Allow parallel restart of all already bootstrapped nodes

### LINKED ISSUE #646: Simplify the recovery when an existing node crashes during an expansion
There can be cases where a scale up operation would be blocked by another crashlooping pod.
In this case the new pod will have the "Starting" label, which prevents the pre-existing pod to come back up after a fix is applied (for example when it's rescheduled on a new worker).

One way we could solve this would be to allow pods to start right away if they host Cassandra nodes that have already bootstrapped in the past, and if they're not part of a replacement.
This way we could have faster startups overall while still protecting ourselves from concurrent bootstraps.

burmanm:
**What this PR does**:
If ``cassandra.datastax.com/allow-parallel-starts: true`` is set, the cass-operator will check first for all nodes that have previously bootstrapped but are down and then try to restart them.

This happens before we try to start any other node in the normal process.

Fixes #646 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -606,7 +606,7 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
 	}
 
 	// step 0 - fastpath
-	if metav1.HasAnnotation(rc.Datacenter.ObjectMeta, api.AllowParallelStartsAnnotations) && rc.Datacenter.Annotations[api.AllowParallelStartsAnnotations] == "true" {
+	if seedCount > 0 && metav1.HasAnnotation(rc.Datacenter.ObjectMeta, api.AllowParallelStartsAnnotations) && rc.Datacenter.Annotations[api.AllowParallelStartsAnnotations] == "true" {
 		notReadyPods, err := rc.startBootstrappedNodes(endpointData)
 		if err != nil {
 			return result.Error(err)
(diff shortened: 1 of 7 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -1753,6 +1772,29 @@ func (rc *ReconciliationContext) findStartingNodes() (bool, bool, error) {
...
+
+	startingNodes := false
+
+	for _, pod := range rc.dcPods {
+		if _, ok := rc.Datacenter.Status.NodeStatuses[pod.Name]; ok {
+			// Verify pod is not going to be replaced
+			if utils.IndexOfString(rc.Datacenter.Status.NodeReplacements, pod.Name) > -1 {
+				continue
+			}
+			notReady, err := rc.startNode(pod, false, endpointData)
```
adejanovski, line 1786: **Issue:** We need a little more orchestration here to deal with the seed nodes. We need at least one seed up and running so that the subsequent nodes can start correctly.
I ran a test where I stopped a DC and then restarted it, and all pods got stuck on startup with the following error:
```
ERROR [main] 2024-07-09 11:58:35,569 CassandraDaemon.java:897 - Exception encountered during startup: The seed provider lists no seeds.
```

adejanovski (changes requested):

pkg/reconciliation/reconcile_racks.go
burmanm: Changed so that the fast path is skipped with the first node and normal seed processing is done.

adejanovski (approved):
Looks great. Could you take a look at the failing tests before merging?
Thanks!

burmanm:
Both failing tests are expected and one requires a fix in the mgmt-api (it's not cass-operator failure, but the fix is not released yet).

PR #686, opened 2024-08-21
truststore reload task

### LINKED ISSUE #651: Enable tsreload from cass-operator
### What is missing?
Management API will [soon](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/483) add a new endpoint which forces a reload of the truststore. This is to enable new CA certs to be added to the node without a full restart.

I'm not sure what the best way to access this functionality is, but I presume it should probably be called from cass-operator, since that is where logic to manage the nodes usually resides. 

I can see two options:

1. We add an annotation to the pod (similar to the restart annotations that we normally add) and cass-operator picks this up before calling the management API endpoint on the specific node that needs a reload. 
1. We create a new CassandraTask which does the same.

burmanm what are your thoughts on this? Any preference?

### Why is this needed?
We have CA reloading problems.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-7) by [Unito](https://www.unito.io)
┆Reviewer: Michael Burman
┆Fix Versions: 2024-10
┆Issue Number: CASS-7

Miles-Garnsey:
**What this PR does**:

Fixes #651 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -265,4 +265,3 @@ kubeconfig
 # For GHA testing
 .act/
 .actrc
-*/build/
--- a/.gitignore
+++ b/.gitignore
@@ -265,3 +265,4 @@ kubeconfig
 # For GHA testing
 .act/
 .actrc
+*/build/
--- a/apis/control/v1alpha1/cassandratask_types.go
+++ b/apis/control/v1alpha1/cassandratask_types.go
@@ -81,6 +81,7 @@ const (
 	CommandGarbageCollect  CassandraCommand = "garbagecollect"
 	CommandFlush           CassandraCommand = "flush"
 	CommandRefresh         CassandraCommand = "refresh"
+	CommandTSReload        CassandraCommand = "tsreload"
 )
 
 type CassandraJob struct {
--- a/apis/control/v1alpha1/cassandratask_types.go
+++ b/apis/control/v1alpha1/cassandratask_types.go
@@ -81,7 +81,7 @@ const (
 	CommandGarbageCollect  CassandraCommand = "garbagecollect"
 	CommandFlush           CassandraCommand = "flush"
 	CommandRefresh         CassandraCommand = "refresh"
-	CommandTSReload        CassandraCommand = "tsreload"
+	CommandInodeTSReload   CassandraCommand = "inodetsreload"
 )
 
 type CassandraJob struct {
--- a/apis/control/v1alpha1/cassandratask_types.go
+++ b/apis/control/v1alpha1/cassandratask_types.go
@@ -81,7 +81,7 @@ const (
 	CommandGarbageCollect  CassandraCommand = "garbagecollect"
 	CommandFlush           CassandraCommand = "flush"
 	CommandRefresh         CassandraCommand = "refresh"
-	CommandInodeTSReload   CassandraCommand = "inodetsreload"
+	CommandTSReload        CassandraCommand = "tsreload"
 )
 
 type CassandraJob struct {
--- a/internal/controllers/control/cassandratask_controller.go
+++ b/internal/controllers/control/cassandratask_controller.go
@@ -317,6 +317,8 @@ JobDefinition:
 			}
 			completed = taskConfig.Completed
 			break JobDefinition
+		case api.CommandTSReload:
+			tsReload(taskConfig)
 		default:
 			err = fmt.Errorf("unknown job command: %s", job.Command)
 			return ctrl.Result{}, err
--- a/internal/controllers/control/cassandratask_controller.go
+++ b/internal/controllers/control/cassandratask_controller.go
@@ -818,6 +818,9 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc 
 					if !features.Supports(taskConfig.SyncFeature) {
 						logger.Error(err, "Pod doesn't support this feature", "Pod", pod, "Feature", taskConfig.SyncFeature)
 						jobStatus.Status = podJobError
+						failed++
+						errMsg = fmt.Sprintf("Pod %s doesn't support %s feature", pod.Name, taskConfig.SyncFeature)
+						return
 					}
 				}
 
--- a/internal/controllers/control/cassandratask_controller.go
+++ b/internal/controllers/control/cassandratask_controller.go
@@ -801,6 +801,16 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc 
 				return ctrl.Result{}, failed, completed, errMsg, err
 			}
 
+			if taskConfig.SyncFeature != "" {
+				if !features.Supports(taskConfig.SyncFeature) {
+					logger.Error(err, "Pod doesn't support this feature", "Pod", pod, "Feature", taskConfig.SyncFeature)
+					jobStatus.Status = podJobError
+					failed++
+					errMsg = fmt.Sprintf("Pod %s doesn't support %s feature", pod.Name, taskConfig.SyncFeature)
+					return ctrl.Result{}, failed, completed, errMsg, err
+				}
+			}
+
 			pod := pod
 
 			go func() {
@@ -813,16 +823,6 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc 
 					// Remove the value from the jobRunner
 					<-jobRunner
 				}()
-
-				if taskConfig.SyncFeature != "" {
-					if !features.Supports(taskConfig.SyncFeature) {
-						logger.Error(err, "Pod doesn't support this feature", "Pod", pod, "Feature", taskConfig.SyncFeature)
-						jobStatus.Status = podJobError
-						failed++
-						errMsg = fmt.Sprintf("Pod %s doesn't support %s feature", pod.Name, taskConfig.SyncFeature)
-						return
-					}
-				}
 
 				if err = taskConfig.SyncFunc(nodeMgmtClient, &pod, taskConfig); err != nil {
 					// We only log, nothing else to do - we won't even retry this pod
--- a/internal/controllers/control/cassandratask_controller.go
+++ b/internal/controllers/control/cassandratask_controller.go
@@ -98,6 +98,7 @@ type TaskConfiguration struct {
 	// Execution functionality per pod
 	AsyncFeature httphelper.Feature
 	AsyncFunc    AsyncTaskExecutorFunc
+	SyncFeature  httphelper.Feature
 	SyncFunc     SyncTaskExecutorFunc
 	PodFilter    PodFilterFunc
 
@@ -649,13 +650,9 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc 
 		if !taskConfig.Filter(&pod) {
 			continue
 		}
-
-		features := &httphelper.FeatureSet{}
-		if taskConfig.AsyncFeature != "" {
-			features, err = nodeMgmtClient.FeatureSet(&pod)
-			if err != nil {
-				return ctrl.Result{}, failed, completed, errMsg, err
-			}
+		features, err := nodeMgmtClient.FeatureSet(&pod)
+		if err != nil {
+			return ctrl.Result{}, failed, completed, errMsg, err
 		}
 
 		if pod.Annotations == nil {
@@ -816,6 +813,13 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc 
 					// Remove the value from the jobRunner
 					<-jobRunner
 				}()
+
+				if taskConfig.SyncFeature != "" {
+					if !features.Supports(taskConfig.SyncFeature) {
+						logger.Error(err, "Pod doesn't support this feature", "Pod", pod, "Feature", taskConfig.SyncFeature)
+						jobStatus.Status = podJobError
+					}
+				}
 
 				if err = taskConfig.SyncFunc(nodeMgmtClient, &pod, taskConfig); err != nil {
 					// We only log, nothing else to do - we won't even retry this pod
--- a/internal/controllers/control/cassandratask_controller.go
+++ b/internal/controllers/control/cassandratask_controller.go
@@ -785,6 +785,16 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc 
 				return ctrl.Result{}, failed, completed, errMsg, err
 			}
 
+			if taskConfig.SyncFeature != "" {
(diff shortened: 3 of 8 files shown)


burmanm:
Try to use personal fork next time so we don't have unnecessary branches here that must be removed.

internal/controllers/control/jobs.go
```diff
@@ -442,6 +442,22 @@ func (r *CassandraTaskReconciler) refreshDatacenter(ctx context.Context, dc *cas
 		}
 	}
 	return ctrl.Result{RequeueAfter: JobRunningRequeue}, nil
+
+}
+
+// Compaction functionality
```
burmanm: Incorrect comment.

internal/controllers/control/jobs.go
```diff
@@ -442,6 +442,22 @@ func (r *CassandraTaskReconciler) refreshDatacenter(ctx context.Context, dc *cas
...
+
+func TSReloadAsync(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) (string, error) {
```
burmanm: What is this function? It doesn't do what it says.

.gitignore
```diff
@@ -264,4 +264,5 @@ kubeconfig
 
 # For GHA testing
 .act/
-.actrc
\ No newline at end of file
+.actrc
+*/build/
```
burmanm: Why? We already have /build there which correctly ignores the logs.

internal/controllers/control/jobs.go
Miles-Garnsey: This wasn't ready for review, but let's go anyway... I removed this, we don't need it.

.gitignore
Miles-Garnsey: Sure, I can remove this. There were changes to the gitignore which meant some of my local test files were turning up, but I can put them in `/build`.

internal/controllers/control/jobs.go
Miles-Garnsey: Fixed, thanks.

Miles-Garnsey:
@burmanm I have this working now, and have also accounted for the fact that this is the only sync task which may fail due to being unsupported by the targeted management API instance. Can you please let me know if you're now happy with it?

apis/control/v1alpha1/cassandratask_types.go
```diff
@@ -81,6 +81,7 @@ const (
 	CommandGarbageCollect  CassandraCommand = "garbagecollect"
 	CommandFlush           CassandraCommand = "flush"
 	CommandRefresh         CassandraCommand = "refresh"
+	CommandInodeTSReload   CassandraCommand = "inodetsreload"
```
burmanm: issue: Wasn't this tsreload? inode means nothing in cass-operator, we always use "internode" in its full form and in some cases (metrics) the actual "inode" which means something quite different. I'm not sure why we would try to save few characters here. Rather, just remove "inode" entirely if needed and we can later extend this feature to include client / server selection through args.

internal/controllers/control/cassandratask_controller_test.go
```diff
@@ -622,10 +622,25 @@ var _ = Describe("CassandraTask controller tests", func() {
 				Expect(callDetails.URLCounts["/api/v0/ops/executor/job"]).To(Equal(0))
 				Expect(callDetails.URLCounts["/api/v0/metadata/versions/features"]).To(BeNumerically(">", 1))
 
+				// verifyPodsHaveAnnotations(testNamespaceName, string(task.UID))
+				Expect(completedTask.Status.Succeeded).To(BeNumerically(">=", 1))
+			})
+			It("Runs a ts reload task against a pod", func() {
```
burmanm, line 667: issue: Well, this is not actually broken in the implementation, but this test doesn't actually verify your Sync check is working. As seen here, the Feature is not in the list of available Features: 

https://github.com/k8ssandra/cass-operator/blob/3e146206ec9493e1ec87d018ea3becae27d799c2/pkg/httphelper/server_test_utils.go#L15

But this succeeds, because the test is ran against a fake server without details endpoint. And because your SyncFeature is not set in the jobs.go.

internal/controllers/control/cassandratask_controller.go
```diff
@@ -815,6 +814,13 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc
 					<-jobRunner
 				}()
 
+				if taskConfig.SyncFeature != "" {
```
burmanm: issue: This feature is not tested, not for actually working or for regression. It would require

a) Test that checks that existing jobs without SyncFeature set still work as is
b) Test that checks that new job with SyncFeature (but no server support) is correctly failing
c) Test that checks that new job with SyncFeature and server support is correctly called

Easiest way would be to modify your existing test to verify against Feature endpoint and add SyncFeature in its definition and run it. The a) is taken care of running a sync against featureset endpoint and still working and b) requires a new test.

internal/controllers/control/cassandratask_controller.go
```diff
@@ -815,6 +814,13 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc
...
+					if !features.Supports(taskConfig.SyncFeature) {
+						logger.Error(err, "Pod doesn't support this feature", "Pod", pod, "Feature", taskConfig.SyncFeature)
+						jobStatus.Status = podJobError
+					}
```
burmanm: issue: This would still run the SyncFunc  in line 824 as nothing prevents it we simply log and mark it failed and let the next section cause the failure.

burmanm (changes requested):
Requires testing and fixing the SyncFeature

internal/controllers/control/cassandratask_controller.go
Miles-Garnsey: Thanks, fixed.

internal/controllers/control/cassandratask_controller_test.go
Miles-Garnsey: Interesting, thanks. I will investigate this some more, it seemed like this was too easy,.

apis/control/v1alpha1/cassandratask_types.go
Miles-Garnsey: Done.

internal/controllers/control/cassandratask_controller_test.go
Miles-Garnsey: I think I've fixed this.

internal/controllers/control/cassandratask_controller.go
Miles-Garnsey: I think all of these are now present.

Miles-Garnsey:
@burmanm I think I've now addressed all of your feedback.

internal/controllers/control/cassandratask_controller.go
```diff
@@ -815,6 +814,16 @@ func (r *CassandraTaskReconciler) reconcileEveryPodTask(ctx context.Context, dc
 					<-jobRunner
 				}()
 
+				if taskConfig.SyncFeature != "" {
+					if !features.Supports(taskConfig.SyncFeature) {
+						logger.Error(err, "Pod doesn't support this feature", "Pod", pod, "Feature", taskConfig.SyncFeature)
+						jobStatus.Status = podJobError
+						failed++
+						errMsg = fmt.Sprintf("Pod %s doesn't support %s feature", pod.Name, taskConfig.SyncFeature)
+						return
```
burmanm: This is a bit trickier to implement as this state is not persisted now at all. Remember, we're inside a goroutine so nothing outside is usable. So failed++ is not doing anything and nor is returning here (as it terminates the goroutine). The jobStatus is stored on the pod and that's done on the line 848 so we need to jump there. 

I wonder if you intended this check to happen at line 779 with similar behavior as the case "if taskConfig.SyncFunc == nil {" ?

internal/controllers/control/cassandratask_controller_test.go
```diff
@@ -622,6 +634,46 @@ var _ = Describe("CassandraTask controller tests", func() {
...
+				taskKey, task := buildTask(api.CommandTSReload, testNamespaceName)
+				task.Spec.Jobs[0].Arguments.PodName = fmt.Sprintf("%s-%s-r0-sts-0", clusterName, testDatacenterName)
+				Expect(k8sClient.Create(context.Background(), task)).Should(Succeed())
+
+				completedTask := waitForTaskFailed(taskKey)
+
+				Expect(callDetails.URLCounts["/api/v0/ops/node/encryption/internode/truststore/reload"]).To(Equal(0)) // This doesn't get called because the test of whether the feature exists doesn't pass. The features endpoint doesn't exist in this mock server.
+
+				// verifyPodsHaveAnnotations(testNamespaceName, string(task.UID))
+				Expect(completedTask.Status.Succeeded).To(BeNumerically("==", 0))
```
burmanm, line 651: suggestion: Failed should be >= 1

internal/controllers/control/cassandratask_controller_test.go
```diff
@@ -622,6 +634,46 @@ var _ = Describe("CassandraTask controller tests", func() {
...
+			It("Runs a ts reload task against a pod", func() {
+				By("Creating a task for tsreload")
+				taskKey, task := buildTask(api.CommandTSReload, testNamespaceName)
+				task.Spec.Jobs[0].Arguments.PodName = fmt.Sprintf("%s-%s-r0-sts-0", clusterName, testDatacenterName)
+				Expect(k8sClient.Create(context.Background(), task)).Should(Succeed())
+
+				completedTask := waitForTaskCompletion(taskKey)
+
+				Expect(callDetails.URLCounts["/api/v0/ops/node/encryption/internode/truststore/reload"]).To(Equal(1))
+				Expect(callDetails.URLCounts["/api/v0/metadata/versions/features"]).To(BeNumerically(">", 0)) // This doesn't get called because the test of whether the feature exists doesn't pass.
```
burmanm, line 676: question: But shouldn't it always be called before calling the method above?

pkg/httphelper/client.go
```diff
@@ -585,6 +586,33 @@ func (client *NodeMgmtClient) CallCompactionEndpoint(pod *corev1.Pod, compactReq
...
+	if err != nil {
+		return err
+	}
+
+	request := nodeMgmtRequest{
+		endpoint: "/api/v0/ops/node/encryption/internode/truststore/reload",
+		host:     podHost,
+		port:     podPort,
+		method:   http.MethodPost,
+		timeout:  60 * time.Second,
```
burmanm, line 605: question: Do you happen to know how long this operation usually takes? If it's closer to 60 seconds, we probably should have an async endpoint done also in the future instead of relying cass-operator to do the async operation.

pkg/httphelper/server_test_utils.go
```diff
@@ -128,10 +129,32 @@ func FakeExecutorServerWithDetailsFails(callDetails *CallDetails) (*httptest.Ser
...
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+	}))
+}
+
 func FakeServerWithoutFeaturesEndpoint(callDetails *CallDetails) (*httptest.Server, error) {
 	return FakeMgmtApiServer(callDetails, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.Method == http.MethodPost && (r.URL.Path == "/api/v0/ops/keyspace/cleanup" || r.URL.Path == "/api/v0/ops/tables/sstables/upgrade" || r.URL.Path == "/api/v0/ops/node/drain" || r.URL.Path == "/api/v0/ops/tables/flush" || r.URL.Path == "/api/v0/ops/tables/garbagecollect" || r.URL.Path == "/api/v0/ops/tables/compact") {
 			w.WriteHeader(http.StatusOK)
+		} else if r.Method == http.MethodPost && r.URL.Path == "/api/v0/ops/node/encryption/internode/truststore/reload" {
```
burmanm, line 156: We probably shouldn't have this option as it should only exists in versions with features endpoint.

pkg/httphelper/server_test_utils.go
```diff
@@ -128,10 +129,32 @@ func FakeExecutorServerWithDetailsFails(callDetails *CallDetails) (*httptest.Ser
...
+func FakeServerWithSyncFeaturesEndpoint(callDetails *CallDetails) (*httptest.Server, error) {
+	return FakeMgmtApiServer(callDetails, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := url.ParseQuery(r.URL.RawQuery)
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+		}
+		if r.Method == http.MethodGet && r.RequestURI == "/api/v0/metadata/versions/features" {
+			w.WriteHeader(http.StatusOK)
+			_, err = w.Write([]byte(featuresReply))
+		} else if r.Method == http.MethodPost && r.URL.Path == "/api/v0/ops/node/encryption/internode/truststore/reload" {
```
burmanm, line 141: suggestion: This could be merged with the FakeExecutorServerWithDetails but just return OK instead of a jobId.

internal/controllers/control/cassandratask_controller.go
Miles-Garnsey: Makes sense, fixed...

internal/controllers/control/cassandratask_controller_test.go
Miles-Garnsey: Does the order of the statements matter in this case? It doesn't make a difference to the semantics.

pkg/httphelper/client.go
Miles-Garnsey: I have no idea, no.

pkg/httphelper/server_test_utils.go
Miles-Garnsey: This is there so that we can verify it fails when there is no features endpoint.

Miles-Garnsey: i.e. [this](https://github.com/k8ssandra/cass-operator/blob/feature/tsreload-task/internal/controllers/control/cassandratask_controller_test.go#L640-L652) test.

pkg/httphelper/server_test_utils.go
Miles-Garnsey: I think the intent is different because this is a sync task and therefore doesn't return jobID. The behaviour is different which should probably be conveyed in the test...

internal/controllers/control/cassandratask_controller_test.go
Miles-Garnsey: Good idea, done.

burmanm (approved):

PR #688, opened 2024-08-22
Fix a bug that would cause a crash if StorageClassName was left to nil

### LINKED ISSUE #687: cass-operator may crash if storageClassName is not set
### What happened?

If `Spec.StorageConfig.CassandraDataVolumeClaimSpec.StorageClassName` is not set on the datacenter CR, `StorageClassName` will be a null pointer.
`StorageClassName` is getting dereferenced here https://github.com/k8ssandra/cass-operator/blame/master/pkg/reconciliation/reconcile_datacenter.go#L191 without a null check before.

If storageClassName is null, the operator will segfault.

I expect cass-operator to handle the case when `storageClassName` is not set by the user.
The operator should try to fetch the `default` storage class instead and call `storageClass()` with the name of the default storage class.

I observed the issue when testing restoration.
I believe the issue will happen when storage expansion is triggered and when storage class name is not set.

I'm not sure why storage expansion got triggered during restoration.

### cass-operator version

1.22

### Kubernetes version

1.29

### Method of installation

helm

_No response_

burmanm:
**What this PR does**:
* Changes handling of PersistentVolumeClaimSpec's *string in storageExpansion() method. 

Fixes #687 

**Checklist**
- [ ] Changes manually tested


garrynigel:
thanks @burmanm , Could we escalate and get this merged quickly?

pkg/reconciliation/reconcile_racks.go
```diff
@@ -248,6 +241,11 @@ func (rc *ReconciliationContext) CheckVolumeClaimSizes(statefulSet, desiredSts *
 				return result.Error(fmt.Errorf(msg))
 			}
 
+			supportsExpansion, err := rc.storageExpansion()
+			if err != nil {
+				return result.Error(err)
+			}
+
```
olim7t, line 248: [question] Are we moving this inside the loop on the assumption that volume expansions will be rare?

olim7t (approved):

pkg/reconciliation/reconcile_racks.go
burmanm: Yes and this check is irrelevant in our other use cases right now.

garrynigel:
@burmanm thanks for merging this!! But I noticed in storage classes provided by cloud provider the default value is stored in  annotations and not labels, so this always returns empty in my case.

See https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/#changing-the-default-storageclass


I modified the lines 166 to 175 slightly and only then it seems to be working.

```
	var defaultScName string
	var defaultScCount int
	
	if err := r.List(ctx, storageClassList); err != nil {
		return err
	}
	for i := range storageClassList.Items {
		storageClass := storageClassList.Items[i]
		if storageClass.Annotations != nil && storageClass.Annotations["storageclass.kubernetes.io/is-default-class"] == "true" {
			defaultScName = storageClass.Name
			defaultScCount++
		}
	
	if defaultScCount == 0 {
		return fmt.Errorf("no default storage class found, please specify StorageClassName in the CassandraDatacenter spec")
	} else if defaultScCount > 1 {
		return fmt.Errorf("found multiple default storage classes, please specify StorageClassName in the CassandraDatacenter spec")
	}
```

PR #690, opened 2024-08-23
Allow k8ssandra.io labels and annotations in services config (fixes #689)

### LINKED ISSUE #689: Revisit label and annotation restrictions in AdditionalServiceConfig
cass-operator does not allow label and annotation keys that start with `cassandra.datastax.com` or `k8ssandra.io` in `AdditionalServiceConfig` ([source](https://github.com/k8ssandra/cass-operator/blob/7199188bc3133157b25d09f1b66bb8d7c318b851/apis/cassandra/v1beta1/cassandradatacenter_webhook.go#L343)). This is intended to prevent collisions with the labels that the operator uses internally.

However, [k8ssandra-operator](https://github.com/k8ssandra/k8ssandra-operator), which relies on cass-operator to manage its datacenters, also uses the `k8ssandra.io` prefix. There are cases where it might need to label services for its own internal needs, but this is currently not possible.

olim7t:
**What this PR does**:

Fixes #689

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 ## unreleased
 
 * [BUGFIX] [#687](https://github.com/k8ssandra/cass-operator/issues/687) Prevent a crash when when StorageClassName was not set in the CassandraDataVolumeClaimSpec
+* [CHANGE] [#689](https://github.com/k8ssandra/cass-operator/issues/689) Allow k8ssandra.io labels and annotations in services config
 
 ## v1.22.0
 
--- a/apis/cassandra/v1beta1/cassandradatacenter_webhook.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_webhook.go
@@ -34,8 +34,7 @@ import (
 )
 
 const (
-	datastaxPrefix  string = "cassandra.datastax.com"
-	k8ssandraPrefix string = "k8ssandra.io"
+	datastaxPrefix string = "cassandra.datastax.com"
 )
 
 var log = logf.Log.WithName("api")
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 
+* [CHANGE] [#689]((https://github.com/k8ssandra/cass-operator/issues/689) Allow some internal labels and annotations in services config
```
burmanm: nitpick: There's a double (( which breaks the markdown rendering

apis/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -38,6 +39,10 @@ const (
 	k8ssandraPrefix string = "k8ssandra.io"
 )
 
+// Some exceptions to the reserved prefixes above:
+var allowedServiceLabels = []string{"k8ssandra.io/cluster-name", "k8ssandra.io/cluster-namespace"}
```
burmanm: suggestion: It would be nice to have these prefixes as consts (while []string can't be const, the individual items can be).

But alternatively, I was going to suggest that if the CassandraDatacenter object itself has owner reference to another object, it would be allowed to have different set of labels/annotations. But we don't apparently do that in the k8ssandra-operator (and control-plane <-> data-plane issue), so that was not going to work. 

That said, cass-operator does not use k8ssandra.io prefixes for CassandraDatacenter (we only use that prefix in the CassandraTask), so perhaps that check can be removed since k8ssandra 1.x is also deprecated. All the labels/annotations are cassandra.datastax.com in cass-operator.

olim7t: > That said, cass-operator does not use k8ssandra.io prefixes for CassandraDatacenter (we only use that prefix in the CassandraTask), so perhaps that check can be removed since k8ssandra 1.x is also deprecated. All the labels/annotations are cassandra.datastax.com in cass-operator.

Makes sense, and it's an even better way to address the issue. I'll revisit this PR.

CHANGELOG.md
olim7t: Good catch, fixed.

burmanm (approved):
Just needs rebase for changelog

PR #698, opened 2024-09-04
Change default labels to use Kubernetes resource name

### LINKED ISSUE #691: Revert label override values
### What is missing?
The original behavior was that ``cassandra.datastax.com/datacenter`` has a value that's identical to the ``CassandraDatacenter.ObjectMeta.Name``. In PR #475 the behavior was changed to include ``CassandraDatacenter.DatacenterName()`` which could be either the ``.Name`` or ``.Spec.DatacenterName``. 

My proposal is to revert this part of the PR with basically the following fix in cass-operator:

```diff
diff --git a/apis/cassandra/v1beta1/cassandradatacenter_types.go b/apis/cassandra/v1beta1/cassandradatacenter_types.go
index ec9aa20..fc4bc2d 100644
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -596,7 +596,7 @@ func (dc *CassandraDatacenter) SetCondition(condition DatacenterCondition) {
 // GetDatacenterLabels ...
 func (dc *CassandraDatacenter) GetDatacenterLabels() map[string]string {
        labels := dc.GetClusterLabels()
-       labels[DatacenterLabel] = CleanLabelValue(dc.DatacenterName())
+       labels[DatacenterLabel] = CleanLabelValue(dc.GetName())
        return labels
 }
```

### Why is this needed?
This is because right now we're unable to use the labels for querying the parent objects. If we look at a pod/service/StatefulSet/any object when override name is in use it has a label like: ``cassandra.datastax.com/datacenter=overrideName`` yet using the Kubernetes API this label value is entirely useless. We can't fetch the actual ``CassandraDatacenter`` with this information at all.

It also does not necessarily match the value in Cassandra itself either as the value stored in the labels is cleaned up removing certain characters. Thus, both the user and the systems are unable to use these values for anything. And we need this feature, we need to be able to link resources from bottom to top so we can fetch the related resources. Right now only the top to bottom is doable since CassandraDatacenter knows what labels we created to sub resources, but sub resources do not know what labels the parts above (or next to) have. 

Reverting this change requires some work also when updating since we do not allow StatefulSet updates automatically without user consent. We can however make an exception and update labels, since this doesn't require a rolling restart of the pods as all labels can be changed in Pods/StatefulSets without rolling update.

For k8ssandra-operator, we are in a more complex scenario. There are already some resources that use the ``.Name`` (like Prometheus ServiceMonitors) and some that use ``.DatacenterName()``. This will require some manual upgrade process to enforce the correct format we want everywhere. While most of them should be automatically updated in the upgrade process, there are some that require manual work, namely the MedusaBackup objects as they were created with the old values and are not actively maintained. 

The k8ssandra-operator also has the ``k8ssandra.io/cluster-name`` which is different from ``cassandra.datastax.com/cluster``. The latter is always using the ``.Spec.ClusterName``and there's no need to change this behavior. However, for k8ssandra-operator we probably want to enforce the ``.Name`` instead. That gives us both values to use for cluster seeking operations.

One raised question was that the update has issues knowing which value we have in the datacenter label when updating from previous versions. This is always problematic no matter if we rollback or not, since there are already versions of cass-operator & k8ssandra-operator adding either value, depending on where you're updating from. 

It might feel painful, but I think we should do the right thing here and revert the change to simplify our labels usage as we ourselves can't even keep track of what values are stored where. If we always keep the Kubernetes names, it's very clear. It's also what we internally always process, whatever is in Cassandra is not really relevant to our management processes.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-4) by [Unito](https://www.unito.io)
┆Reviewer: Miles Garnsey
┆Fix Versions: 2024-10,2024-11
┆Issue Number: CASS-4

### LINKED ISSUE #699: allow-updatespec once can be removed prematurely
### What happened?
Three rack system with one node on each rack was requiring an update can sometimes cause the update annotation to be removed before all the StatefulSets have been updated. For example, in this log we update the r1 15:55:48 and then in 15:56:55 we notice that r2 would have required an update also, but the annotation was removed before we processed this part.

```
2024-09-05T15:52:51.101Z	INFO	update is blocked, but statefulset needs an update. Marking datacenter as requiring update.	{"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "CassandraDatacenter": {"name":"dc1","namespace":"test-upgrade-operator"}, "namespace": "test-upgrade-operator", "name": "dc1", "reconcileID": "f923057c-694c-410f-bd99-ff77f0eaf243", "namespace": "test-upgrade-operator", "datacenterName": "my-super-dc", "clusterName": "cluster1", "rackName": "r1"}
2024-09-05T15:52:56.631Z	INFO	update is blocked, but statefulset needs an update. Marking datacenter as requiring update.	{"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "CassandraDatacenter": {"name":"dc1","namespace":"test-upgrade-operator"}, "namespace": "test-upgrade-operator", "name": "dc1", "reconcileID": "3ef08510-5548-413f-8c7e-95e841c9474d", "namespace": "test-upgrade-operator", "datacenterName": "my-super-dc", "clusterName": "cluster1", "rackName": "r1"}
2024-09-05T15:55:48.095Z	INFO	statefulset needs an update	{"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "CassandraDatacenter": {"name":"dc1","namespace":"test-upgrade-operator"}, "namespace": "test-upgrade-operator", "name": "dc1", "reconcileID": "2ce2ae60-7928-4c90-96b4-84e5fa1ca6a0", "namespace": "test-upgrade-operator", "datacenterName": "my-super-dc", "clusterName": "cluster1", "rackName": "r1"}
2024-09-05T15:56:55.263Z	INFO	update is blocked, but statefulset needs an update. Marking datacenter as requiring update.	{"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "CassandraDatacenter": {"name":"dc1","namespace":"test-upgrade-operator"}, "namespace": "test-upgrade-operator", "name": "dc1", "reconcileID": "5db48d2d-cfa9-4662-be7d-92662768cd93", "namespace": "test-upgrade-operator", "datacenterName": "my-super-dc", "clusterName": "cluster1", "rackName": "r2"}
2024-09-05T15:57:02.976Z	INFO	update is blocked, but statefulset needs an update. Marking datacenter as requiring update.	{"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "CassandraDatacenter": {"name":"dc1","namespace":"test-upgrade-operator"}, "namespace": "test-upgrade-operator", "name": "dc1", "reconcileID": "b09c4e3d-3066-4c3a-8109-d3a0d592eeb1", "namespace": "test-upgrade-operator", "datacenterName": "my-super-dc", "clusterName": "cluster1", "rackName": "r2"}
```

And that means the datacenter is now in state where it requires an update, but the r1 is updated while r2 and r3 are not. We shouldn't get to the end of reconciliation if there's still other racks requiring an update.

All racks should be updated before we remove the annotation.

operator_upgrade test in some cases triggers this behavior.

### cass-operator version
1.22.1

### Kubernetes version
1.30

### Method of installation
_No response_

_No response_

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-63) by [Unito](https://www.unito.io)
┆Fix Versions: 2024-10,2024-11
┆Issue Number: CASS-63

burmanm:
**What this PR does**:
Changes datacenter label value to be Kubernetes resource name for Datacenter again. Provides backwards compatibility with older versions using the wrong one.

All the label names are updated after a successful reconcile (this is an existing feature). Tested with Helm installation of 1.22.1 and then updating the cass-operator image only.

Fixes #691
Fixes #699 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -488,6 +488,9 @@ type CassandraDatacenterStatus struct {
 	// This field is used to perform validation checks preventing a user from changing the override
 	// +optional
 	DatacenterName *string `json:"datacenterName,omitempty"`
+
+	// +optional
+	MetadataVersion int64 `json:"metadataVersion,omitempty"`
 }
 
 // CassandraDatacenter is the Schema for the cassandradatacenters API
@@ -599,7 +602,7 @@ func (dc *CassandraDatacenter) SetCondition(condition DatacenterCondition) {
 // GetDatacenterLabels ...
 func (dc *CassandraDatacenter) GetDatacenterLabels() map[string]string {
 	labels := dc.GetClusterLabels()
-	labels[DatacenterLabel] = CleanLabelValue(dc.DatacenterName())
+	labels[DatacenterLabel] = CleanLabelValue(dc.Name)
 	return labels
 }
 
@@ -664,19 +667,19 @@ func (dc *CassandraDatacenter) GetSeedServiceName() string {
 }
 
 func (dc *CassandraDatacenter) GetAdditionalSeedsServiceName() string {
-	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.SanitizedName() + "-additional-seed-service"
+	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.LabelResourceName() + "-additional-seed-service"
 }
 
 func (dc *CassandraDatacenter) GetAllPodsServiceName() string {
-	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.SanitizedName() + "-all-pods-service"
+	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.LabelResourceName() + "-all-pods-service"
 }
 
 func (dc *CassandraDatacenter) GetDatacenterServiceName() string {
-	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.SanitizedName() + "-service"
+	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.LabelResourceName() + "-service"
 }
 
 func (dc *CassandraDatacenter) GetNodePortServiceName() string {
-	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.SanitizedName() + "-node-port-service"
+	return CleanupForKubernetes(dc.Spec.ClusterName) + "-" + dc.LabelResourceName() + "-node-port-service"
 }
 
 func (dc *CassandraDatacenter) ShouldGenerateSuperuserSecret() bool {
@@ -973,9 +976,17 @@ func SplitRacks(nodeCount, rackCount int) []int {
 	return topology
 }
 
-// SanitizedName returns a sanitized version of the name returned by DatacenterName()
-func (dc *CassandraDatacenter) SanitizedName() string {
-	return CleanupForKubernetes(dc.DatacenterName())
+func (dc *CassandraDatacenter) DatacenterNameStatus() bool {
+	return dc.Status.DatacenterName != nil
+}
+
+// LabelResourceName returns a sanitized version of the name returned by DatacenterName()
+func (dc *CassandraDatacenter) LabelResourceName() string {
+	// If existing cluster, return dc.DatacenterName() else return dc.Name
+	if dc.DatacenterNameStatus() {
+		return CleanupForKubernetes(*dc.Status.DatacenterName)
+	}
+	return CleanupForKubernetes(dc.Name)
 }
 
 // DatacenterName returns the Cassandra DC name override if it exists,
--- a/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -261,8 +261,7 @@ type CassandraDatacenterSpec struct {
 	// CDC allows configuration of the change data capture agent which can run within the Management API container. Use it to send data to Pulsar.
 	CDC *CDCConfiguration `json:"cdc,omitempty"`
 
-	// DatacenterName allows to override the name of the Cassandra datacenter. Kubernetes objects will be named after a sanitized version of it if set, and if not metadata.name. In Cassandra the DC name will be overridden by this value.
-	// It may generate some confusion as objects created for the DC will have a different name than the CasandraDatacenter object itself.
+	// DatacenterName allows to override the name of the Cassandra datacenter. In Cassandra the DC name will be overridden by this value.
 	// This setting can create conflicts if multiple DCs coexist in the same namespace if metadata.name for a DC with no override is set to the same value as the override name of another DC.
 	// Use cautiously.
 	// +optional
--- a/apis/cassandra/v1beta1/cassandradatacenter_types_test.go
+++ b/apis/cassandra/v1beta1/cassandradatacenter_types_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"testing"
 
+	"github.com/Jeffail/gabs/v2"
 	"github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
@@ -189,4 +190,51 @@ func TestUseClientImageEnforce(t *testing.T) {
 
 		assert.True(dc.UseClientImage())
 	}
+}
+
+func TestDatacenterNoOverrideConfig(t *testing.T) {
+	assert := assert.New(t)
+	dc := CassandraDatacenter{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "dc1",
+		},
+		Spec: CassandraDatacenterSpec{
+			ClusterName: "cluster1",
+		},
+	}
+
+	config, err := dc.GetConfigAsJSON(dc.Spec.Config)
+	assert.NoError(err)
+
+	container, err := gabs.ParseJSON([]byte(config))
+	assert.NoError(err)
+
+	dataCenterInfo := container.ChildrenMap()["datacenter-info"]
+	assert.NotEmpty(dataCenterInfo)
+	assert.Equal(dc.Name, dataCenterInfo.ChildrenMap()["name"].Data().(string))
+	assert.Equal(dc.DatacenterName(), dc.Name)
+}
+
+func TestDatacenterOverrideInConfig(t *testing.T) {
+	assert := assert.New(t)
+	dc := CassandraDatacenter{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "dc1",
+		},
+		Spec: CassandraDatacenterSpec{
+			ClusterName:    "cluster1",
+			DatacenterName: "Home_Dc",
+		},
+	}
+
+	config, err := dc.GetConfigAsJSON(dc.Spec.Config)
+	assert.NoError(err)
+
+	container, err := gabs.ParseJSON([]byte(config))
+	assert.NoError(err)
+
+	dataCenterInfo := container.ChildrenMap()["datacenter-info"]
+	assert.NotEmpty(dataCenterInfo)
+	assert.Equal(dc.Spec.DatacenterName, dataCenterInfo.ChildrenMap()["name"].Data().(string))
+	assert.Equal(dc.DatacenterName(), dc.Spec.DatacenterName)
 }
--- a/config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml
+++ b/config/crd/bases/cassandra.datastax.com_cassandradatacenters.yaml
@@ -300,8 +300,7 @@ spec:
                 type: string
               datacenterName:
                 description: |-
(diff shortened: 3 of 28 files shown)


apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -596,7 +599,7 @@ func (dc *CassandraDatacenter) SetCondition(condition DatacenterCondition) {
 // GetDatacenterLabels ...
 func (dc *CassandraDatacenter) GetDatacenterLabels() map[string]string {
 	labels := dc.GetClusterLabels()
-	labels[DatacenterLabel] = CleanLabelValue(dc.DatacenterName())
+	labels[DatacenterLabel] = CleanLabelValue(dc.Name)
```
Miles-Garnsey, line 604: Question: Is `CleanLabelValue` still necessary now that we're using the Kubernetes meta.name which will be DNS compliant anyway?

Miles-Garnsey (commented):
I need to do more testing on this in several respects;

1. Does an upgrade from the previous version to this one work correctly when an override name is in place?
2. Does this work when incorporated into downstream projects (e.g. k8ssandra-operator)? If not, should we make changes to the downstream projects, ensure everything works, and only then merge the changes?

Also @burmanm can you give me an overview of how the backwards compatibility is supposed to work? I don't think this was on the original ticket and I'm curious as to whether that status field you're using is pre-existing, and where it comes from generally.

burmanm:
Yes, the operator upgrade test will deploy with override name in place (with an operator version that uses the override naming) and update from there. 

The status field is not pre-existing, yet it wouldn't matter if it did or did not. A CRD upgrade will add it and the default is treated as 0 (such as in the cases that the field was not previously populated or does not exists). If the CRD is old one, it would simply fetch too much information on every reconcile but not break anything. I did test this manually against existing cluster.

If you wish to test that, the simplest way is to deploy cass-operator with Helm and do an upgrade. For example:

```
helm install --namespace cass-operator --create-namespace k8ssandra k8ssandra/cass-operator --set global.clusterScoped=true
```

Deploy something, like the one used in operator upgrade:

```
kubectl apply -f tests/testdata/default-three-rack-three-node-dc-4x.yaml
```

And then update cass-operator to a version built from this repo:

```
helm upgrade --namespace cass-operator k8ssandra k8ssandra/cass-operator --reuse-values --set image.registry="" --set image.tag=v1.23.0-dev.3f1fd20-20240906
```

And the cluster should stay healthy and get the updated values. This process does not update the CRD, only the operator image. 

As for downstream projects, it's not really a place here to think about that (if the downstream is unable to use upstream, then the fix must happen in the downstream). That said, k8ssandra-operator uses these methods directly so that would work as soon as the first reconcile has finished here.

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: Well, not really, but they do have a bit different rules:

```
	dns1123SubdomainFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
	dns1123LabelFmt     string = "(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?"
```

So better safe than sorry I guess, especially if Kubernetes would one day update these..

Miles-Garnsey: Suggestion: I think the dns1123LabelFmt is more liberal than the dns1123SubdomainFmt, I also don't think k8s will ever change these (since they're linked to the DNS spec). 

I'd remove the extra function call just so it is clear to readers that the label is always going to be the exact meta.Name, and not something else that we've interfered with. 

This is just a suggestion though, so it isn't blocking.

Miles-Garnsey:
> As for downstream projects, it's not really a place here to think about that (if the downstream is unable to use upstream, then the fix must happen in the downstream). That said, k8ssandra-operator uses these methods directly so that would work as soon as the first reconcile has finished here.

In theory that's true, but if we merge this to `master` and then cut a cass-operator release, then we are blocked on incorporating that release into downstream projects until they are also updated and tested to ensure that the new cass-operator version doesn't break them. As a result, I tend to think we should prep and test the downstream changes before we merge this.

On the other hand, changes to cass-operator can always break downstream repos and that's why we do integration tests in those downstreams before release. So I'm not necessarily going to block this PR on that basis, if you really want to merge now. Just saying that we may need to do quite a lot of work in the downstreams before we can then upgrade cass-operator.

> If you wish to test that, the simplest way is to deploy cass-operator with Helm and do an upgrade. For example:

This is very helpful, thank you! That'll help me get this tested much faster.

Miles-Garnsey (commented):
There's a doc string in the types you need to change that currently reads:
```

**Review by Miles-Garnsey — COMMENTED** [review]
Issue: There's a doc string in the types you need to change that currently reads:
```// DatacenterName allows to override the name of the Cassandra datacenter. Kubernetes objects will be named after a sanitized version of it if set, ...
```

Suggestion: It would be good to have some webhook protection against excessively long dc and cluster names to avoid this kind of problem:

```
2024-09-12T07:45:13.712Z	ERROR	Could not create headless service	{"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "CassandraDatacenter": {"name":"correct-dc-name","namespace":"default"}, "namespace": "default", "name": "correct-dc-name", "reconcileID": "3e8bfbde-c188-408b-8244-0a74c55edb56", "namespace": "default", "datacenterName": "dcnamespecialcharacters", "clusterName": "cluster1&&OveriddenName", "error": "Service \"cluster1overiddenname-dcnamespecialcharacters-additional-seed-service\" is invalid: metadata.name: Invalid value: \"cluster1overiddenname-dcnamespecialcharacters-additional-seed-service\": must be no more than 63 characters"}
```

### Checks:

**Master Branch**

Pods get`cassandra.datastax.com/datacenter: dcNmCharacters`
Services get `cassandra.datastax.com/datacenter: dcNmCharacters`
PVCs get `cassandra.datastax.com/datacenter: dcNmCharacters`
sts gets `cassandra.datastax.com/datacenter: dcNmCharacters`
pvs aren't labelled at all.
ConfigMaps - it doesn't look like there are any.

In the status I see `datacenterName: dcNm!&^^ Characters` 

One odd thing with secrets, it looked like I ended up with two: `cl1oridename-superuser  cluster1-superuser`. One had `cassandra.datastax.com/datacenter: My_Super_Dc` and the other had `cassandra.datastax.com/datacenter: dcNmCharacters`. I think the former might not have been cleaned up. **Suggestion**: investigate if this still happens after this PR, since it is possible that the cleanup failures are related to the buggy labelling behaviour that we are fixing here.

**After upgrade**

Pods get `cassandra.datastax.com/datacenter: dcNmCharacters`
Services get `cassandra.datastax.com/datacenter: dcNmCharacters`
I'm gonna stop here, it doesn't look like things have been upgraded...

In the status I see `datacenterName: dcNm!&^^ Characters` 

Checking the logs, I get 

```
kubectl logs -n cass-operator k8ssandra-cass-operator-84ff466f58-cxlxq
Error from server (BadRequest): container "cass-operator" in pod "k8ssandra-cass-operator-84ff466f58-cxlxq" is waiting to start: trying and failing to pull image
```

I assume this is because the image isn't published anywhere? Do I need to build it myself and then push it somewhere to get this tested @burmanm ?

burmanm:
> I assume this is because the image isn't published anywhere? Do I need to build it myself and then push it somewhere to get this tested @burmanm ?

Right, it should be available in michaelburman290/cass-operator with the tag I mentioned. Your comments around doc string don't show up at all, no idea what happened here.

burmanm:
@Miles-Garnsey ping

Miles-Garnsey:
Going again... `datacenterName: DC&n_!!`

Checks:
**Master Branch**

Pods get cassandra.datastax.com/datacenter: DCn
Services get cassandra.datastax.com/datacenter: DCn
PVCs get cassandra.datastax.com/datacenter: DCn
sts gets cassandra.datastax.com/datacenter: DCn
pvs aren't labelled at all.
ConfigMaps - it doesn't look like there are any.

In the status I see datacenterName: DC&n_!!

**After upgrade**

Pods get cassandra.datastax.com/datacenter: test-dc
Services get cassandra.datastax.com/datacenter: test-dc
PVCs get cassandra.datastax.com/datacenter: test-dc
statefulset gets cassandra.datastax.com/datacenter: DCn (@burmanm is this expected?) 

Otherwise this looks about right to me.

Miles-Garnsey (approved):
@burmanm has raised another ticket to investigate the STS labels not being changed [here](https://github.com/k8ssandra/cass-operator/issues/719). I'll approve this since he wants to investigate that behaviour elsewhere.

PR #711, opened 2024-09-25
Allow update of CassandraDatacenter to be processed to StatefulSets i…

### LINKED ISSUE #583: Allow modification of configs if the first pod of rack fails
### What is missing?
Currently, if user deploys something that's broken and causes the first pod of the rack to not start, we require the user to manually set "forceUpgradeRacks" property which will then override the current configuration despite the datacenter not being ready from previous change.

This process feels like something we could automate. If we're still processing the first pod (but rest are up), we should probably allow user to fix the configuration and then we would apply it to make the cluster healthy again. In other words, try to detect when "forceUpgradeRacks" is going to be required and just simply apply it in our internal process.

Even better would be to rollback to previous setting, but we don't want to downgrade the user set CRDs by ourselves as that could mess up with tools like Argo.

### Why is this needed?
Automatic recovery is not easy if it requires modifying the CRD and then cass-operator will modify the CRD to ensure next time it's not applicable. We want to simplify the user experience by automatically detecting this issue as these incorrect settings do appear every now and then (user mistakes happen).

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-17) by [Unito](https://www.unito.io)
┆Fix Versions: 2024-10
┆Issue Number: CASS-17

burmanm:
…f the Datacenter is running in unhealthy Kubernetes configuration. These could be scheduling related or pods crashing

**What this PR does**:

Fixes #583 

**Checklist**
- [x] Changes manually tested


adejanovski (commented):
I was able to test this with a failed upgrade to a version of Cassandra that doesn't exist.
But I ran into a problem with the following scenario:
- I created a cluster with 1G for initial and max heap.
- Once it was started, I updated it with a 256M initial heap size and 128M max heap size.
- The pod never gets more than one restart, so any fix wouldn't be applied.
- I tried to use forceUpgradeRacks then, but it didn't apply the fix either:
```
  forceUpgradeRacks:
    - rack1
  managementApiAuth:
    insecure: {}
  racks:
    - name: rack1
```

Here's my full manifest:

```
kind: CassandraDatacenter
metadata:
  name: test-cluster
  namespace: cass-operator
spec:
  clusterName: development
  config:
    cassandra-yaml:
      authenticator: PasswordAuthenticator
      authorizer: CassandraAuthorizer
      num_tokens: 16
      role_manager: CassandraRoleManager
    jvm-server-options:
      initial_heap_size: 1G
      max_heap_size: 1G
  configBuilderResources: {}
  managementApiAuth:
    insecure: {}
  racks:
    - name: rack1
  resources:
    requests:
      cpu: 1500m
      memory: 2Gi
  serverType: cassandra
  serverVersion: 4.1.5
  size: 3
  storageConfig:
    cassandraDataVolumeClaimSpec:
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
      storageClassName: standard
```

adejanovski (approved):

PR #733, opened 2024-11-18
Make parallel restart of already bootstrapped nodes the default - CASS-79

### LINKED ISSUE #731: Concurrent restarts of previously bootstrapped nodes should be the default
### What is missing?
We've recently introduced the ability to allow concurrent startup of previously bootstrapped nodes as a feature flag.
This prevents cluster downtime in case one pods fails to restart (because of corrupt sstables for example) and blocks other pods from restarting if they need to.

We'd need to make this the default now and remove the annotation that was used as a feature flag.

### Why is this needed?
This feature protects the availability of clusters in case of incidents.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-79) by [Unito](https://www.unito.io)
┆Reviewer: Miles Garnsey
┆Fix Versions: 2024-12
┆Issue Number: CASS-79

burmanm:
**What this PR does**:
The previous fastpath is now made the default behavior. User can disable the fastpath by setting `cassandra.datastax.com/allow-parallel-starts: "false"``

Fixes #731 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -672,6 +672,10 @@ func (rc *ReconciliationContext) checkSeedLabels() (int, error) {
 	return seedCount, nil
 }
 
+func shouldUseFastPath(dc *api.CassandraDatacenter, seedCount int) bool {
+	return seedCount > 0 && !(metav1.HasAnnotation(dc.ObjectMeta, api.AllowParallelStartsAnnotations) && dc.Annotations[api.AllowParallelStartsAnnotations] == "false")
+}
+
 // CheckPodsReady loops over all the server pods and starts them
 func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMetadataEndpoints) result.ReconcileResult {
 	rc.ReqLogger.Info("reconcile_racks::CheckPodsReady")
@@ -719,7 +723,7 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
 	}
 
 	// step 0 - fastpath
-	if seedCount > 0 && metav1.HasAnnotation(rc.Datacenter.ObjectMeta, api.AllowParallelStartsAnnotations) && rc.Datacenter.Annotations[api.AllowParallelStartsAnnotations] == "true" {
+	if shouldUseFastPath(rc.Datacenter, seedCount) {
 		notReadyPods, err := rc.startBootstrappedNodes(endpointData)
 		if err != nil {
 			return result.Error(err)
--- a/pkg/reconciliation/reconcile_racks_test.go
+++ b/pkg/reconciliation/reconcile_racks_test.go
@@ -210,19 +210,16 @@ func TestReconcileRacks_ReconcilePods(t *testing.T) {
 	rc, _, cleanupMockScr := setupTest()
 	defer cleanupMockScr()
 
-	var (
-		one = int32(1)
-	)
-
+	rc.Datacenter.Spec.Size = 1
 	desiredStatefulSet, err := newStatefulSetForCassandraDatacenter(
 		nil,
 		"default",
 		rc.Datacenter,
-		2)
+		1)
 	assert.NoErrorf(t, err, "error occurred creating statefulset")
 
-	desiredStatefulSet.Spec.Replicas = &one
-	desiredStatefulSet.Status.ReadyReplicas = one
+	desiredStatefulSet.Spec.Replicas = ptr.To[int32](1)
+	desiredStatefulSet.Status.ReadyReplicas = 1
 
 	trackObjects := []runtime.Object{
 		desiredStatefulSet,
@@ -238,8 +235,8 @@ func TestReconcileRacks_ReconcilePods(t *testing.T) {
 	rc.Client = fake.NewClientBuilder().WithStatusSubresource(rc.Datacenter).WithRuntimeObjects(trackObjects...).Build()
 
 	nextRack := &RackInformation{}
-	nextRack.RackName = "default"
-	nextRack.NodeCount = 1
+	nextRack.RackName = desiredStatefulSet.Labels[api.RackLabel]
+	nextRack.NodeCount = int(*desiredStatefulSet.Spec.Replicas)
 	nextRack.SeedCount = 1
 
 	rackInfo := []*RackInformation{nextRack}
@@ -1175,9 +1172,12 @@ func mockReadyPodsForStatefulSet(sts *appsv1.StatefulSet, cluster, dc string) []
 		pod.Labels[api.ClusterLabel] = cluster
 		pod.Labels[api.DatacenterLabel] = dc
 		pod.Labels[api.CassNodeState] = "Started"
+		pod.Labels[api.RackLabel] = sts.Labels[api.RackLabel]
 		pod.Status.ContainerStatuses = []corev1.ContainerStatus{{
+			Name:  "cassandra",
 			Ready: true,
 		}}
+		pod.Status.PodIP = fmt.Sprintf("192.168.1.%d", i)
 		pods = append(pods, pod)
 	}
 	return pods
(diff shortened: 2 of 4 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -672,6 +672,10 @@ func (rc *ReconciliationContext) checkSeedLabels() (int, error) {
 	return seedCount, nil
 }
 
+func shouldUseFastPath(dc *api.CassandraDatacenter, seedCount int) bool {
+	return seedCount > 0 && !(metav1.HasAnnotation(dc.ObjectMeta, api.AllowParallelStartsAnnotations) && dc.Annotations[api.AllowParallelStartsAnnotations] == "false")
```
Miles-Garnsey, line 676: Question: Would it be possible to still have concurrent restarts for nodes which aren't seed nodes?

Issue: Can we modify the changelog to reflect that concurrent starts only applies if there is more than one seed (if not).

Miles-Garnsey: Suggestion: I had to think about this logic for a few minutes. It might be clearer to read if you remove the parenthese, distribute the negation and split it over multiple lines

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -1175,9 +1172,12 @@ func mockReadyPodsForStatefulSet(sts *appsv1.StatefulSet, cluster, dc string) []
 		pod.Labels[api.ClusterLabel] = cluster
 		pod.Labels[api.DatacenterLabel] = dc
 		pod.Labels[api.CassNodeState] = "Started"
```
Miles-Garnsey, line 1174: Suggestion: for clarity, it might be good to have a binary flag `boostrapped` or similar, this node state thing (which I think is what you're using to determine whether the node has bootstrapped) is distributed through the codebase in a pretty abstruse fashion - the label is defined [here](https://github.com/k8ssandra/cass-operator/blob/master/apis/cassandra/v1beta1/cassandradatacenter_types.go#L80) in the API, while the possible values are defined [here](https://github.com/k8ssandra/cass-operator/blob/master/pkg/reconciliation/reconcile_racks.go#L44) in the controller.

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -2952,3 +2952,99 @@ func TestCheckRackLabels(t *testing.T) {
...
+		metav1.SetMetaDataLabel(&mp.ObjectMeta, api.SeedNodeLabel, "true")
+		trackObjects = append(trackObjects, mp)
+	}
+
+	rc.Client = fake.NewClientBuilder().WithStatusSubresource(rc.Datacenter).WithRuntimeObjects(trackObjects...).Build()
+
+	nextRack := &RackInformation{}
+	nextRack.RackName = desiredStatefulSet.Labels[api.RackLabel]
+	nextRack.NodeCount = int(*desiredStatefulSet.Spec.Replicas)
+	nextRack.SeedCount = 1
```
Miles-Garnsey, line 2987: Issue: I might be misreading this but it looks like there's a mismatch between the fact that you've labelled all nodes as seeds on line 2978 while here you have a seedCount of 1. Should these line up?

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -2952,3 +2952,99 @@ func TestCheckRackLabels(t *testing.T) {
...
+
+	for i := 0; i < int(*desiredStatefulSet.Spec.Replicas); i++ {
+		ep := httphelper.EndpointState{
+			RpcAddress: fmt.Sprintf("192.168.1.%d", i+1),
+			Status:     "UN",
+		}
+		epData.Entity = append(epData.Entity, ep)
+	}
+
+	res := &http.Response{
```
Miles-Garnsey, line 3010: Suggestion: would it be possible to separate out the code which fakes the reconciliation from the code which asserts the intended state? I'm struggling to figure out what the intention of some of this stuff is to be honest.

Miles-Garnsey: Considering I'm assuming that the reconciliation-faking code is common across all of these tests, it might even be nice to run it in a separate controller in a goroutine, so that we can factor it out of all of these tests in the longer term.

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -2952,3 +2952,99 @@ func TestCheckRackLabels(t *testing.T) {
 	require.Equal(result.Continue(), res, "Label updates should not cause errors")
 	require.Subset(desiredStatefulSet.Labels, rc.Datacenter.GetRackLabels("default"))
 }
+
+func TestCheckPodsReadyAllStarted(t *testing.T) {
```
Miles-Garnsey, line 2956: Issue: I'm not seeing anything here that actually tests that multiple pods will be restarting at the same time. Is it possible to add this check? I can't manually test for this either (having tried a few times I can't catch it in the act of restarting multiple pods concurrently) so it seems that an envtest will be the only way to confirm that this PR is correct.

Miles-Garnsey (changes requested):
I've left a few comments here, I'd love to see a test that shows the pods restarting concurrently.

pkg/reconciliation/reconcile_racks_test.go
burmanm: This is not about bootstrapping (Started does not mean that). Bootstrapped nodes are listed in the Status nodeStatuses.

pkg/reconciliation/reconcile_racks_test.go
burmanm: They do not matter for this test as long as seedNodeCount > 0.

pkg/reconciliation/reconcile_racks_test.go
burmanm: For separate controller test we have envtests, they wouldn't use fake client. And no, the states are not common between all tests (as these only test a single check inside the controller flow). Think of these as unit tests.

burmanm: So each test has its own code with only the parts set that are actually required in the method that's being tested. Any other extra should be removed to  catch the modification of the implementation.

pkg/reconciliation/reconcile_racks_test.go
burmanm: That is tested in TestStartBootstrappedNodes. This PR isn't about the implementation or functionality of starting bootstrapped nodes, but only about reversing the logic from behind feature flag -> default. I simply added more tests to improve the coverage of scenarios.

pkg/reconciliation/reconcile_racks_test.go
Miles-Garnsey: So why are we setting all nodes as seeds in this case?

pkg/reconciliation/reconcile_racks_test.go
Miles-Garnsey: That's true, apologies, I think I lost sight of the objective here by asking for this. Consider it resolved.

pkg/reconciliation/reconcile_racks_test.go
Miles-Garnsey: OK, I sort of understand your point here. I'm unclear on what this is testing though - `TestCheckPodsReadyAllStarted` isn't quite describing what additional scenario you're testing here. Can you explain what additional testing you're doing and why this PR needs it?

pkg/reconciliation/reconcile_racks.go
burmanm: This has nothing to do with being a seed node or not, we bootstrap all the nodes. That count is the amount of available seed nodes in the cluster. The seed node is not what user sets or controls. If there are 0 seed nodes, that means there are 0 nodes up or 0 bootstrapped nodes. We will never start multiple nodes that need bootstrapping at the same time so having 0 seed nodes would not really be useful scenario. 

The only case this can happen is that the cluster was Stopped and then resumed. In that case, we need to first start one node -> which is then the seed and then start the rest after that one has started.

Removing the parenthese would change the logic and make it at least for me more difficult to read. Now it's two part logic (either we fail in the first one or next ones are true), removing parenthese would make it 3 parts to keep in mind. 

Especially since the last two have to be tied together with && or we get a nil pointer (we can't evaluate the value if the annotations are not set). 

I'm not entirely sure even what is your suggestion for the alternative logic. Can you code it?

pkg/reconciliation/reconcile_racks_test.go
burmanm: Because I didn't bother to set if idx == 0 { .. } there as it didn't make any difference.

pkg/reconciliation/reconcile_racks_test.go
burmanm: There was no existing test for CheckPodsReady() at all in these tests. So I wanted a test that ensures the behavior does not change in the normal scenario where everything is already up (not by this or any other PR by accident). 

It's a safeguard and documentation what's required for that method to work. If you start requiring some additional data in that method (such as by adding new function call in that method, which itself is already quite horrible code) then the test will fail. 

It does not test any new behavior, there simply wasn't anything to test its current requirements. This PR does not need it for other than ensuring I didn't break it when I reversed the logic.

Miles-Garnsey: OK, sounds fine then.

pkg/reconciliation/reconcile_racks_test.go
Miles-Garnsey: OK, I think this clutters the intention of the test, but if this is how you want it...

pkg/reconciliation/reconcile_racks.go
Miles-Garnsey: I'm suggesting:

```
return seedCount > 0 && 
  !metav1.HasAnnotation(dc.ObjectMeta, api.AllowParallelStartsAnnotations) &&
  !dc.Annotations[api.AllowParallelStartsAnnotations] == "false"
```

Miles-Garnsey (approved):
I've left one suggestion in place, please implement it if you think it is worthwhile. If not I'm approving in any event.

pkg/reconciliation/reconcile_racks.go
burmanm: That code wouldn't even compile. What you probably meant was:

```
return seedCount > 0 && 
  !metav1.HasAnnotation(dc.ObjectMeta, api.AllowParallelStartsAnnotations) &&
  dc.Annotations[api.AllowParallelStartsAnnotations] != "false"
```

And that changes the logic. Not just because dc.Annotations could be nil in this case (as HasAnnotation checks it's not nil or it would return false, yet now we get false false -> true), but also because the check is different. The test even revealed this.

If dc.Annotations[api.AllowParallelStartsAnnotations] == "true", then we should use the fastpath. The code you proposed returns false in this case, since the annotation has been set and in that case the proposed code would return false. And it's allowed to be set, we should only reject the fastpath if it's set to false, no other case.

PR #734, opened 2024-11-18
Always use dc.ObjectMeta.Name-pdb for the PodDisruptionBudget

### LINKED ISSUE #720: PodDisruptionBudget are not named according to overrides
### What happened?
When using datacenterName overrides, the dc name label is set to the overridden one, and the pdb object is created with this name as well.
If two cassdc are created with a `dc1` overridden name in the same namespace, they will end up competing for the ownership of the same `dc1-pdb` PodDisruptionBudget object, and reconcile in a loop.

The pdb objects should be created with the cassdc `meta.name` as prefix for their name, which will then avoid creating conflicts as two cassdc objects cannot have the same name in the same namespace.

Create a first cassdc using dc1 as datacenterName:

```
kind: CassandraDatacenter
metadata:
  name: cluster1-dc1
  namespace: test
spec:
  clusterName: cluster1
  datacenterName: dc1
  ...
```

and another one with the same datacenterName value:

```
kind: CassandraDatacenter
metadata:
  name: cluster2-dc1
  namespace: test
spec:
  clusterName: cluster2
  datacenterName: dc1
  ...
```

They'll both end up creating a `dc1-pdb` object.

### cass-operator version
v1.22.4

### Kubernetes version
v1.30

### Method of installation
Helm

_No response_

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-75) by [Unito](https://www.unito.io)
┆Fix Versions: 2024-11
┆Issue Number: CASS-75

burmanm:
**What this PR does**:
Uses the ObjectMeta.Name of the CassandraDatacenter for the pdb name prefix instead of the Spec.DatacenterName

Fixes #720 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/constructor.go
+++ b/pkg/reconciliation/constructor.go
@@ -20,7 +20,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// Create a PodDisruptionBudget object for the Datacenter
+// newPodDisruptionBudgetForDatacenter creates a PodDisruptionBudget object for the Datacenter
 func newPodDisruptionBudgetForDatacenter(dc *api.CassandraDatacenter) *policyv1.PodDisruptionBudget {
 	minAvailable := intstr.FromInt(int(dc.Spec.Size - 1))
 	labels := dc.GetDatacenterLabels()
@@ -31,7 +31,7 @@ func newPodDisruptionBudgetForDatacenter(dc *api.CassandraDatacenter) *policyv1.
 
 	pdb := &policyv1.PodDisruptionBudget{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        dc.LabelResourceName() + "-pdb",
+			Name:        dc.Name + "-pdb",
 			Namespace:   dc.Namespace,
 			Labels:      labels,
 			Annotations: anns,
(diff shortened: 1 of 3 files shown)


pkg/reconciliation/constructor.go
```diff
@@ -31,7 +31,7 @@ func newPodDisruptionBudgetForDatacenter(dc *api.CassandraDatacenter) *policyv1.
 
 	pdb := &policyv1.PodDisruptionBudget{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        dc.LabelResourceName() + "-pdb",
+			Name:        dc.Name + "-pdb",
```
frivoire, line 34: Why use a `-pdb` suffix here ?
It seems useless to me, since K8S resource name only needs to be unique **for a given kind** of resource.

So, I would suggest to:
```suggestion
			Name:        dc.Name,
```

frivoire (commented):
Hello 👋 
I'm pretty new in K8ssandra, but I would like to discuss one thing:

adejanovski (approved):

pkg/reconciliation/constructor.go
burmanm: You're right that it would be enough without the pdb suffix, but in this case the name was picked years ago. For most clusters created with cass-operator, this change will make no difference and the backwards compatibility is retained without any extra objects.

PR #738, opened 2024-12-12
Verify PVCs are not used by any Pods before deleting them

### LINKED ISSUE #737: Add additional checks before PVC deletion
### What is missing?
While Kubernetes has pvc-protection that checks if PersistentVolumeClaim is still used by a pod, we could do the same before issuing pvc deletion.

### Why is this needed?
In some cases (such as migrations / error cases / restore failures), users might have had orphaned pods + PVCs and the generated CassandraDatacenter wasn't correct. Deleting this CassandraDatacenter might have deleted the PVCs also as their name pair (clusterName + datacenterName) will match the PVCs, even if the Pods themselves are not actually part of the datacenter.

Thus, lets prevent this from happening and prevent the PVC deletion if a PVC is actually still in use.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-82) by [Unito](https://www.unito.io)
┆Reviewer: Radovan Zvoncek
┆Issue Number: CASS-82

burmanm:
…a PVC, verify that no Pods match this index search

**What this PR does**:
Verify that PersistentVolumeClaim is not used by any Pod before deleting it, in case there were some ghost pods still existing with the PVC mounted.

Fixes #737 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_datacenter.go
+++ b/pkg/reconciliation/reconcile_datacenter.go
@@ -77,6 +77,30 @@ func (rc *ReconciliationContext) ProcessDeletion() result.ReconcileResult {
 			// How could we have pods if we've decommissioned everything?
 			return result.RequeueSoon(5)
 		}
+	} else {
+		// This is small mini reconcile to make everything 0 sized before we finish deletion, but do not run decommission in Cassandra
+		rc.ReqLogger.Info("Proceeding with deletion, setting all StatefulSets to 0 replicas")
+		if err := rc.CalculateRackInformation(); err != nil {
+			return result.Error(err)
+		}
+
+		if res := rc.CheckRackCreation(); res.Completed() {
+			return res
+		}
+
+		waitingForRackScale := false
+		for _, sts := range rc.statefulSets {
+			currentReplicas := int(*sts.Spec.Replicas)
+			if currentReplicas > 0 {
+				waitingForRackScale = true
+				if err := rc.UpdateRackNodeCount(sts, 0); err != nil {
+					return result.Error(err)
+				}
+			}
+		}
+		if waitingForRackScale {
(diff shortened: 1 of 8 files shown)


pkg/reconciliation/reconcile_datacenter.go
```diff
@@ -125,6 +150,14 @@ func (rc *ReconciliationContext) deletePVCs() error {
 		"numPVCs", len(persistentVolumeClaimList))
 
 	for _, pvc := range persistentVolumeClaimList {
+
+		if isBeingUsed, err := rc.isBeingUsed(pvc); err != nil {
+			logger.Error(err, "Failed to check if PVC is being used")
+			return err
+		} else if isBeingUsed {
+			return fmt.Errorf("PersistentVolumeClaim %s is still being used by a pod", pvc.Name)
```
rzvoncek, line 158: Question: Upon ordinarily deleting a k8ssandra cluster, I've seen a bunch of ERROR lines caused by this in the operator logs. I understand it's a normal pattern to error out until a condition (in this case pod existence) changes. But the "ERROR	Reconciler error" seems a bit harsh. Perhaps we can do something to soften this?

rzvoncek (approved):
I cound not easily reproduce a failure scenario (bad migration or restore), so I just manually tested deleting clusters. I have a small comment on the secerity of the log lines, otherwise lgtm.

pkg/reconciliation/reconcile_datacenter.go
burmanm: That logging comes from the controller-runtime, so there's not much we can do about it. A warn in the case of reconciler error which retry might fix makes sense, but I'm not aware of having that ability. 

In the end, it's still an error. It's not like we would log this at "FATAL".

PR #740, opened 2024-12-19
Updating controller-runtime and dependencies

### LINKED ISSUE #527: Migrate to Kustomize 5
### What is missing?
A lot of templates are using features that are deprecated in Kustomize 5.0 and this causes unnecessary output from Kustomize. Since all of these features can be migrated to 4.5.7+5.0.x supported format, we should do that.

### Why is this needed?
To ensure future compatibility with Kustomize.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-23) by [Unito](https://www.unito.io)
┆Issue Number: CASS-23

### LINKED ISSUE #617: Migrate from deprecated methods and to new controllers..
### What is missing?
controller-runtime has been updated quite a lot since our last update and has following changes.

* admission.Validator is deprecated and replaced with CustomValidator. We need to move to that.
* ObjectReconciler is created, thus we can reduce the amount of boilerplate from our reconciler
* ~~Verify if some bugged tests can be re-enabled~~
* ~~"k8s.io/utils/pointer" is deprecated, use k8s.io/utils/ptr~~
* /metrics can be secured with TLS
* "sigs.k8s.io/controller-runtime/pkg/config/v1alpha1" is deprecated, remove and replace its usage
* We can allow pprof for easier debugging/profiling of cass-operator, check Manager.Options.PprofBindAddress
* ~~Use reconcile.TerminalError(...) instead of just bailing out without error~~

### Why is this needed?
controller-runtime is quite aggressive in removing deprecated methods and as such we need to keep up with the changes. Also, some of these are making our life easier.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-11) by [Unito](https://www.unito.io)
┆Issue Number: CASS-11

### LINKED ISSUE #718: Update to latest k8s.io dependencies
### What is missing?
cass-operator is currently pinned to the go dependencies for k8s 1.29.

There is a breaking change in a newer k8s.io/controller-runtime version that prevents us from upgrading our own repo to >= k8s 1.30 because of the diamond dependency problem.

k8s 1.31 was released in August and 1.32 is due in December. It would be great to upgrade cass-operator's k8s.io dependencies to 1.31.

The breaking change that will require additional changes to cass-operator is in [controller-runtime v0.18.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.18.0). The [pkg/config package is removed](https://github.com/kubernetes-sigs/controller-runtime/pull/2648/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5) and users are asked to define their own configuration. This package is [currently referenced in apis/config/v1beta1](https://github.com/k8ssandra/cass-operator/blob/3b43dfe0c394ef7ee1a2c3b7d7fdf3157c2afcf7/apis/config/v1beta1/operatorconfig_types.go#L21), so a new config implementation will need to be added to replace this.

### Why is this needed?
Updating to latest k8s.io dependencies allows users to use newer k8s.io functionality.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-73) by [Unito](https://www.unito.io)
┆Issue Number: CASS-73

burmanm:
**What this PR does**:

Fixes #718 
Fixes #527 
Fixes #617 

**Checklist**
- [ ] Changes manually tested


config/default/manager_metrics_patch.yaml
```diff
@@ -0,0 +1,4 @@
+# This patch adds the args to allow exposing the metrics endpoint using HTTPS
+- op: add
+  path: /spec/template/spec/containers/0/args/0
+  value: --metrics-bind-address=:8443
```
adejanovski, line 4: **Question:** I assume we'll need to do this in the helm charts as well?

internal/webhooks/cassandra/v1beta1/webhook_suite_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021.
+Copyright 2024 Just Me.
```
adejanovski: 🙂

internal/webhooks/cassandra/v1beta1/webhook_suite_test.go
```diff
@@ -61,69 +71,80 @@ var _ = BeforeSuite(func() {
...
 		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "config", "crd", "bases")},
 		ErrorIfCRDPathMissing: false,
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",
+			fmt.Sprintf("1.31.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
adejanovski, line 68: **Suggestion:** could we make that version a variable? I'm worried that this is quite hidden in the code.

config/webhook/manifests.yaml
```diff
@@ -1,12 +1,37 @@
 ---
 apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
```
adejanovski, line 3: **Question:** why is a mutating webhook introduced?

adejanovski (approved):

config/default/manager_metrics_patch.yaml
burmanm: Yes, the Helm chart will require changes as well.

internal/webhooks/cassandra/v1beta1/webhook_suite_test.go
burmanm: Well, I can put it to the start of the file, but this isn't used for normal envtest runs - only if running from vscode directly.

config/webhook/manifests.yaml
burmanm: Well, actually we always had the feature, but if you tried to use it, it wouldn't work. This is for the DefaultingWebhook that was always present.

PR #752, opened 2025-02-06
CASS-89 If status.datacenterName is set, but empty, do not use it

### LINKED ISSUE #751: If datacenterName in the Status is set to "", we use that value
### What happened?
A user has .status.datacenterName with value "" which we accepted as datacenterName for resources, since we only check in the processing that it's not set at all (it's a *string, so we checked for the nil value and not empty string).

We should use the metadata if the datacenterName is empty. How it ended up being empty is (was?) probably a bug in the k8ssandra-operator.

This happens in some edge case when upgrading a cluster from older version of k8ssandra-oeprator.

### cass-operator version
1.23

### Kubernetes version
1.30

### Method of installation
Helm

Update happened from 1.12 to 1.21 of k8ssandra-operator.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-89) by [Unito](https://www.unito.io)
┆Issue Number: CASS-89

burmanm:
…eNames

**What this PR does**:
Changes datacenterName processing to not accept "" as valid value for resource names.

Fixes #751 

**Checklist**
- [ ] Changes manually tested


adejanovski (approved):

KeepItSimpleStupid:
Hi ! Would it be possible to package a release with this bugfix, please ? :)

burmanm:
Yes, I'll create one today (1.23.1)

PR #754, opened 2025-02-10
Change start sequence to match StS order, prevent failing pods to block start of others

### LINKED ISSUE #750: Cluster fails to start, when entire cluster have been shutdown, and first "seed" node is corrupted
### What happened?
During an AKS node upgrade, the node running the  `-0` pod decided not to wait for the pod to stop, leaving it corrupted. In this case the other 2 nodes in our cluster got restarted, and was unable to come back up, because they where waiting for `-0` node to start

Cluster should start anyway. I'm aware that it can be hard to detect if it is because of corruption, but one solution could be to either have a timeout, or a max number of retries, before trying to use a different node as the initial seed

Stop all node, corrupt data on `-0`, for example a commitlog file, and the restart all nodes

### cass-operator version
1.22.4

### Kubernetes version
1.30.7

### Method of installation
Helm / FluxCD

Original Discord discussion:

https://discord.com/channels/836217371453685760/836217371453685763/1336286221495959552

### Workaround
1. Manually start node using the API: `curl -XPOST http://localhost:8080/api/v0/lifecycle/start`
1. Patch pod with the following labels:

```
cassandra.datastax.com/seed-node=true
cassandra.datastax.com/node-state=Started
```

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-88) by [Unito](https://www.unito.io)
┆Issue Number: CASS-88

### LINKED ISSUE #757: Modify startup sequence to match StatefulSet controller's
### What is missing?
We currently start pods from the lowest index to highest index (0->2), which does not match the behavior of how StatefulSet controller applies changes (2->0). 

### Why is this needed?
If user needs to make a change to get cluster to healthy state, we want to start the nodes that have received latest changes first. Currently, they would get started as the last nodes thus potentially leaving us in unfortunate states (for example if the old nodes have issues).

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-90) by [Unito](https://www.unito.io)
┆Issue Number: CASS-90

burmanm:
**What this PR does**:
* Prevent single bad behaving pod from blocking the start sequence in case of cluster stop/start or if new nodes are spawned.
* Start from the highest index number first like StatefulSet controller does

Fixes #750 
Fixes #757 

**Checklist**
- [ ] Changes manually tested


adejanovski (changes requested):
Issue: This PR doesn't seem to behave as it should.
When I create a new cassdc, the operator will first start pods `-0`, then it'll move to the last pod and go backwards.
I'd expect the last pod to be the first to start and have the operator go backwards then, with pod `-0` being the last to start.

burmanm:
You're right, I actually modified only the startAllNodes, in a new cluster the first node has special process.

burmanm:
I'll fix that one also.

Miles-Garnsey:
Just as a heads up (in case its relevant to this PR) I've seen a lot of issues where cassandra pods fail to come back up after a restart task is run on a DC. Typically, the symptoms are that no cassandra logs exist, and it looks like management API is running but hasn't started the Cassandra process. cass-operator then appears to continually restart and then delete the pod but no progress is ever made. There aren't typically any signs of trouble in the management API logs. 

This issue is intermittent and doesn't appear to be associated with any particular bad config, a node replacement task fixes it, as does simply force-deleting the pod (although that often then requires a few more automated restarts to avoid issues with the IP already existing and not having run a manual replacement).

I mention this because the behaviour will cause flakes in CI when running the encryption tests (there are a lot of restarts) and has made debugging the tests a bit tricky. Not sure if this PR might help fix it, but wanted to put it on the radar while we're looking at the node start process.

burmanm:
```
➜  cass-operator git:(start_sequence) ✗ kubectl get pods
NAME                    READY   STATUS    RESTARTS   AGE
cluster2-dc2-r1-sts-0   1/2     Running   0          118s
cluster2-dc2-r1-sts-1   1/2     Running   0          118s
cluster2-dc2-r1-sts-2   2/2     Running   0          118s
➜  cass-operator git:(start_sequence) ✗
```

```
➜  cass-operator git:(start_sequence) ✗ kubectl get pods
NAME                    READY   STATUS    RESTARTS   AGE
cluster2-dc2-r1-sts-0   1/2     Running   0          2m46s
cluster2-dc2-r1-sts-1   2/2     Running   0          2m46s
cluster2-dc2-r1-sts-2   2/2     Running   0          2m46s
➜  cass-operator git:(start_sequence) ✗
```

-> 

```
➜  cass-operator git:(start_sequence) ✗ kubectl get pods
NAME                    READY   STATUS    RESTARTS   AGE
cluster2-dc2-r1-sts-0   2/2     Running   0          4m37s
cluster2-dc2-r1-sts-1   2/2     Running   0          4m37s
cluster2-dc2-r1-sts-2   2/2     Running   0          4m37s
➜  cass-operator git:(start_sequence) ✗
```

adejanovski (approved):

PR #761, opened 2025-02-24
Scaffold up scheduledTask API and controller

### LINKED ISSUE #747: Create a ScheduledCassandraTask API
With some workloads it’s necessary to run compactions or garbagecollect on a regular basis on some keyspaces/tables.

It would be nice to have an API which would be similar to the MedusaBackupSchedule API in k8ssandra-operator which could run a given CassandraTask using a cron expression.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-86) by [Unito](https://www.unito.io)
┆Issue Number: CASS-86

Miles-Garnsey:
**What this PR does**:
Creates a new ScheduledTask API, controller and reconciliation logic.

Fixes #747 

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
This is close, I think the envtests are passing but the e2e's have flaked out, probably due to permissions issues. I've got a question for Micke [here (internal)](https://datastax.slack.com/archives/C02023LQC1E/p1740468091599149) about how to land my permissions in the right role instead of generating a whole new one in codegen. I'm doing something slightly wrong there.

apis/scheduledtask.k8ssandra.io/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "scheduledtask.k8ssandra.io", Version: "v1alpha1"}
```
burmanm: Is there a reason you didn't just add this to the control.k8ssandra.io group and the same api pkg? Why do we need a new group?

cmd/main.go
```diff
@@ -222,22 +216,37 @@ func main() {
 		return pvcNames
 	}); err != nil {
 		setupLog.Error(err, "unable to set up field indexer")
-		os.Exit(1)
-	}
+		if err = (&scheduledtaskcontrollers.ScheduledTaskReconciler{
```
burmanm: issue: If err != nil, this should call os.Exit(1), now it after receiving an exit worthy error makes new reconcilers?

burmanm: I feel like there's a missing } here and that messed up the logic entirely.

Miles-Garnsey: I think you're right, this is a bit of a mess. I've just killed off my changes and tried again, maybe this was why the e2e tests were failing on me. Good catch, thank you...

apis/scheduledtask.k8ssandra.io/v1alpha1/groupversion_info.go
Miles-Garnsey: I don't see a lot of difference either way, I can move it into the existing group if you have feelings about it.

burmanm (approved):
No e2e tests, but since this code is lifted from the MedusaScheduler the implementation is quite tested. But follow-up should be including a real test (since this has no envtest or e2e tests).

PR #764, opened 2025-02-26
Use ENDPOINT_IP as replacement address

### LINKED ISSUE #763: Allow overriding the replaced address in the ReplaceNode CassandraTask
### What is missing?
In some setups, the pod IP address is not the broadcasted address for the Cassandra nodes. In this scenario we would need the ability to explicitly set the replaced address in the CassandraTask so that it can go through correctly.
This could take the form of an additional argument in the CassandraTask spec.

### Why is this needed?
In case the pod IP isn't the broadcasted address, the ReplaceNode CassandraTask cannot operate. This requires to go through a convoluted and hacky process to get things done.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-94) by [Unito](https://www.unito.io)
┆Issue Number: CASS-94

burmanm:
…acement instead of NATIVE_ADDRESS_AND_PORT

**What this PR does**:
NATIVE_ADDRESS_AND_PORT is currently used as the default replacement IP when replacing nodes. The correct however looks to be the ENDPOINT_IP in certain environment, so we should use that instead the default.

Fixes #763 

**Checklist**
- [ ] Changes manually tested


adejanovski (approved):
Successfully tested, thanks!

PR #770, opened 2025-03-20
Fix 5.1 nightly support

### LINKED ISSUE #771: Multiple racks get their update the same time during certain upgrade scenario
### What happened?

Running tests in #770 has gotten me to situation where multiple racks go down at the same time. This should not happen, we should wait for the previous update to succeed before applying the next one.

This only happens if the upgrade is touching the StatefulSet's target labels.

_No response_

770,

### cass-operator version

1.24-dev

### Kubernetes version

1.32

### Method of installation

_No response_

_No response_

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-98) by [Unito](https://www.unito.io)
┆Issue Number: CASS-98

burmanm:
**What this PR does**:
Adds support for running 5.1 nightly images.

* Updates Vector to 0.45.0
* Updates k8ssandra-client to 0.6.2 to provide 5.1 support
* Adds smoke test against 5.1-nightly image.

Fixes #771 

**Checklist**
- [ ] Changes manually tested


.github/workflows/kindIntegTest.yml
```diff
@@ -231,23 +231,26 @@ jobs:
...
-            serverImage: datastax/dse-mgmtapi-6_8:6.8.53-ubi8 # DSE 6.8.53
+          - version: 6.8.54
+            serverImage: datastax/dse-mgmtapi-6_8:6.8.54-ubi8 # DSE 6.8.54
             serverType: dse
-          - version: 6.9.6
-            serverImage: datastax/dse-mgmtapi-6_8:6.9.6-ubi # DSE 6.9.6
+          - version: 6.9.7
+            serverImage: datastax/dse-mgmtapi-6_8:6.9.7-ubi # DSE 6.9.7
             serverType: dse
+          - version: 5.1.0
```
emerkle826: comment:
Hopefully trunk doesn't become 6.0 soon, but we'll deal with that if/when it happens

emerkle826 (approved):
This looks good to me

emerkle826:
The only thing that concerns me from the tests is the 5.1.0 failing one here:
https://github.com/k8ssandra/cass-operator/actions/runs/13989112394/job/39169014237?pr=770
If you can get the server logs when that test is running, it will shed some light on potential issues in the image. If there are issues in the image, let me know and I can get them fixed ASAP

.github/workflows/kindIntegTest.yml
burmanm: It shouldn't affect the test itself even if the version changes as there would be no difference in processing 5.1.0 vs 6.0. The version just needs to be hardcoded to something here.

burmanm:
No, there's definitely something going on, but I'm not sure if it's the mgmt-api changes or something else. The pods fail to report their logs entirely, so I'm wondering if there's some resource issue right now.

PR #809, opened 2025-07-01
If IsLocal is set, use that information for the HostID setting

### LINKED ISSUE #814: Modify HostID parsing to use IS_LOCAL field in EndpointStates
### What is missing?

The mgmt-api returns "IS_LOCAL" to indicate which gossipinfo in the endpointstates is for this Pod that was called. We should use that information to fill the HostID using that instead of trying to search for matching IP.

This is supported in mgmt-api v0.1.69 and newer (from 2023).

### Why is this needed?

If the IP is different (such as overridden broadcast_address) due to a CNI configuration, we won't be able to match the PodIP with the native_port.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-117) by [Unito](https://www.unito.io)
┆Issue Number: CASS-117

burmanm:
**What this PR does**:
mgmt-api v0.1.69 and newer support checking EndpointStates with IS_LOCAL. We could use that information for the HostID instead of the IP matching that's required with older versions.

This change is required for certain multi-CNI or otherwise modified broadcast_addresses.

Fixes #814 

**Checklist**
- [ ] Changes manually tested


adejanovski (approved):
Looks good

PR #823, opened 2025-07-18
Remove certain RBAC requirements

### LINKED ISSUE #822: Reduce RBAC requirements
### What happened?

Due to historical reasons, some of the RBAC requirements in cass-operator are outdated and are no longer necessary. We should review these to ensure our serviceAccounts only have what we need, nothing extra.

_No response_

-

### cass-operator version

1.26.0

### Kubernetes version

1.33

### Method of installation

_No response_

_No response_

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-119) by [Unito](https://www.unito.io)
┆Issue Number: CASS-119

burmanm:
**What this PR does**:
Updates the RBAC rules to remove certain rights that are no longer necessary for cass-operator.

What's missing however is proper envtests using RBAC restrictions so we could easily test if all operations are really working. We will rely on e2e tests for now.

Fixes #822 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -55,9 +55,6 @@ rules:
 - apiGroups:
   - apps
   resources:
-  - daemonsets
-  - deployments
-  - replicasets
   - statefulsets
   verbs:
   - create
@@ -67,12 +64,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - apps
-  resources:
-  - deployments/finalizers
-  verbs:
-  - update
 - apiGroups:
   - cassandra.datastax.com
   resources:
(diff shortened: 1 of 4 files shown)


config/rbac/role.yaml
```diff
@@ -30,7 +30,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - configmaps
```
adejanovski: **Issue:** isn't that necessary for the system logger to read the vector configuration cm ?

burmanm: To this, no it isn't since it's mounted (so our operator does not need the rights), but it's now (from another PR) required by the ImageConfig.

PR #826, opened 2025-08-15
Refactor the handling of ImageConfig to a stateful imageRegistry

### LINKED ISSUE #830: Add new ImageConfig with structured settings
### What is missing?

Our current ImageConfig uses a set of strings to define a version with ability to override blocks only globally. Some users find this difficult to use in their environment and some have settings that require changing the string representation, thus making it difficult to update without manually also updating the strings.

We could use a structured YAML with separated parts for repository, registry, image-name and tag.

### Why is this needed?

User requirements.

burmanm:
**What this PR does**:
Improves the code quality by removing direct global calls to the images pkg and instead allows implementation of different ImageRegistry concepts. 

Fixes #830

**Checklist**
- [ ] Changes manually tested


apis/config/v1beta2/imageconfig_types.go
```diff
@@ -0,0 +1,149 @@
...
+	// The Docker repository to use.
+	// +optional
+	Repository string `json:"repository,omitempty"`
+
+	// The image name to use.
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// The image tag to use. Defaults to "latest".
+	// +kubebuilder:default="latest"
```
adejanovski, line 35: **Question:** should we keep using `latest` as a default? This created some interesting issues in the past.

apis/config/v1beta2/imageconfig_types.go
```diff
@@ -0,0 +1,149 @@
...
+	// +optional
+	// +kubebuilder:validation:Enum:=Always;IfNotPresent;Never
+	PullPolicy corev1.PullPolicy `json:"pullPolicy,omitempty"`
+
+	// The secret to use when pulling the image from private repositories. If specified, this secret will be passed to
+	// individual puller implementations for them to use. For example, in the case of Docker, only DockerConfig type
+	// secrets are honored. More info:
+	// https://kubernetes.io/docs/concepts/containers/images#specifying-imagepullsecrets-on-a-pod
+	// +optional
+	PullSecret string `json:"pullSecret,omitempty"`
```
adejanovski, line 49: **Question:** no need to provide a list of secrets instead?

apis/config/v1beta2/imageconfig_types.go
burmanm: I lifted this from k8ssandra-operator, but yeah, I should definitely remove the default value entirely.

apis/config/v1beta2/imageconfig_types.go
burmanm: You can't have multiple secrets targeting one container pull (this is per-container setting, not the list of secrets definition).

PR #832, opened 2025-09-01
Add jobs parameter to garbagecollect task

### LINKED ISSUE #831: make the K8ssandraTask garbagecollect command support no. of jobs
### What is missing?

LS,

First of all thanks for your amazing application! i think it is a high quality kubernetes operator, it puts many other operators to shame.

Currently the garbagecollect command doesn't support configuring the number of jobs. It would be to add that, to speed up the freeing of disk space in case of disk pressure on the nodes.

I think implementing is as changing two functions, since the C* management api already supports it:
```
// line 1420 in pkg/httphelper/client.go
func (client *NodeMgmtClient) CallGarbageCollect(pod *corev1.Pod, jobs int, keyspaceName string, tables []string) (string, error) {
	client.Log.Info(
		"calling Management API keyspace flush - POST /api/v1/ops/tables/garbagecollect",
		"pod", pod.Name,
	)

	req, err := createKeySpaceRequest(pod, jobs, keyspaceName, tables, "/api/v1/ops/tables/garbagecollect")
	if err != nil {
		return "", err
	}
```
```
// line 328 in internal/controllers/control/jobs.go
func callGarbageCollectAsync(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) (string, error) {
	keyspaceName := taskConfig.Arguments.KeyspaceName
	jobs := taskConfig.Arguments.JobsCount
	tables := taskConfig.Arguments.Tables
	return nodeMgmtClient.CallGarbageCollect(pod, jobs, keyspaceName, tables)
```
I wanted to create a branch to commit these changes myself, but it seems I am missing permissions, or I am too clueless (or both).

cheers,
Hidde

### Why is this needed?

Currently the garbagecollect command doesn't support configuring the number of jobs. It would be to add that, to speed up the freeing of disk space in case of disk pressure on the nodes.

burmanm:
**What this PR does**:
Passes jobs argument to the garbagecollect calls

Fixes #831 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/internal/controllers/control/jobs.go
+++ b/internal/controllers/control/jobs.go
@@ -328,13 +328,21 @@ func flush(taskConfig *TaskConfiguration) {
 func callGarbageCollectAsync(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) (string, error) {
 	keyspaceName := taskConfig.Arguments.KeyspaceName
 	tables := taskConfig.Arguments.Tables
-	return nodeMgmtClient.CallGarbageCollect(pod, keyspaceName, tables)
+	jobCount := -1
+	if taskConfig.Arguments.JobsCount != nil {
+		jobCount = *taskConfig.Arguments.JobsCount
+	}
+	return nodeMgmtClient.CallGarbageCollect(pod, jobCount, keyspaceName, tables)
 }
 
 func callGarbageCollectSync(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) error {
 	keyspaceName := taskConfig.Arguments.KeyspaceName
 	tables := taskConfig.Arguments.Tables
-	return nodeMgmtClient.CallGarbageCollectEndpoint(pod, keyspaceName, tables)
+	jobCount := -1
+	if taskConfig.Arguments.JobsCount != nil {
+		jobCount = *taskConfig.Arguments.JobsCount
+	}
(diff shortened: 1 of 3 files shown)


internal/controllers/control/jobs.go
```diff
@@ -328,13 +328,21 @@ func flush(taskConfig *TaskConfiguration) {
 func callGarbageCollectAsync(nodeMgmtClient httphelper.NodeMgmtClient, pod *corev1.Pod, taskConfig *TaskConfiguration) (string, error) {
 	keyspaceName := taskConfig.Arguments.KeyspaceName
 	tables := taskConfig.Arguments.Tables
-	return nodeMgmtClient.CallGarbageCollect(pod, keyspaceName, tables)
+	jobCount := -1
```
adejanovski, line 331: **Question:** The default in nodetool garbagecollect is 2. If unset explicitly here we'll get -1: do we know what that means once garbagecollect triggers?

adejanovski (approved):

internal/controllers/control/jobs.go
barry-hubris: when i create a task like so (which makes a call with jobs = -1):
```
kind: K8ssandraTask
metadata:
  name: garbagecollect
  namespace: cassandra
spec:
  cluster:
    name: cluster
  datacenters:
    - dc01
  template:
    jobs:
      - name: garbagecollect
        command: garbagecollect
        args:
          keyspace_name: dummy
```
on this version:
cr.k8ssandra.io/k8ssandra/cass-management-api:5.0.1-ubi
  
it starts garbagecollection with a single compaction task, while the maximum concurrency in this cluster is 8.

PR #847, opened 2025-09-30
Add support for setting digest in the ImageConfig and Image types.

### LINKED ISSUE #846: Allow deploying digest only without defining a tag
### What is missing?

As per Kubernetes image (OCI) the deployment should be done with either tag, tag@digest or digest only. Our current ImageConfig does not allow setting the last one instead requiring the tag@digest to be used if a digest deployment is required. This is unnecessary as we can detect from the [OCI spec](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests) what the correct digest format is and simply use that for deployment without incorrectly adding another `:` to the deploy target.

### Why is this needed?

Some environments prefer digest instead of tags when deploying images. That should be supported.

burmanm:
**What this PR does**:
Add support for deploying with digest information instead of only tags with digest as suffix.

Fixes #846 

**Checklist**
- [ ] Changes manually tested


adejanovski (approved):

adejanovski:
LGTM @burmanm, needs a rebase before it can be merged

PR #849, opened 2025-10-02
Better PVC resize error detection, no ClusterRole mode

### LINKED ISSUE #838: Ability to run in reduced mode without any ClusterRole requirements
### What is missing?

In certain cases, the user is unable to deploy a ClusterRole that's required for certain resources such as PersistentVolumes and StorageClasses. We need to have a mode that disables all such features if the ClusterRole is not deployed.

### Why is this needed?

While these are features many users would prefer to have, in certain deployment models it's just not possible / convenient to have them. Instead of preventing the installation entirely, lets ensure we can run in a slightly degraded mode instead.

### LINKED ISSUE #848: Improve PVC resize failure detection
### What is missing?

We have relied on the StorageClass detection if resizing is available, but there's a need to detect errors in other ways also. First is running without ClusterRole rights (#838) and in that case the validation can not function. So user might deploy a resized cluster without actually being able to resize it since the underlying StorageClass does not allow it.

The other cases are when the PVC resizing fails due to reasons that are out of our hands. These could be transient or permanent (such as running out of disk / hitting maximum limit at the cloud provider etc).

### Why is this needed?

We need to support non-ClusterRole installation also and that means less validations. And we want to detect more error cases to ensure the operator gives proper feedback to the users.

burmanm:
**What this PR does**:
Add a running mode that disables the usage of features that require ClusterRoles. Also, improve the detection of PVC resizing failures since we could be missing a validation stage that requires ClusterRole to be present.

Reduce the amount of RBAC requirements for ClusterRole by replacing the PV Capacity poll by trusting the PVC's Capacity status.

Fixes #848 
Fixes #838 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -257,9 +257,17 @@ func (rc *ReconciliationContext) CheckPVCResizing() result.ReconcileResult {
 	}
 
 	for _, pvc := range pvcList {
-		if isPVCResizing(&pvc) {
-			rc.ReqLogger.Info("Waiting for PVC resize to complete",
-				"pvc", pvc.Name)
+		if pvcResizingFailed(&pvc) {
+			msg := fmt.Sprintf("PVC resize failed for pvc %s, check events for more details", pvc.Name)
+			rc.Recorder.Eventf(rc.Datacenter, corev1.EventTypeWarning, events.ResizingPVCFailed, msg)
+			if err := rc.setCondition(
+				api.NewDatacenterConditionWithReason(api.DatacenterValid, corev1.ConditionFalse, "pvcResizeFailed", msg)); err != nil {
+				return result.Error(err)
+			}
+		}
+
+		if pvcResizing(&pvc) {
+			rc.ReqLogger.Info("Waiting for PVC resize to complete", "pvc", pvc.Name)
 			return result.RequeueSoon(10)
 		}
 	}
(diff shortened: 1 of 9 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -319,21 +334,24 @@ func (rc *ReconciliationContext) CheckVolumeClaimSizes(statefulSet, desiredSts *
...
-
-			if !supportsExpansion {
-				msg := fmt.Sprintf("PVC resize requested, but StorageClass %s does not support expansion", *claim.Spec.StorageClassName)
-				rc.Recorder.Eventf(rc.Datacenter, corev1.EventTypeWarning, events.InvalidDatacenterSpec, msg)
-				if err := rc.setCondition(
-					api.NewDatacenterConditionWithReason(api.DatacenterValid,
-						corev1.ConditionFalse, "storageClassDoesNotSupportExpansion", msg,
-					)); err != nil {
+			// We can't do this check without ClusterRole for StorageClass. Some environments don't want these.
+			if rc.ClusterResources {
```
adejanovski, line 338: **Question:** what's the impact of skipping the check? Is it making the feedback loop longer or would it cause some trouble?

burmanm: Instead of validation error we'll get an error when the resize starts. So processing error vs. validation, it's not a large difference.

adejanovski (approved):

PR #886, opened 2026-02-03
Implement parallel pods in tasks

### LINKED ISSUE #861: Support parallel pod operations in CassandraTasks
### What is missing?

We'd need the ability to run CassandraTasks operations concurrently, with tunable concurrency:
- sequential (one node at a time)
- Specific number of nodes (not sure about this one, gives more flexibility but seems harder to tune/implement).

One followup task here is to allow configuring the concurrency of the automated cleanups that cass-operator starts automatically after an expansion, and same in k8ssandra-operator for the rebuilds after a DC expansion.

### Why is this needed?

There are several cases where CassandraTasks are running too slow for operational requirements.
Cleanup is one such example, and so are upgradesstables and rebuilds for example.
On large clusters (10s or 100s of nodes) running sequentially will just take too long.

### LINKED ISSUE #885: Add configurable retry count to tasks
### What is missing?

At the moment, RetryOnFailure will retry a single time and then fail the pod. We should have a setting to make this configurable. 

### Why is this needed?

Some operations can be retried multiple times if they fail without incurring much overhead (they are either idempotent or next invocations are fast as the operation was mostly completed anyway).

burmanm:
**What this PR does**:

* Implements parallel pod processing for all async tasks (sync tasks are still limited but there really shouldn't be any users left for those endpoints)
* Implement configurable retryCount
* Move from PodAnnotations to the CassandraTask Status to allow better visibility and tracking

Fixes #885 
Fixes #861 

**Checklist**
- [ ] Changes manually tested


adejanovski (approved):
tested with 4 nodes per rack across 3 racks, with some long lasting operations with a few GBs of data per node. It all worked as expected.

I also tested taking pods down while the task is running to test retries. It sometimes took a bit of time to recover, but always went through eventually.

You need a rebase as the fast rollout PR merge created conflicts it seems.

PR #889, opened 2026-02-10
default retry policy for cassandratask has changed to onfailure

### LINKED ISSUE #850: Failed cleanups aren't actionable
### What is missing?

After a scale out operation, cleanup will be triggered using a CassandraTask.
If cleanup fails on one of the database pods, it won't get retried and is only reported in the CassandraTask status.
This doesn't allow proper alerting as there's no way to convert this into a metric.
CassandraTasks get deleted quite quickly after completion as well, so there's a timing after which we cannot know if the last cleanup has succeeded.
I think it's not possible either to know which pod has failed the task, we only have counters.

The first thing we need is a metric with the number of failed pods for a CassandraTask, using the task type as a label.

We should also consider variable TTLs depending on the task to allow longer time for inspection, and providing the list of pod names that have failed the task.
A failed cleanup can have quite a lot of consequences and we should consider a retry policy as sstables which have been cleaned up already get skipped, which doesn't create overhead.

On large clusters, the sequential execution of cleanup is too slow, which delays the ability to perform another scaling operation (depending on the chosen strategy, but still). We need to have options to specify the number of compactors used by the cleanup task, and also offer the ability to run cleanup on all nodes in a rack at a time.

### Why is this needed?

The way CassandraTasks, and specifically the cleanup one, work creates some operational challenges.
We need to improve the production experience and provide proper monitoring for all these features along with more options to tune the pressure of distributed tasks.

artsiom-bukhautsou:
**What this PR does**:
- It changes default retry policy for `CassandraTask` CRD to OnFailure

Fixes #850 

**Checklist**
- [x] Changes manually tested
- [X] Automated Tests added/updated
- [X] Documentation added/updated
- [X] CHANGELOG.md updated (not required for documentation PRs)

### DIFF
--- a/internal/controllers/control/cassandratask_controller_test.go
+++ b/internal/controllers/control/cassandratask_controller_test.go
@@ -592,7 +592,9 @@ var _ = Describe("CassandraTask controller tests", func() {
 
 			It("Runs a cleanup task against the datacenter pods", func() {
 				By("Creating a task for cleanup")
-				taskKey := createTask(api.CommandCleanup, testNamespaceName)
+				taskKey, task := buildTask(api.CommandCleanup, testNamespaceName)
+				task.Spec.RestartPolicy = corev1.RestartPolicyNever
+				Expect(k8sClient.Create(context.Background(), task)).Should(Succeed())
 
 				completedTask := waitForTaskCompletion(taskKey)
 
@@ -606,7 +608,9 @@ var _ = Describe("CassandraTask controller tests", func() {
 
 			It("Runs a upgradesstables task against the datacenter pods", func() {
 				By("Creating a task for upgradesstables")
-				taskKey := createTask(api.CommandUpgradeSSTables, testNamespaceName)
+				taskKey, task := buildTask(api.CommandUpgradeSSTables, testNamespaceName)
+				task.Spec.RestartPolicy = corev1.RestartPolicyNever
+				Expect(k8sClient.Create(context.Background(), task)).Should(Succeed())
 
 				completedTask := waitForTaskCompletion(taskKey)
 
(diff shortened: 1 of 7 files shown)


internal/controllers/control/cassandratask_controller_test.go
```diff
@@ -591,7 +592,9 @@ var _ = Describe("CassandraTask controller tests", func() {
 
 			It("Runs a cleanup task against the datacenter pods", func() {
 				By("Creating a task for cleanup")
-				taskKey := createTask(api.CommandCleanup, testNamespaceName)
+				taskKey, task := buildTask(api.CommandCleanup, testNamespaceName)
```
artsiom-bukhautsou, line 594: For some reason after the rebase, those changes were required. I'm not sure how the default RetryPolicy `OnFailure` affected it.

internal/controllers/control/cassandratask_controller_test.go
```diff
@@ -484,20 +484,22 @@ var _ = Describe("CassandraTask controller tests", func() {
 					defer mockServer.Close()
 
 					By("create datacenter", createDatacenter("dc1", testFailedNamespaceName))
-					By("Create a task for cleanup")
-					taskKey := createTask(api.CommandCleanup, testFailedNamespaceName)
+					By("Creating a task for cleanup")
+					taskKey, task := buildTask(api.CommandCleanup, testFailedNamespaceName)
```
burmanm: suggestion: This can keep using createTask, it bundles the buildTask + create + Expect. The buildTask() case was used where we wanted to modify the base task before creating it.

artsiom-bukhautsou: Makes sense. Fixed

burmanm (approved):
You need to rebase to get the CHANGELOG in order. Can you also reorder it based on the rules at the top? It seems merges from different PRs have messed up the ordering.

PR #895, opened 2026-02-23
prevent infinite loop when seed refresh times out during node failures

### LINKED ISSUE #887: Infinitely terminating instances will block reconcile
### What happened?

We've seen a case where a pod has been stuck to a state of terminating without Kubernetes being unable to recover from this. This has happened with a sequence of two Kubernetes worker nodes going down (intentionally, but without proper cordon/drain) at the same time. What we see is then some interesting behavior where even replacing two nodes is not possible (with one down node this works, but not with two).

In the starting sequence, the seedRefresh will block any attempts to start new nodes as the seedRefresh happens before any fastStart or normal bootstrapping:

```
2026-01-27T15:59:27.402Z	ERROR	controllers.CassandraDatacenter	calculateReconciliationActions returned an error	{"cassandradatacenter": {"name":"test1-dc1","namespace":"p1-4e85bcg0"}, "requestNamespace": "p1-4e85bcg0", "requestName": "test1-dc1", "loopID": "0458905a-6881-49f1-9944-18da2194dbdf", "error": "Post \"https://192.168.66.30:8080/api/v0/ops/seeds/reload\": context deadline exceeded"}
```

This will then cause reconcile and the loop continues forever. This one is tricky in that technically we could have a pod with incorrect seeds still. In reality, this is probably not as big of a problem as we tend to have 3 seed nodes so even if one is in the wrong state, everything should be fine (and new bootstrapping nodes should have DNS refresh in any case).

Second issue comes that this pod will cause some weird issues in certain other parts of the code, such as:

```
2026-01-28T14:51:59.500Z	ERROR	Observed a panic	{"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "CassandraDatacenter": {"name":"test-dc1","namespace":"p2-311t5u4p"}, "namespace": "p2-311t5u4p", "name": "test-dc1", "reconcileID": "aac79841-fc1c-4bfd-bbf6-e909eea980ea", "panic": "runtime error: invalid memory address or nil pointer dereference", "panicGoValue": "\"invalid memory address or nil pointer dereference\"", "stacktrace": "goroutine 285 [running]:\nk8s.io/apimachinery/pkg/util/runtime.logPanic({0x2518b88, 0xc0015fe9c0}, {0x1e7c6c0, 0x37304d0})\n\t/go/pkg/mod/k8s.io/apimachinery@v0.33.3/pkg/util/runtime/runtime.go:132 +0xbc\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile.func1()\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:108 +0x112\npanic({0x1e7c6c0?, 0x37304d0?})\n\t/usr/local/go/src/runtime/panic.go:792 +0x132\ngithub.com/k8ssandra/cass-operator/pkg/reconciliation.isServerReady(...)\n\t/workspace/pkg/reconciliation/reconcile_racks.go:2249\ngithub.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).createStartSequence(0xc000da28c0)\n\t/workspace/pkg/reconciliation/reconcile_racks.go:2113 +0x386\ngithub.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).startAllNodes(0xc000da28c0, {{0xc000c42008?, 0x21eec25?, 0x21d04c4?}})\n\t/workspace/pkg/reconciliation/reconcile_racks.go:2088 +0x6a\ngithub.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).CheckPodsReady(0xc000da28c0, {{0xc000c42008?, 0x21cd80c?, 0x1e?}})\n\t/workspace/pkg/reconciliation/reconcile_racks.go:801 +0x6e5\ngithub.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).ReconcileAllRacks(0xc000da28c0)\n\t/workspace/pkg/reconciliation/reconcile_racks.go:2694 +0x4ed\ngithub.com/k8ssandra/cass-operator/pkg/reconciliation.(*ReconciliationContext).CalculateReconciliationActions(0xc000da28c0)\n\t/workspace/pkg/reconciliation/handler.go:70 +0xc5\ngithub.com/k8ssandra/cass-operator/internal/controllers/cassandra.(*CassandraDatacenterReconciler).Reconcile(0xc000029590, {0x2518b88, 0xc0015fe9c0}, {{{0xc0012cfb50, 0xb}, {0xc0012cfbb0, 0x8}}})\n\t/workspace/internal/controllers/cassandra/cassandradatacenter_controller.go:147 +0xa3a\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile(0xc0015fe930?, {0x2518b88?, 0xc0015fe9c0?}, {{{0xc0012cfb50?, 0x0?}, {0xc0012cfbb0?, 0x0?}}})\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:119 +0xbf\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler(0x25430c0, {0x2518bc0, 0xc0000294a0}, {{{0xc0012cfb50, 0xb}, {0xc0012cfbb0, 0x8}}}, 0x0)\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:340 +0x3ad\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem(0x25430c0, {0x2518bc0, 0xc0000294a0})\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:300 +0x21b\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.1()\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:202 +0x85\ncreated by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2 in goroutine 76\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:198 +0x28f\n"}
```

This is probably caused by the Pod missing some fields. We might need to be more defensive here, but also we would probably require something to avoid incorrect pods from blocking everything. Yet, I'm not sure how we can detect these from real terminating pods which can take a while and that's perfectly normal. 

I'd expect cass-operator able to recover from this state without large manual process. I'd say this requires some investigation into how refreshSeeds is required as a process at this point when pods have anyway bootstrapped and if we really need to call it so often or block processing otherwise. It's not necessarily true that just fixing this would solve the issue, but this is the first step for investigation.

B) Two nodes down - replacenode does not work
* 4 nodes DSE cluster
* Stop the AWS instance for TWO of the k8s worker nodes
* The DSE node pods on the down k8s worker nodes go into Terminating state
* Create a replacenode cassandratask for ONE of the down node. I was thinking to try replace one by one
* run kubectl delete pod <terminating pod with replacenode task> --force. 
* A new DSE node pod is created on a new k8s worker node. But it does not start. Its stuck in the 1/2 Running state for over an hour.
* At this point I thought maybe replacing the other node might do somthing:
  * Create a replacenode cassandratask for OTHER of the down node
  * run kubectl delete pod <terminating pod with new replacenode task> --force
  * A new DSE node pod is created on a new k8s worker node. But it does not start. Its stuck in the 1/2 Running state 
* Now both nodes (sts-1, sts-2) are in a loop of Init -> Running 1/2 (DSE never tries to start)  -> Wait ~10 mins -> Terminating

### cass-operator version

1.28

### Kubernetes version

Unknown

### Method of installation

_No response_

_No response_

artsiom-bukhautsou:
**What this PR does**:
Prevents infinite loop when seed refresh times out during node failures, allowing the operator to recover.

Fixes #887 

**Checklist**
- [X] Changes manually tested
- [X] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -704,18 +704,21 @@ func (rc *ReconciliationContext) CheckRackStoppedState() result.ReconcileResult 
 }
 
 // checkSeedLabels loops over all racks and makes sure that the proper pods are labelled as seeds.
-func (rc *ReconciliationContext) checkSeedLabels() (int, error) {
+func (rc *ReconciliationContext) checkSeedLabels() (seedCount int, didSeedsChange bool, err error) {
 	rc.ReqLogger.Info("reconcile_racks::CheckSeedLabels")
-	seedCount := 0
+	seedCount = 0
 	for idx := range rc.desiredRackInformation {
 		rackInfo := rc.desiredRackInformation[idx]
-		n, err := rc.labelSeedPods(rackInfo)
+		n, didRackSeedsChange, err := rc.labelSeedPods(rackInfo)
 		seedCount += n
 		if err != nil {
-			return 0, err
+			return 0, false, err
+		}
+		if didRackSeedsChange {
+			didSeedsChange = true
 		}
 	}
-	return seedCount, nil
+	return seedCount, didSeedsChange, nil
 }
 
 func shouldUseFastPath(dc *api.CassandraDatacenter, seedCount int) bool {
@@ -759,17 +762,15 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
 
 	// get the nodes labelled as seeds before we start any nodes
 
-	seedCount, err := rc.checkSeedLabels()
+	seedCount, didSeedsChange, err := rc.checkSeedLabels()
 	if err != nil {
 		return result.Error(err)
 	}
-	err = rc.refreshSeeds()
-	if err != nil {
-		rc.ReqLogger.Error(err,
-			"failed to refresh seeds, but continuing reconciliation",
-			"seedCount", seedCount,
-			"totalPods", len(rc.dcPods),
-		)
+	if didSeedsChange {
+		err = rc.refreshSeeds()
+		if err != nil {
+			return result.Error(err)
+		}
 	}
 
 	// step 0 - fastpath
@@ -1480,8 +1481,8 @@ func (rc *ReconciliationContext) isClusterHealthy() bool {
 
 // labelSeedPods iterates over all pods for a statefulset and makes sure the right number of
 // ready pods are labelled as seeds, so that they are picked up by the headless seed service
-// Returns the number of ready seeds.
-func (rc *ReconciliationContext) labelSeedPods(rackInfo *RackInformation) (int, error) {
+// Returns the number of ready seeds and informs if rack seeds were updated.
+func (rc *ReconciliationContext) labelSeedPods(rackInfo *RackInformation) (count int, didSeedsChange bool, err error) {
 	logger := rc.ReqLogger.WithName("labelSeedPods")
 
 	rackPods := rc.rackPods(rackInfo.RackName)
@@ -1489,7 +1490,7 @@ func (rc *ReconciliationContext) labelSeedPods(rackInfo *RackInformation) (int, 
 	sort.SliceStable(rackPods, func(i, j int) bool {
 		return rackPods[i].Name < rackPods[j].Name
 	})
-	count := 0
+	count = 0
 	for _, pod := range rackPods {
 		patch := client.MergeFrom(pod.DeepCopy())
 
@@ -1528,16 +1529,17 @@ func (rc *ReconciliationContext) labelSeedPods(rackInfo *RackInformation) (int, 
 		}
 
 		if shouldUpdate {
+			didSeedsChange = true
 			pod.SetLabels(newLabels)
 			if err := rc.Client.Patch(rc.Ctx, pod, patch); err != nil {
 				logger.Error(
 					err, "Unable to update pod with seed label",
 					"pod", pod.Name)
-				return 0, err
+				return 0, false, err
 			}
 		}
 	}
-	return count, nil
+	return count, didSeedsChange, nil
 }
 
 // GetStatefulSetForRack returns the statefulset for the rack
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -502,7 +502,7 @@ func (rc *ReconciliationContext) CheckRackPodTemplateDetails(force bool, failedR
 			desiredSts.DeepCopyInto(statefulSet)
 
 			rc.Recorder.Eventf(rc.Datacenter, corev1.EventTypeNormal, events.UpdatingRack,
-				"Updating rack %s", rackName, "force", force)
+				"Updating rack %s force=%t", rackName, force)
 
 			if err := rc.setConditionStatus(api.DatacenterUpdating, corev1.ConditionTrue); err != nil {
 				return result.Error(err)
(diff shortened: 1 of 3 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -704,18 +704,21 @@ func (rc *ReconciliationContext) CheckRackStoppedState() result.ReconcileResult
 }
 
 // checkSeedLabels loops over all racks and makes sure that the proper pods are labelled as seeds.
-func (rc *ReconciliationContext) checkSeedLabels() (int, error) {
+func (rc *ReconciliationContext) checkSeedLabels() (seedCount int, didSeedsChange bool, err error) {
```
burmanm: Lets not use any named return values, there's no real use case here. int, bool, err is fine.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -2139,6 +2145,13 @@ func (rc *ReconciliationContext) createStartSequence() []*corev1.Pod {
 		for podRankWithinRack := maxPodRankInThisRack; podRankWithinRack >= 0; podRankWithinRack-- {
 			podName := getStatefulSetPodNameForIdx(statefulSet, int32(podRankWithinRack))
 			pod := rc.getDCPodByName(podName)
+			if pod == nil {
```
burmanm, line 2156: issue: This could modify behavior later down the line. If the StatefulSet controller hasn't caught up and we don't have a pod, we might be applying changes to the StatefulSets that were not intended to be happening yet. Did you verify this situation is caught later in the sequence by the StS checks?

pkg/reconciliation/reconcile_racks_test.go
```diff
@@ -4155,3 +4155,139 @@ func TestCheckDcPodDisruptionBudget(t *testing.T) {
 	pdb = &policyv1.PodDisruptionBudget{}
 	require.NoError(rc.Client.Get(rc.Ctx, pdbName, pdb))
 }
+
+func TestRefreshSeeds(t *testing.T) {
+	assert := assert.New(t)
+	const (
```
burmanm: question: const inside a test function?

pkg/reconciliation/reconcile_racks.go
artsiom-bukhautsou: Agree fixed

pkg/reconciliation/reconcile_racks.go
artsiom-bukhautsou: No, I didn't check it. But after you pointed out I see that we'll catch this on step 5, because of this condition
```go
desiredSize <= readyPodCount
```

Anyways, I agree that  it would be better to catch it at the level of `startAllNodes` since its purpose is to ensure that all nodes are ready. I changed it to return an error to be safe.

pkg/reconciliation/reconcile_racks_test.go
artsiom-bukhautsou: it's more like personal preference to show that these variables shouldn't be changed. But probably overkill for the tests. Fixed.

burmanm (approved):

PR #903, opened 2026-03-11
Prevent any scaling operations from happening if trackedTask is running

### LINKED ISSUE #902: Cass-operator should block resize operations while cleanup following resize is in progress
Nodes were decommissioned as part of a resize (scale down) while cleanup (following scale up) was in progress.

This led to the cleanup task being unable to be successfully completed, because the nodes it was attempting to cleanup did not exist.

Where a cleanup task is in progress, cass-operator should not take actions to perform any scaling up or down.

burmanm:
**What this PR does**:
Adds additional blocks for operations that try to scale the cluster if a tracked task is running and it is expected to waited for. 

Fixes #902

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -937,10 +937,6 @@ func (rc *ReconciliationContext) CheckRackScale() result.ReconcileResult {
 		maxReplicas := *statefulSet.Spec.Replicas
 
 		if maxReplicas < desiredNodeCount {
-			if _, res := rc.waitForTrackedCleanupTask(); res.Completed() {
-				return res
-			}
-
 			// Check to see if we are resuming from stopped and update conditions appropriately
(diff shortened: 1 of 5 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -2756,6 +2771,10 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
 		return recResult.Output()
 	}
 
+	if recResult := rc.CheckCleanupAfterScaling(); recResult.Completed() {
```
burmanm, line 2774: This position won't work. It can't reliably prevent what we wanted to prevent, but it also shouldn't spawn the cleanup process here, since right now we would spawn it too early.

What we need is a webhook change preventing a change to scale down if ScalingUp is True in CassandraDatacenter.

burmanm:
This is replaced by the other PR adding webhook validations.

PR #909, opened 2026-03-20
Add support for maxUnavailable

### LINKED ISSUE #893: Support maxUnavailable for faster StatefulSet updates
### What is missing?

This behavior or request to a similar feature has been mentioned elsewhere, but here's a recap. In Kubernetes 1.35, the maxUnavailable is already in beta stage so it is also by default enabled in the Kubernetes. The alpha feature and the CRD has been available since 1.24. 

After introducing multiple concurrent properties, we should also allow changes of the StatefulSets to be applied in parallel fashion to improve the rollout speed. The alternative approach would be to use "OnDelete" and then deleting N-amount of pods after the StatefulSet was updated, but since there's a built-in feature in Kubernetes also, it makes sense to use that instead as it's hopefully stable. 

We should support two modes here, not just the ones defined by the StatefulSet:

* exact count of pods to update simultaneously
* percentage of pods

But also "entire rack" where we would calculate the correct amounts. That said, this feature should not cross rack boundaries so we would still maintain that only a single StatefulSet is updated before moving to another StatefulSet. In terms of code changes, this means that we would not touch the StatefulSet rollout code itself, only the field controlling.

### Why is this needed?

Larger changes take a long time to rollout. We already support fast parallel restarts, including from CassandraTasks as well as from Stopped state, but we do not support changes to the version for example.

burmanm:
**What this PR does**:
Adds support for maxUnavailable setting in the Spec. This allows to update multiple pods in the same time with new configuration (at a maximum of single rack at once).

I can't add e2e tests yet as we can't run against Kubernetes 1.35 reliably (and it's not enabled by default in older versions).

Fixes #893 

**Checklist**
- [ ] Changes manually tested


burmanm:
I still need to fix PDB

pkg/reconciliation/constructor.go
```diff
@@ -23,6 +23,25 @@ import (
...
+		rackNodeCounts := api.SplitRacks(int(dc.Spec.Size), len(racks))
+		maxRackNodeCount := 0
+		for _, rackNodeCount := range rackNodeCounts {
+			if rackNodeCount > maxRackNodeCount {
+				maxRackNodeCount = rackNodeCount
+			}
+		}
+
+		if maxUnavailable, err := intstr.GetScaledValueFromIntOrPercent(dc.Spec.MaxUnavailable, maxRackNodeCount, true); err == nil {
+			// Even if someone decides to set "200%" or something similarly silly here, it doesn't matter as we only process a single rack when it comes to MaxUnavailable changes
```
artsiom-bukhautsou: Question: Could you clarify this comment?

How does it cover the situation where a user specifies a very large number for `MaxUnavailable`? 

In my understanding, we might have an issue with the calculation below: 
```go
minAvailable = intstr.FromInt(calculatedMinAvailable)
```
if `minAvailable` is negative, we won't be able to set the value for the PDB, as this field must be >= 0.

Therefore, I was wondering if we could add a safety check, for example:
```go
if calculatedMinAvailable < 0 {
        calculatedMinAvailable = 0
}
```

burmanm: Yeah, it wouldn't matter in the context of processing, but you're right that PDB's api-server validation might not like it. I added a check that it can't be over 100%.

artsiom-bukhautsou (approved):

PR #916, opened 2026-04-13
Replacing mocks with fakeclient and httptest

burmanm:
**What this PR does**:
Removes mocks for Kubernetes Client and HttpClient. Replaces them with client-go's fakeclient and httptest. This allows to upgrade to go 1.26 since most of the mocks don't work with testify anymore (or forks of testify).

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


pkg/httphelper/client_test.go
```diff
@@ -461,70 +459,66 @@ func TestListRoles(t *testing.T) {
...
-	require.True(mockHttpClient.AssertExpectations(t))
 }
 
 func TestCallDurationMetricSuccess(t *testing.T) {
 	require := require.New(t)
 
-	mockHttpClient := mocks.NewHttpClient(t)
-	mockHttpClient.On("Do", mock.MatchedBy(func(req *http.Request) bool {
-		return req.URL.Path == "/api/v0/ops/node/drain" && req.Method == http.MethodPost
-	})).Return(newHttpResponseMarshalled("OK", http.StatusOK), nil).Times(2)
```
artsiom-bukhautsou, line 520: Question: I've noticed that in this case we're missing the number of calls expectation.

What do you think about adding it as an optional parameter (e.g., `opts ...AssertingHttpClientOption`)?

so it will look like this 

```go
httpClient := newAssertingHttpClient(
    t,
    func(req *http.Request) {
        require.Equal(t, http.MethodPost, req.Method)
        require.Equal(t, "/api/v0/ops/node/drain", req.URL.Path)
    },
    func() *http.Response {
        return newHttpResponseMarshalled("OK", http.StatusOK)
    },
    ExpectCalls(2),
)
```

artsiom-bukhautsou: or without modifying the HttpClient


```go
drainNodeCalls := 0

httpClient := newAssertingHttpClient(
    t,
    func(req *http.Request) {
        drainNodeCalls++
        require.Equal(t, http.MethodPost, req.Method)
        require.Equal(t, "/api/v0/ops/node/drain", req.URL.Path)
    },
    func() *http.Response {
        return newHttpResponseMarshalled("OK", http.StatusOK)
    },
)

.....

require.Equal(t, 2, drainNodeCalls, "expected drain endpoint to be called twice")
```

artsiom-bukhautsou (approved):

pkg/httphelper/client_test.go
burmanm: Ouh, I removed this one on purpose. The drain is actually called only once by the client, but we run the same drainCall twice, that's where the original .Times(2) comes from (otherwise it would fail to mock). 

So that didn't really test the client, it only tests the test which I don't think is useful.

artsiom-bukhautsou: I see, but in general it would be great to have this parameter for other tests

burmanm: There was no test that would verify it. Why would client.go make multiple calls as it has no logic? This test call is only used for client.go, not the others.

artsiom-bukhautsou: You're right. Thanks for clarifying

PR #921, opened 2026-04-23
Allow VolumeAttributesClassName changes in storageConfig

### LINKED ISSUE #922: Allow VolumeAttributesClassName changes in storageConfig webhook
## Problem

Changing `VolumeAttributesClassName` (VAC) in the `storageConfig.CassandraDataVolumeClaimSpec` of a `CassandraDatacenter` is rejected by the admission webhook `vcassandradatacenter.kb.io`:

```
admission webhook "vcassandradatacenter.kb.io" denied the request:
CassandraDatacenter write rejected, attempted to change
storageConfig.CassandraDataVolumeClaimSpec, diff:
  -  VolumeAttributesClassName: &"gp3-cassandra-general"
  +  VolumeAttributesClassName: &"gp3-cassandra"
```

## Background

`VolumeAttributesClassName` is a field on `PersistentVolumeClaimSpec` that references a `VolumeAttributesClass` object. It has been GA since kubernetes 1.34. On AWS EBS, this is the mechanism for in-place volume performance class changes (e.g. changing IOPS tier, Tags, etc.) without replacing the PVC.

The webhook currently blocks all changes to `CassandraDataVolumeClaimSpec` beyond storage size increases (gated by `cassandra.datastax.com/allow-storage-changes`). Since VAC changes are non-destructive and in-place — the PVC is not replaced, only its performance class is updated — they should be unconditionally permitted.

## Expected Behaviour

Changing `VolumeAttributesClassName` on an existing `CassandraDatacenter` should be allowed without any annotation.

## Impact

Operators running Cassandra on AWS EKS cannot dynamically change EBS volume performance tiers (e.g. switch between gp3 profiles) without bypassing webhook validation entirely via `cassandra.datastax.com/bypass-webhook-validations`.

SreedevT:
**What this PR does**:

VolumeAttributesClassName (VAC) is a field on `PersistentVolumeClaimSpec` used for in-place volume performance class changes (e.g. AWS EBS VolumeAttributesClass). It does not replace the PVC — only its performance tier is updated. The webhook was blocking all changes to CassandraDataVolumeClaimSpec beyond storage size, which prevented legitimate VAC updates.

This PR adds end-to-end VAC reconciliation with an annotation gate, CSI status monitoring, and STS recreation.
```mermaid
flowchart TD
    A["User changes VAC in CassandraDatacenter spec"] --> B{"Annotation<br/>allow-volume-attributes-changes<br/>= true?"}
    B -->|No| B1["❌ Error + InvalidDatacenterSpec event"]
    B -->|Yes| C["Patch all PVCs with new VAC"]
    C --> D["🔄 Requeue — wait for CSI driver"]
    D --> E{"CSI driver status?"}
    E -->|In-progress| F["⚠️ ModifyVolumeAttributesInProgress<br/>🔄 Requeue 10s"]
    F --> E
    E -->|Failed| G["❌ ReconcileFailed event<br/>Valid=False condition"]
    E -->|Done| H["✅ VolumeAttributesClassUpdated event"]
    H --> I["Recreate STS per rack<br/>(orphan cascade — pods survive)"]
```

As mentioned since the VAC as a feature is only GA in Kubernetes v1.34 and stable in the upcoming v1.36, I've kept it behind a feature gate `cassandra.datastax.com/allow-volume-attributes-changes`. This flow for this is heavily inspired by storage resize flow but there are slight differences.

Fixes #922

**Checklist**
- [x] Changes manually tested


internal/webhooks/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -204,12 +204,15 @@ func ValidateDatacenterFieldChanges(oldDc *api.CassandraDatacenter, newDc *api.C
...
-	// CassandraDataVolumeClaimSpec changes are disallowed
-	if metav1.HasAnnotation(newDc.ObjectMeta, api.AllowStorageChangesAnnotation) && newDc.Annotations[api.AllowStorageChangesAnnotation] == "true" {
-		// If the AllowStorageChangesAnnotation is set, we allow changes to the CassandraDataVolumeClaimSpec sizes, but not other fields
-		oldClaimSpec.Resources.Requests = newClaimSpec.Resources.Requests
+		if metav1.HasAnnotation(newDc.ObjectMeta, api.AllowStorageChangesAnnotation) && newDc.Annotations[api.AllowStorageChangesAnnotation] == "true" {
+			// If the AllowStorageChangesAnnotation is set, we allow changes to the CassandraDataVolumeClaimSpec sizes, but not other fields
+			oldClaimSpec.Resources.Requests = newClaimSpec.Resources.Requests
+		}
+
+		// VolumeAttributesClassName changes are always allowed as they represent in-place volume
```
burmanm: This isn't strictly true, the StorageClass must support ModifyVolume. Also, did you verify if the StatefulSet controller accepts these changes and applies them? 

There is also no monitoring in here that the VolumeAttributesClass change has been approved as the CSI driver can reject it. That would require changes like the resizing parameter checks and correct reconciliation until the change has been approved. 

This is relevant because in some drivers such as AWS, one can change from gp2 to gp3 and this operation might not be instant and as such will require the operator to correctly wait for the changes to be applied. There are also restrictions on how many times per day such operations can be done and we can't set the datacenter to a ready state if the changes haven't been applied (or for example incorrect / outdated parameters are used).

burmanm:
>  It also consolidates the two separate nil guards for CassandraDataVolumeClaimSpec into one block, fixing a latent nil panic if AllowStorageChangesAnnotation was set on a DC with no StorageConfig.

Where is this change and in which scenarios can it happen?

> VolumeAttributesClassName (VAC) is a Kubernetes 1.29+ field on PersistentVolumeClaimSpec used for in-place volume performance class changes (e.g. AWS EBS VolumeAttributesClass). 

They only GAed in 1.34 and marked as stable in upcoming 1.36.

internal/webhooks/cassandra/v1beta1/cassandradatacenter_webhook.go
SreedevT: Thats fair, I have made the PR a draft. Current PR only bypasses the validation and does not make any changes.
Can I make it a gated feature with more checks in place?

internal/webhooks/cassandra/v1beta1/cassandradatacenter_webhook.go
```diff
@@ -212,6 +212,11 @@ func ValidateDatacenterFieldChanges(oldDc *api.CassandraDatacenter, newDc *api.C
 		oldClaimSpec.Resources.Requests = newClaimSpec.Resources.Requests
 	}
 
+	// VolumeAttributesClassName changes are always allowed. Annotation check during reconciliation will determine if the change should be applied to existing PVCs or not.
```
SreedevT, line 215: This is different from storage resize, where annotation check happens in the webhook itself. 
I removed this because the annotation check is being done on reconcile and the warning thrown there is more informative. But it can be added here if needed

SreedevT:
Was able to test the changes on EKS.

SreedevT:
@burmanm 
Would it make sense to unify PVC modifications (resize and VAC changes) into a single reconcile step? With the current code, if a user submits both a storage size increase and a VAC change at the same time, they're applied as separate PVC patches in separate reconcile passes — resize first, then VAC. On some providers this risks hitting the per-volume modification cooldown window since they count as two separate ModifyVolume calls.

If both changes were applied as a single combined PVC patch, they'd count as one modification. This would require refactoring CheckVolumeClaimSizes and ReconcileVolumeAttributesClass into a unified reconcile step that computes the full desired PVC spec diff and applies it atomically.

PR #931, opened 2026-05-27
fix(services): set resource-hash on the NodePort service so it reconciles

### LINKED ISSUE #930: NodePort Service is never updated (missing resource-hash annotation), so additionalServiceConfig changes are ignored
### Bug

cass-operator creates the NodePort Service for a `CassandraDatacenter` but never updates it afterwards. Changes to `spec.additionalServiceConfig.nodePortService` (annotations, labels) — and ports — are silently ignored on the already-existing NodePort Service.

### Root cause

`newNodePortServiceForCassandraDatacenter` in `pkg/reconciliation/construct_service.go` is the only service builder that does **not** call `utils.AddHashAnnotation(service)` before returning (the CQL, seed, all-pods and additional-seed builders all do).

So the desired NodePort Service is built without the `cassandra.datastax.com/resource-hash` annotation, and the Service created in the cluster has no such annotation either. In `CheckHeadlessServices` (`pkg/reconciliation/reconcile_services.go`) an existing service is only updated when:

```go
if !utils.ResourcesHaveSameHash(currentService, desiredSvc) { /* update */ }
```

`ResourcesHaveSameHash` compares the `cassandra.datastax.com/resource-hash` annotation of both objects. Since neither the current nor the freshly built desired NodePort Service has it, the comparison is `"" == ""` → `true` → the update branch is skipped. The NodePort Service is therefore only ever created, never updated.

### Impact

After the NodePort Service exists, editing `additionalServiceConfig.nodePortService.additionalAnnotations` (e.g. `consul.hashicorp.com/service-name` for Consul catalog sync) or labels has no effect — the live Service keeps its original values. Tell-tale sign: the NodePort Service has no `cassandra.datastax.com/resource-hash` annotation, while the other DC services do.

### Reproduce

1. Create a `CassandraDatacenter` with `networking.nodePort` enabled and some `additionalServiceConfig.nodePortService.additionalAnnotations`.
2. Change one of those annotations.
3. The `*-node-port-service` keeps the old annotation value; the operator logs no service update.

### Fix

Add `utils.AddHashAnnotation(service)` in `newNodePortServiceForCassandraDatacenter` like the other builders, so the resource-hash is set and drift is detected. PR follows.

LipatovAlexander:
**What this PR does**:

`newNodePortServiceForCassandraDatacenter` was the only service builder that did not call `utils.AddHashAnnotation(service)`. Because of that the NodePort service was created without the `cassandra.datastax.com/resource-hash` annotation, and `CheckHeadlessServices` only updates an existing service when `ResourcesHaveSameHash(current, desired)` reports a difference. With no hash on either side the comparison was always `"" == ""` → `true`, so the NodePort service was created once and never updated — changes to `additionalServiceConfig.nodePortService` (annotations, labels) and ports were silently ignored.

This adds `utils.AddHashAnnotation(service)` to the NodePort builder, exactly like the other service builders, so the resource-hash is set and the operator reconciles updates to the NodePort service. A regression test (`TestNewNodePortServiceForCassandraDatacenterHasResourceHash`) asserts the annotation is present.

Fixes #930

**Checklist**
- [x] Changes manually tested


burmanm:
Hey, thank you for the PR. Given that the new test doesn't actually verify anything that reconciliation now works, could you just move that assert to the TestLabelsWithNewNodePortServiceForCassandraDatacenter for example? That at least verifies the annotation stays there (even if the test is called TestLabels). 

And the changelog could just indicate that we added the missing resource-hash annotation to detect changes. Right now the PR looks a bit too Claude slop with lots of added stuff that's not really giving anything useful.

LipatovAlexander:
Thanks for the review! Done:
- Removed the standalone `TestNewNodePortServiceForCassandraDatacenter...` test and moved the resource-hash assertion into `TestLabelsWithNewNodePortServiceForCassandraDatacenter`, so it verifies the annotation stays present.
- Trimmed the CHANGELOG entry to a single line noting the missing resource-hash annotation was added so changes are detected.

burmanm:
Thanks.

PR #937, opened 2026-06-16
Re-assert datacenter progress/condition metrics on every reconcile

### LINKED ISSUE #936: Datacenter progress/condition metrics disappear after operator restart or leader change
## Bug

`cass_operator_datacenter_progress` and `cass_operator_datacenter_status` are only written when the state changes:

- `setOperatorProgressStatus` returns early when the progress state is unchanged, before touching the metric.
- `setCondition` updates the metric only inside its `if updated` block.

Both gauges live in the process-local Prometheus registry. So when the operator pod restarts (upgrade, eviction, OOM) or leader election moves reconciliation to another replica, a CassandraDatacenter that is already in a steady state never transitions again and these series are never re-emitted — they simply disappear until the next real transition.

`cass_operator_datacenter_pods_status` is not affected because `UpdatePodStatusMetric` re-asserts it on every reconcile.

## Impact

Dashboards and alerts built on `*_datacenter_progress` / `*_datacenter_status` show gaps after any operator restart. With leader election enabled and more than one replica it is worse: only the active leader ever holds the series, and they are lost on every failover.

## Steps to reproduce

1. Deploy a CassandraDatacenter, let it reach `Ready`.
2. Restart the cass-operator pod (or trigger a leader change).
3. While the datacenter sits idle, observe that `cass_operator_datacenter_progress` and `cass_operator_datacenter_status` are absent, while `cass_operator_datacenter_pods_status` is present.

## Proposed fix

Re-assert the progress and condition gauges from the persisted `dc.Status` once per reconcile, the same way pod statuses are already handled. PR to follow.

LipatovAlexander:
**What this PR does**:

`cass_operator_datacenter_progress` and `cass_operator_datacenter_status` are only written on a state transition (`setOperatorProgressStatus` returns early when unchanged; `setCondition` touches the metric only inside `if updated`) and live in the process-local registry. After an operator restart or a leader-election change, a datacenter already in a steady state never transitions again, so those series disappear until the next change. `cass_operator_datacenter_pods_status` is unaffected because it is re-asserted every reconcile.

This re-asserts the progress and condition gauges from the persisted `dc.Status` once per reconcile, mirroring how pod statuses are handled. Purely additive; no extra Kubernetes API calls.

Fixes #936

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 ## unreleased
 
+* [BUGFIX] [#936](https://github.com/k8ssandra/cass-operator/issues/936) Re-assert datacenter progress and condition metrics from status on every reconcile so they are not lost on operator restart or leader change.
 * [BUGFIX] [#933](https://github.com/k8ssandra/cass-operator/issues/933) Fix regression introduced in the CassandraTask replacement process introduced in the full rack replacement feature.
 
 ## v1.31.0
--- a/pkg/monitoring/metrics.go
+++ b/pkg/monitoring/metrics.go
@@ -124,6 +124,18 @@ func UpdateOperatorDatacenterProgressStatusMetric(dc *api.CassandraDatacenter, s
 	DatacenterOperatorStatusVec.WithLabelValues(dc.Namespace, dc.Spec.ClusterName, dc.DatacenterName(), string(state)).Set(1)
 }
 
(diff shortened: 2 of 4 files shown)


burmanm:
Hey, thank you for the PR, but I ended up slightly rewriting the approach with some changes to the exported functions as well as doing the refresh on the Reconcile() method instead of handler.go as that could return incorrect status to the metrics.

This should ensure that the end state of reconciliation is always visible at the metrics. If there's no changes, it will be a quick operation that refreshes the metrics even if they were lost due to a restart of the operator.

PR #940, opened 2026-06-24
Add support-bundle creation to integration tests

burmanm:
**What this PR does**:
* Simplifies the code to share the TestLifecycle for all the tests instead of repeating the same code everywhere.
* Adds a support-bundle creation for troubleshooting when the tests fail

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


tests/util/ginkgo/lib.go
```diff
@@ -21,13 +22,152 @@ import (
...
+		"--interactive=false",
+		"--output",
+		outputPath,
+	)
+	if err != nil {
+		t.Logf("Failed to collect support bundle: %v", err)
+		return
+	}
+
+	fmt.Printf("\n\tSupport bundle dumped at: %s\n\n", outputPath)
```
artsiom-bukhautsou, line 119: question: Why it was decided to use `fmt.Printf` instead of `t.Logf` here? 
The same [here](https://github.com/burmanm/cass-operator/blob/e0e9f244187962c5375f00b93bd9d37dffbdfa73/tests/util/ginkgo/lib.go)

artsiom-bukhautsou (approved):

tests/util/ginkgo/lib.go
burmanm: It was just the way it used to be. There are a lot of reasons not to use t.Logf() though, since it's basically fmt.Sprintf to stderr instead of stdout and these are not errors. Also, this log is never output if the test doesn't fail. Although we don't go to this path without failure..

PR #749, opened 2025-01-30
If PodDeletionTimestamp != nil, return Pod state as being Terminating

burmanm:
**What this PR does**:
Modify the metrics state of the pods to return Terminating if DeletionTimestamp is not nil and we are not decommissioning it.

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [ ] CHANGELOG.md updated (not required for documentation PRs)
- [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

PR #944, opened 2026-07-01
Refresh metrics on every reconcile, add terminating state

### LINKED ISSUE #936: Datacenter progress/condition metrics disappear after operator restart or leader change

### LINKED ISSUE #943: Pod being deleted should have terminating status in metrics
### What is missing?

We do not present a pod with deletionTimestamp != nil in the metrics at this point as separate one as we only take status from the Pod status. However, kubectl shows this as Terminating and that's the terminology that users know.

### Why is this needed?

Some pods that have been deleted might be in incorrect state in the metrics as we don't show that the pod has actually been deleted.

burmanm:
**What this PR does**:
This merges and reworks some of the stuff from PRs #937 and #749

All the metrics are refreshed on every reconciliation (excluding cooldown) to avoid missing them if operator has restarted or changed leadership.

Adds new state for Terminating if Pod's DeletionTimestamp != nil. Otherwise the Pod would show in a different state and this is inconsistent with what the kubectl output is.

Fixes #943 
Fixes #936

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/internal/controllers/cassandra/cassandradatacenter_controller.go
+++ b/internal/controllers/cassandra/cassandradatacenter_controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/k8ssandra/cass-operator/pkg/dynamicwatch"
 	"github.com/k8ssandra/cass-operator/pkg/images"
+	"github.com/k8ssandra/cass-operator/pkg/monitoring"
 	"github.com/k8ssandra/cass-operator/pkg/oplabels"
 	"github.com/k8ssandra/cass-operator/pkg/reconciliation"
 	appsv1 "k8s.io/api/apps/v1"
@@ -162,6 +163,9 @@ func (r *CassandraDatacenterReconciler) Reconcile(ctx context.Context, request c
 	if res.RequeueAfter > 0 && res.RequeueAfter < minimumRequeueTime {
 		res.RequeueAfter = minimumRequeueTime
 	}
+
+	monitoring.RefreshDatacenterMetrics(rc.Datacenter)
+
 	return res, err
 }
 
--- a/pkg/monitoring/metrics_test.go
+++ b/pkg/monitoring/metrics_test.go
@@ -73,6 +73,20 @@ func TestMetricAdder(t *testing.T) {
 	require.NoError(err)
 	require.Equal("ready", status)
 
+	now := metav1.Now()
+	pods[1].SetDeletionTimestamp(&now)
+	UpdatePodStatusMetric(pods[1])
+	status, err = getCurrentPodStatus("pod1")
+	require.NoError(err)
+	require.Equal("terminating", status)
+
+	// Decommissioning should be prefered to Terminating if we are decommissioning the pod
+	pods[5].SetDeletionTimestamp(&now)
+	UpdatePodStatusMetric(pods[1])
+	status, err = getCurrentPodStatus("pod5")
+	require.NoError(err)
+	require.Equal("decommissioning", status)
+
 	RemoveDatacenterPods("ns", "cluster1", "datacenter1")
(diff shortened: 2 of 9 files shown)


pkg/monitoring/metrics_test.go
```diff
@@ -73,6 +73,20 @@ func TestMetricAdder(t *testing.T) {
...
+	now := metav1.Now()
+	pods[1].SetDeletionTimestamp(&now)
+	UpdatePodStatusMetric(pods[1])
+	status, err = getCurrentPodStatus("pod1")
+	require.NoError(err)
+	require.Equal("terminating", status)
+
+	// Decommissioning should be preferred to Terminating if we are decommissioning the pod
+	pods[5].SetDeletionTimestamp(&now)
+	UpdatePodStatusMetric(pods[1])
```
rzvoncek: question: is this really supposed to be pod 1 ?

internal/controllers/cassandra/cassandradatacenter_controller.go
```diff
@@ -162,6 +163,9 @@ func (r *CassandraDatacenterReconciler) Reconcile(ctx context.Context, request c
 	if res.RequeueAfter > 0 && res.RequeueAfter < minimumRequeueTime {
 		res.RequeueAfter = minimumRequeueTime
 	}
+
+	monitoring.RefreshDatacenterMetrics(rc.Datacenter)
```
rzvoncek, line 167: nit: in the pr description you mention this is happening intentinally after the checks for the cooldown. a comment mentioning this is intentionally might be useful.

but now I'm also interested, why don't we do this refresh before the cooldown checks?

rzvoncek (changes requested):
Requesting changes only because I think there's a typo in the assertion (see the comment).

internal/controllers/cassandra/cassandradatacenter_controller.go
burmanm: Because we don't have the correct information on our Status copy (and might even have stale data there) as we skip the entire reconciliation process - we only create the context. 

So there's no point running it since we don't the current values.

rzvoncek: k, thanks

pkg/monitoring/metrics_test.go
burmanm: No, but it doesn't actually make any difference here. I'll fix it anyway and add a description why is it like that.

rzvoncek (approved):

Issue #947, opened 2026-07-06
Operator OOMs in cluster-scoped mode: informer cache is unbounded and scales with cluster size


LipatovAlexander:
## What happened

We run cass-operator (under k8ssandra-operator) on mid-size production clusters. With a namespace list in `WATCH_NAMESPACE` the operator is healthy at ~120Mi. When we switched to cluster-scoped mode (empty `WATCH_NAMESPACE`), **both operator deployments were OOM-killed within seconds of startup** (512Mi limit, exit 137, CrashLoopBackOff before the readiness probe ever passed), on every startup.

Cluster where this reproduces 100%: 286 namespaces, 6554 pods, 2845 configmaps, 2561 services, 2533 endpointslices, 27k+ events, thousands of secrets. We also verified the same on clusters up to **17.5k pods**.

## Root cause

The manager cache is configured only through `cache.Options.DefaultNamespaces` (`cmd/main.go`). With an empty namespace list the map stays empty, which controller-runtime interprets as "cache every object of every watched GVK across all namespaces". There is no `ByObject` scoping, no label/field selectors, and no transform.

The concrete memory drivers:

1. **Events**: `setupCacheIndexers` registers a field index on `corev1.Event`, which eagerly instantiates a **cluster-wide Event informer** at startup. Events are the highest-cardinality, highest-churn resource (27k+ objects in our cluster). The index has exactly one consumer: `failureModeDetection` reads the events of a single pod with a namespaced field-selected List.
2. **Pods/StatefulSets/PDBs/Services**: fully cached cluster-wide, although the operator only ever reads objects it created (which all carry `app.kubernetes.io/managed-by=cass-operator` — the same invariant the existing `managedByCassandraOperatorPredicate` already asserts on events).
3. **Secrets**: fully cached cluster-wide (two `Watches(&corev1.Secret{})`), including every unrelated tenant secret in the cluster. Besides memory this is also a security-posture concern: a heap/core dump of the operator exposes every secret in the cluster.
4. **Lazily-created informers**: any cached read instantiates a cluster-wide informer for that type. This silently created informers for PersistentVolumeClaims, StorageClasses, Endpoints and EndpointSlices.

Note that `builder.WithPredicates(...)` does not help: predicates filter which events trigger reconciles, but run after the informer has already listed, deep-copied and stored the objects.

[...]

PR #948, opened 2026-07-06
Read Events through an uncached APIReader instead of a cluster-wide informer

LipatovAlexander:
**What this PR does**:
Reads Events through an uncached `mgr.GetAPIReader()` instead of the cached client, and drops the `corev1.Event` field index.

The index existed to serve a single lookup (`failureModeDetection` reads the events of one pod), but registering it eagerly instantiated a **cluster-wide Event informer** at startup — Events being the highest-cardinality, highest-churn resource (27k+ objects on a mid-size cluster). In cluster-scoped deployments this informer is a major OOM driver. The field selector `involvedObject.name` is server-side supported for Events, so the live read is cheap and behavior-identical.

`APIReader` is threaded through `CassandraDatacenterReconciler` and `ReconciliationContext` so future high-cardinality reads can use the same path.

Part of #947 (first of two PRs; the second bounds the rest of the cache)

**Checklist**
- [x] Changes manually tested — running in production on five clusters for several days, cluster-scoped, incl. chaos-mesh pod-kill runs
- [x] Automated Tests added/updated — TestFailureDetection exercises the APIReader path; envtest controller suite passes
- [x] CHANGELOG.md updated
- [x] CLA Signed

### DIFF
--- a/internal/controllers/cassandra/cassandradatacenter_controller.go
+++ b/internal/controllers/cassandra/cassandradatacenter_controller.go
@@ -73,9 +73,12 @@ var (
 // CassandraDatacenterReconciler reconciles a cassandraDatacenter object
 type CassandraDatacenterReconciler struct {
 	client.Client
-	Log      logr.Logger
-	Scheme   *runtime.Scheme
-	Recorder events.EventRecorder
+	// APIReader is an uncached reader (mgr.GetAPIReader) used for high-cardinality
+	// reads deliberately kept out of the informer cache, such as Events.
+	APIReader client.Reader
+	Log       logr.Logger
+	Scheme    *runtime.Scheme
+	Recorder  events.EventRecorder
 
 	// SecretWatches is used in the controller when setting up the watches and
 	// during reconciliation where we update the mappings for the watches.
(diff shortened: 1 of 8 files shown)


internal/controllers/cassandra/cassandradatacenter_controller.go
```diff
@@ -73,9 +73,12 @@ var (
 // CassandraDatacenterReconciler reconciles a cassandraDatacenter object
 type CassandraDatacenterReconciler struct {
 	client.Client
-	Log      logr.Logger
-	Scheme   *runtime.Scheme
-	Recorder events.EventRecorder
+	// APIReader is an uncached reader (mgr.GetAPIReader) used for high-cardinality
```
burmanm, line 76: suggestion: Please remove these types of comments, they don't serve any usage. We don't document every type, user should know what they mean from their definition or use their editor to check it. It's in the end the reason we have typed language.

pkg/reconciliation/reconcile_racks.go
```diff
@@ -182,7 +182,12 @@ func (rc *ReconciliationContext) failureModeDetection() (bool, string) {
 					// Pod has been over 5 minutes in Pending state. This can be normal, but lets see
 					// if we have some detected failures events like FailedScheduling
 					events := &corev1.EventList{}
-					if err := rc.Client.List(rc.Ctx, events, &client.ListOptions{Namespace: pod.Namespace, FieldSelector: fields.SelectorFromSet(fields.Set{"involvedObject.name": pod.Name})}); err != nil {
+					// Read Events through the uncached APIReader: only this one Pod's
```
burmanm, line 185: This comment is also better described in the commit, not as comment on the source code. One can see the reasoning using git blame in that case.

burmanm (approved):

PR #949, opened 2026-07-06
Bound the informer cache for cluster-scoped deployments

### LINKED ISSUE #947: Operator OOMs in cluster-scoped mode: informer cache is unbounded and scales with cluster size

LipatovAlexander:
**What this PR does**:
Bounds the manager's informer cache so operator memory no longer scales with cluster size (details and production numbers in #947):

- `cache.Options.ByObject` label-scopes the Pod/StatefulSet/PodDisruptionBudget/Service informers to `app.kubernetes.io/managed-by=cass-operator`. The operator only reads objects it created and labels; the existing `managedByCassandraOperatorPredicate` on `Owns()` already asserts the same invariant on events. Hot per-reconcile reads stay cached.
- Secrets are read live (`client.CacheOptions.DisableFor`): the operator also reads user-provided secrets (SuperuserSecretName, `Spec.Users`, ConfigSecret, management-API TLS) that carry no operator label, so they cannot be label-scoped. Both Secret watches become metadata-only projections (`builder.OnlyMetadata`); their map fn / predicate only inspect annotations. `configSecretMapFn` drops its `*corev1.Secret` type assertion accordingly.
- PVC/StorageClass/Endpoints/EndpointSlice/ConfigMap reads are cold paths (resize, decommission, additional-seeds, image config) and also go live, so no cluster-wide informers are lazily created for them.
- All cached objects are stripped of managedFields and the kubectl last-applied-configuration annotation.

Production results (cluster-scoped): OOMKilled at startup with 512Mi → **87-91Mi** steady state on clusters from 5.7k to 17.5k pods. Namespace-scoped deployments also drop ~7x.

Independent of the Events PR; both compose.

Fixes #947

**Checklist**
- [x] Changes manually tested — production on five clusters, cluster-scoped, incl. chaos-mesh pod-kill runs against managed Cassandra clusters
- [x] Automated Tests added/updated — envtest controller suite exercises the metadata-only watches and rewritten map fn
- [x] CHANGELOG.md updated
- [x] CLA Signed

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -29,8 +29,14 @@ import (
 	"go.uber.org/zap/zapcore"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -53,6 +59,7 @@ import (
 	controlcontrollers "github.com/k8ssandra/cass-operator/internal/controllers/control"
 	apiwebhook "github.com/k8ssandra/cass-operator/internal/webhooks/cassandra/v1beta1"
 	"github.com/k8ssandra/cass-operator/pkg/images"
+	"github.com/k8ssandra/cass-operator/pkg/oplabels"
 	"github.com/k8ssandra/cass-operator/pkg/utils"
 )
 
@@ -185,8 +192,46 @@ func main() {
 		setupLog.Info("webhooks disabled because --webhook-cert-path was not provided")
 	}
 
+	// Bound the informer cache so cluster-scoped mode (empty WATCH_NAMESPACE) does not
+	// cache every object of every watched type across all namespaces:
+	//
+	//   * Pod/StatefulSet/PodDisruptionBudget/Service informers are label-scoped to
+	//     managed-by=cass-operator: the operator only ever reads objects it created,
+	//     and it labels all of them. These are the hot read paths (every reconcile),
+	//     so they stay cached.
+	//   * Secrets are read through live API calls (Client.Cache.DisableFor) because
+	//     the operator also reads user-provided secrets (SuperuserSecretName,
+	//     Spec.Users, ConfigSecret, management API TLS) that carry no operator label;
+	//     the two Secret watches use metadata-only projection (builder.OnlyMetadata),
+	//     so no full-object Secret informer ever exists. PVC/StorageClass/
+	//     EndpointSlice/Endpoints/ConfigMap reads are cold paths and also go live,
+	//     avoiding lazily-created cluster-wide informers for those types.
+	//   * All cached objects are stripped of managedFields and the kubectl
+	//     last-applied-configuration annotation, which often dominate object size.
+	managedByOperator := labels.SelectorFromSet(labels.Set{
+		oplabels.ManagedByLabel: oplabels.ManagedByLabelValue,
+	})
 	options.Cache = cache.Options{
 		DefaultNamespaces: map[string]cache.Config{},
+		DefaultTransform:  stripHeavyMetadata(),
+		ByObject: map[client.Object]cache.ByObject{
+			&corev1.Pod{}:                   {Label: managedByOperator},
+			&appsv1.StatefulSet{}:           {Label: managedByOperator},
+			&policyv1.PodDisruptionBudget{}: {Label: managedByOperator},
+			&corev1.Service{}:               {Label: managedByOperator},
+		},
+	}
+	options.Client = client.Options{
+		Cache: &client.CacheOptions{
+			DisableFor: []client.Object{
+				&corev1.Secret{},
+				&corev1.PersistentVolumeClaim{},
+				&corev1.ConfigMap{},
+				&corev1.Endpoints{},
+				&discoveryv1.EndpointSlice{},
+				&storagev1.StorageClass{},
+			},
+		},
 	}
 
 	clusterScoped := len(ns) == 0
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -343,6 +388,30 @@ func setupWebhookServer(webhookCertPath, webhookCertName, webhookCertKey string,
 	}), webhookCertWatcher, true, nil
 }
 
+// stripHeavyMetadata extends cache.TransformStripManagedFields by also dropping the
```
burmanm, line 391: issue: I don't see much value in this comment since it describes what the code does. The only "added" value is that kubectl applies this annotation, but there lies the problem also. This transformer is used in the default caching options that take care of "Pod/StatefulSet/PodDisruptionBudget/Service", yet none of these actually have anything that's managed by the kubectl.

It doesn't look like there's any added value compared to using `DefaultTransform: cache.TransformStripManagedFields()` since items generated by the cass-operator will not have the LastAppliedConfigAnnotation.

cmd/main.go
```diff
@@ -185,8 +192,46 @@ func main() {
...
+		ByObject: map[client.Object]cache.ByObject{
+			&corev1.Pod{}:                   {Label: managedByOperator},
+			&appsv1.StatefulSet{}:           {Label: managedByOperator},
+			&policyv1.PodDisruptionBudget{}: {Label: managedByOperator},
+			&corev1.Service{}:               {Label: managedByOperator},
+		},
+	}
+	options.Client = client.Options{
+		Cache: &client.CacheOptions{
+			DisableFor: []client.Object{
```
burmanm, line 226: issue: This is problematic. On every reconcile (which can happen a lot, especially in clusters with a lot of nodes and CassandraDatacenters) we poll Secrets (such as superUserSecret). PersistentVolumeClaims are always checked that they refer to the correct size of the racks, so every reconcile operation as well. Endpoints/EndpointSlices are always verified if deployed as part of the k8ssandra-operator (as there's additionalSeeds used in that case). Endpoints is only listed for the deprecated check though.

StorageClass/ConfigMap are safe, StorageClass is only checked in the case of PVC expansion so it doesn't matter much and ConfigMaps are not processed by the controller-runtime reconciling at all (only in the main.go). 

But the other three are trickier and would cause a lot of reads to the API server on every reconcile on large clusters or with large amount of datacenters. And that slows down the processing in the reconcile queue as well. We probably need to up the MaxConcurrentReconciles to some other value than the current default, but that'll need additional testing infra I haven't finished yet. 

I will need to verify if there's available labels in those object types to allow filtered caching. I'm not sure if StatefulSet copies any labels automatically to PVCs. SuperuserSecret can be generated by us or be external, so that might work from a dynamicWatcher with the caching - but this needs to be verified. EndpointSlices do have enough labels, the same managedByOperator. Endpoints I'll need to verify from older versions.

internal/controllers/cassandra/cassandradatacenter_controller.go
```diff
@@ -191,13 +191,14 @@ func (r *CassandraDatacenterReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Owns(&policyv1.PodDisruptionBudget{}, builder.WithPredicates(managedByCassandraOperatorPredicate)).
 		Owns(&corev1.Service{}, builder.WithPredicates(managedByCassandraOperatorPredicate))
 
+	// Uses only object metadata (no *corev1.Secret assertion) so it works with the
```
burmanm, line 194: issue: This comment adds nothing that's not explained by the code.

burmanm (changes requested):
Sorry for the delay. There's a couple of problematic issues with removing caches of so many items which will need addressing as well some tidying up of comments. The overall approach is correct, but needs a bit of tuning for few use cases. The MaxConcurrent issue is not something that should be fixed in this PR, it's just to explain part of the issue.

burmanm (approved):

burmanm:
Thanks for the PR, I made some changes to it before merging (to improve secret caching since they keep getting called every reconcile for mTLS & superuser).

PR #953, opened 2026-07-20
Add events patch role

### LINKED ISSUE #954: Leader election requires events patch rights
### What happened?

Leader election requires events patch rights to function correctly as it may upgrade previous events. This right was pruned when we removed unused Role rights, but was not caught because it's from a runtime dependency.

_No response_

Start.

### cass-operator version

1.32

### Kubernetes version

1.36

### Method of installation

_No response_

_No response_

burmanm:
**What this PR does**:
Adds patch right to the events API since LeaderElection uses it

Fixes #954 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/utils/crypto.go
+++ b/pkg/utils/crypto.go
@@ -38,7 +38,7 @@ func setupKey() (*big.Int, time.Time, *rsa.PrivateKey, string, time.Time, error)
 	return nil, time.Time{}, nil, "", time.Time{}, err
 }
 
-func GetNewCAandKey(leafdomain, namespace string) (keypem, certpem string, err error) {
+func GetNewCAandKey(leafdomain, namespace string) (string, string, error) {
 	serialNumber, notBefore, priv, privPem, notAfter, err := setupKey()
 	if err == nil {
 		buffer := bytes.NewBufferString("")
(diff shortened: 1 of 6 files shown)


pkg/utils/crypto.go
```diff
@@ -68,7 +68,7 @@ func GetNewCAandKey(leafdomain, namespace string) (keypem, certpem string, err e
 	return "", "", err
 }
 
-func prepare_ca(ca *corev1.Secret) (ca_cert_bytes []byte, ca_certificate *x509.Certificate, ca_key *rsa.PrivateKey, err error) {
+func prepare_ca(ca *corev1.Secret) (ca_cert_bytes []byte, ca_certificate *x509.Certificate, ca_key *rsa.PrivateKey, err error) { //nolint:nonamedreturns
```
rzvoncek, line 71: question: why do we need to name the returns here?

rzvoncek (approved):

pkg/utils/crypto.go
burmanm: I don't think we do, but this is part of the code that's been here for ages and something I don't plan to modify at all anymore (since we would want to use cert-manager in reality). So I thought it was the safest to keep it unchanged.

rzvoncek: ack, thanks

PR #958, opened 2026-08-07
Allow to configure mgmt-api-http port

### LINKED ISSUE #242: Allow management-api in different port
**What is missing?**
Currently management-api is considered to be always in the port 8080 by the cass-operator. We should allow PodHost to define a port as well to simplify testing and potentially allow deployments where the port is different.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/CASS-52) by [Unito](https://www.unito.io)
┆Issue Number: CASS-52

### LINKED ISSUE #950: Readiness/Liveness probes configuration ignored when mTLS enabled
### What is missing?

# Summary

The current probe implementation in `cass-operator should follow the existing [Kubernetes Probe Concepts](https://kubernetes.io/docs/concepts/workloads/pods/probes).

In certain cases, the `cass-operator` will break-user configured probes.

# Details

When an mTLS connection is enabled the probe handler needs to be converted to `Exec` handler. This is required to ensure the mutual authentication with the Management API; that is, both client and server must present valid certificates. Mechanisms to use client certificate, client key, and CA are unsupported by other handlers. This pattern is common in applications requiring mTLS for health endpoints.

The current probes implementation has the following behaviour when mTLS is enabled.

* **Ignores user-configured HTTPGet probes:** Ignores existing `httpGet` probe configuration which is stated in a TODO code comment.
* **Overwrites user-configured TCPSocket probes:** If the `tcpSocket` handler is configured, it is set to nil, overwriting the configuration.
* **Overwrites user-configured Exec probes:** Sets `exec` unconditionally, overwriting any existing configuration.
* **Creates invalid multi-mechanism probes with GRPC:** If the `grpc` handler is configured, it is never set to nil, resulting in a probe with both `grpc` and `exec` set. This state violates the Kubernetes specification.
* **Ignores handler configuration:** Overwrites timing, thresholds, and other settings.
* **Uses hardcoded values:** Derives the Management API endpoint from the hardcoded value of `localhost:8080`.
 
The following table summarises the current behaviour.

| Probe Handler | Without mTLS | Implementation Notes |
|---------------|----------------|-----------------------|
| `httpGet` | Works | Destroyed, replaced with Exec | HTTP configurations lost |
| `tcpSocket` | Works | Destroyed, set to nil | Socket configurations lost |
| `exec` | Works | Overwritten | Exec configuration lost |
| `grpc` | Works | Invalid state | This is a bug; both `grpc` + `exec` are set |
| No probe (nil) | Works | Probe created | Behaviour changed |

### Why is this needed?

The `cass-operator` should to take the following into account when mTLS is enabled:

* **Single Mechanism:** Only one of the four probe mechanisms should be set.
* **Configuration Perseveration:** The provided `HTTPGet` probe configuration should be preserved and translated an `Exec` probe.
* **Nil vs Empty:** nil probe means no probe; kubelet considers Success, empty struct means probe with defaults.
* **Non-HTTPGet Probes:** GRPC, TCPSocket, and Exec probes must be supported and preserved.

burmanm:
**What this PR does**:
Extracts the information from containerPort in the cassandra container to check if the port has been overridden. Applies the correct adjustments to probes/envVars/preStop.

Fixes #242 
Fixes #950 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/kindIntegTest.yml
+++ b/.github/workflows/kindIntegTest.yml
@@ -180,6 +180,7 @@ jobs:
           - additional_serviceoptions
           - scheduled_task
           - additional_volumes
+          - mgmt_api_port_change
           # - delete_node_terminated_container # This does not test any operator behavior
           # - podspec_simple
           # - terminate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 
 * [ENHANCEMENT] [#943](https://github.com/k8ssandra/cass-operator/issues/943) Add terminating state as one of the Pod status metrics
 * [ENHANCEMENT] [#936](https://github.com/k8ssandra/cass-operator/issues/936) Every reconcile of the CassandraDatacenter should refresh the metrics for up-to-date state
+* [ENHANCEMENT] [#242](https://github.com/k8ssandra/cass-operator/issues/242) Allow configuring the mgmt-api port. Setting cassandra container's mgmt-api-http containerPort will override the default 8080 value.
 * [BUGFIX] [#933](https://github.com/k8ssandra/cass-operator/issues/933) Fix regression introduced in the CassandraTask replacement process introduced in the full rack replacement feature.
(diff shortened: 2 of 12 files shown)


rzvoncek (approved):

ossarga (approved):
Hi @burmanm,

Thank you for raising this pull request to fix issue #950 

Changes look good to me.

PR #965, opened 2026-08-21
Add three new flags to configure concurrency, reconcile timeouts and pprof

### LINKED ISSUE #964: Add ability to configure reconcile concurrency and timeout.
### What is missing?

We don't support configuring the controller-runtime concurrency to allow reconciling multiple CassandraDatacenters simultaneously. At the same time, we should add a timeout to prevent a slow reconcile from blocking the eventloop.

### Why is this needed?

While we shouldn't block the reconciling eventloop, it can sometimes happen with slow responding pods or network issues. There should be a way to increase concurrency in these cases to allow other DCs to continue reconciling.

If concurrency is increased, we should also allow setting the profiling to find out if there's any production gaps (disable by default).

burmanm:
**What this PR does**:
Allow modifying controller-runtime settings around concurrency, timeouts and pprof.

Previously the reconcile could take infinite time, this will limit it to 2 minutes.

Also fix CHANGELOG with missing versions

Fixes #964 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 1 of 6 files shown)


cmd/main.go
```diff
@@ -174,9 +185,13 @@ func main() {
 	options := ctrl.Options{
 		Scheme:                 scheme,
 		Metrics:                metricsServerOptions,
+		PprofBindAddress:       pprofAddr,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "b569adb7.cassandra.datastax.com",
+		Controller: config.Controller{
+			ReconciliationTimeout: reconciliationTimeout,
```
artsiom-bukhautsou, line 193: question: it seems like by specifying timeout here it will affect both cassandra dc reconciliation and cassandra task, as well as other controllers. Is it intentional?

burmanm: Yes, none of our tasks themselves should take that long to execute a single reconcile. Tasks should push to a background processing.

artsiom-bukhautsou (approved):

PR #970, opened 2026-09-07
Fix db service ports

### LINKED ISSUE #959: Add database service ports 9160 and 9142 based on the database configuration
### What is missing?

Currently, database services kubernetes resources created by cass-operator contain ports 9142 and 9160. 
For example:
```
[test:0] ~$ k get svc
NAME                                            TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)                                                 AGE
test-tls-test-tls-dc1-service                   ClusterIP   None             <none>        9042/TCP,9142/TCP,8080/TCP,9103/TCP,9000/TCP,9160/TCP   4h16m
```

However, the usage of these ports is dependant on the configuration of the database. In many cases they are not used, but cass-operator adds them to the service anyway. The additional ports defined in the service create false positives and overhead for security scans.

**9142: CQL TLS native transport**
- Per: https://docs.datastax.com/en/cassandra-oss/3.x/cassandra/configuration/configCassandra_yaml.html#configCassandra_yaml__native_transport_port_ssl
- Port 9142 is not used unless native_transport_port_ssl is set in cassandra.yaml
- The service should not include this port by default, and if native_transport_port_ssl is set, the specific defined port should be added to the service

**9160: Thrift port**
- Per https://github.com/aiven/cassandra/blob/cassandra-3.11/NEWS.txt#L806
- Starting in 3.11 start_rpc is defaulting to false so 9160 would be closed
- There is already some logic for this:
```
	if strings.HasPrefix(dc.Spec.ServerVersion, "3.") || dc.Spec.ServerType == "dse" {
		ports = append(ports,
			namedServicePort("thrift", 9160, 9160))
	}
```
- But it adds port 9160 for all DSE versions. 
- The service should not include 9160 by default and if start_rpc is true cassandra.yaml, then add it, or the specific defined port in rpc_port setting

### Why is this needed?

To match the actual ports used by the database to the service defined by cass-operator and reduce the surface area for security scanning

artsiom-bukhautsou:
**What this PR does**:

Fixes #959 

**Checklist**
- [X] Changes manually tested
- [X] Automated Tests added/updated
- [X] CHANGELOG.md updated (not required for documentation PRs)


apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -780,6 +787,72 @@ func namedPort(name string, port int) corev1.ContainerPort {
 	return corev1.ContainerPort{Name: name, ContainerPort: int32(port)}
 }
 
+func GetPortsFromCassCfg(cfg json.RawMessage) (CassandraConfigPorts, error) {
```
burmanm: issue: Sadly, no. We don't want to allow changing ports in the cassandra-yaml section (and that ticket wasn't about it). We already defined earlier that port overrides will happen by using containerPort in the Kubernetes definition.

So what this can only do is fetch if an extra port is wanted, such as rpc_start set to true, not rpc_port. Otherwise we run into a situation that there's two different methods of changing ports that we externally provide. 

Other thing, cassandra-yaml is not necessarily set, there's also a Secret method where the configuration is not visible in the CassandraDatacenter, but only referenced through a secret.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -780,6 +787,72 @@ func namedPort(name string, port int) corev1.ContainerPort {
...
+	cfgPorts.NativeTransportPortSSL = nativeTransportPortSSL
+	thriftPort, err := buildThriftPort(casYamlMap)
+	if err != nil {
+		return CassandraConfigPorts{}, err
+	}
+	cfgPorts.ThriftPort = thriftPort
+	return cfgPorts, nil
+}
+
+func buildThriftPort(casYamlMap map[string]any) (*int, error) {
```
burmanm: Also, this still stands:

if strings.HasPrefix(dc.Spec.ServerVersion, "3.") || dc.Spec.ServerType == "dse" {

Otherwise, exit immediately.

CHANGELOG.md
```diff
@@ -10,8 +10,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 ```
 
 ## unreleased
-
-* [CHANGE] [#969](https://github.com/k8ssandra/cass-operator/issues/969) Pull cass-config-builder from GHCR
+* [BUGFIX] [#959](https://github.com/k8ssandra/cass-operator/issues/959) Configure rpc_port/native_transport_port_ssl service ports from cassandra.yaml
```
burmanm: And this isn't a bugfix, there was nothing bugged - user just wanted different behavior. We can call it enhancement rather.

apis/cassandra/v1beta1/cassandradatacenter_types.go
artsiom-bukhautsou: > We don't want to allow changing ports in the cassandra-yaml section (and that ticket wasn't about it)

But in the acceptance criteria of the ticket it's mentioned: 
> - Port 9142 is not used unless native_transport_port_ssl is set in cassandra.yaml
> - The service should not include this port by default, and if native_transport_port_ssl is set, the specific defined port should be added to the service.

Which means we should change the port if user specified it.

artsiom-bukhautsou: > - We already defined earlier that port overrides will happen by using containerPort in the Kubernetes definition.
> - Other thing, cassandra-yaml is not necessarily set, there's also a Secret method where the configuration is not visible in the CassandraDatacenter, but only referenced through a secret.


I haven't taken into account the cassandra-yaml in the secret and the container port overrides in the `podTemplateSpec`. This adds complexity to the solution.

While we have container port overrides using the podTemplateSpec, they don't affect the service ports.

Therefore, I think we first need to get the ports from the cassandra-yaml and the cassandra-yaml secret and merge them, giving the ports from the secret higher priority if there is a conflict. Then, we should override the ports using the `podTemplateSpec`.

However, in this situation, we have an issue: the cassandra-yaml and the cassandra-yaml secret change both container and service ports, while the override from the podTemplateSpec applies only to container ports.

artsiom-bukhautsou: > So what this can only do is fetch if an extra port is wanted, such as rpc_start set to true, not rpc_port. Otherwise we run into a situation that there's two different methods of changing ports that we externally provide.

I see that makes sense. But won't it be confusing for the user? They change the cassandra-yaml configuration by specifying value for `rpc_port` and `native_transport_port_ssl` and it doesn't affect `CassandraDatacenter` CR

apis/cassandra/v1beta1/cassandradatacenter_types.go
artsiom-bukhautsou: I've removed it, because the new behaviour was requested in the issue:
> - But it adds port 9160 for all DSE versions. 
> - The service should not include 9160 by default and if start_rpc is true cassandra.yaml, then add it, or the specific defined port in rpc_port setting

CHANGELOG.md
artsiom-bukhautsou: Sure. I'll fix it.

I've made it as a bug, because 9142 and 9160 ports are enabled by default, but they aren't reachable, which leads to unexpected behaviour.

Here the description from the ticket:
> The additional ports defined in the service create false positives and overhead for security scans.

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: cassandra-yaml is not intended to modify the CassandraDatacenter, it's not related. cassandra-yaml is never the end result of what we deploy, it goes through some processing in the config-builder also. There can be fields that are not really in Cassandra at all and there can be fields that we will remove even if user sets them. For example, networking is one of those, we will override anything one sets in the rpc_address, broadcast_rpc_address, cluster_name etc for all versions (for DSE we override even more). 

But none of that should have any effect on the deployment, it's only Cassandra configuration. It's not a "use this to override any setting in the deployment". Same way you can't change interfaces, ingress or any other setting there. Thrift port and native SSL ports are a weird oddity in this case, they're only used by DSE really at this point anymore. 

We just haven't had a structure to disable/set certain ports. But if you set for example graph to be enabled, it will modify in DSE's case the ports, despite what you set in the config section. 

The thing is, there's always 2 port selections. One is external port, one is internal to the Pod. External configuration has nothing to do with Cassandra's config.

But yes, containerPorts should have an effect on the servicePorts.

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: Yes, but only DSE and 3.11 have thrift service at all (or that setting).

CHANGELOG.md
burmanm: Not entirely true (even if the ticket says so, but the ticket is wrong since it does not understand how 3.11 is deployed):

https://github.com/datastax/cass-config-definitions/blob/9898218b4dcc3fbf0e429941b03252cd4211b50e/resources/cassandra-yaml/cassandra/cassandra-yaml-cassandra-3.11.6.edn#L239

apis/cassandra/v1beta1/cassandradatacenter_types.go
artsiom-bukhautsou: > cassandra-yaml is not intended to modify the CassandraDatacenter, it's not related. cassandra-yaml is never the end result of what we deploy, it goes through some processing in the config-builder also. There can be fields that are not really in Cassandra at all and there can be fields that we will remove even if user sets them. For example, networking is one of those, we will override anything one sets in the rpc_address, broadcast_rpc_address, cluster_name etc for all versions (for DSE we override even more).
> 
> But none of that should have any effect on the deployment, it's only Cassandra configuration. It's not a "use this to override any setting in the deployment". Same way you can't change interfaces, ingress or any other setting there. Thrift port and native SSL ports are a weird oddity in this case, they're only used by DSE really at this point anymore.

Does this mean that we shouldn't care at all about which ports are specified in cassandra.yaml?


> We just haven't had a structure to disable/set certain ports. 
> But if you set for example graph to be enabled, it will modify in DSE's case the ports, 
> despite what you set in the config section.

Okay, then instead of relying on `cassandra-yaml`, we should add new fields to the `CassandraDatacenter` spec that allow us to set `native_transport_port_ssl`, `rpc_start`, and `rpc_port`. These fields would affect both the container ports and the service ports. Is that correct?

apis/cassandra/v1beta1/cassandradatacenter_types.go
artsiom-bukhautsou: Got it 👍

apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: We already have containerPort with names, why not use them? But we should fix our Service creation to read that data.

artsiom-bukhautsou: Sure, we can. Then I'll do the following:
- Remove the existing `namedPort("tls-native", 9142)` from the container ports and service ports. If the client needs it, they can set it using `podTemplateSpec`.
- Update the `containerPorts` override logic to update the service ports as well.

burmanm: In the construct_service.go, could we go a bit further?

func newAllPodsServiceForCassandraDatacenter(dc *api.CassandraDatacenter) *corev1.Service {

In that function, instead of setting those names, pick the required ports from list + fetching from StatefulSet podTemplateSpec (IIRC it's already written at this point). Also, Prometheus metrics endpoint (9103) should probably be disabled in most cases since our defaults do not use it. In that case, if env variable MGMT_API_DISABLE_MCAC is set to true, disable it. This happens with readOnlyFs for example, so almost all deployments at this point.

There's multiple places with hardcoded ports, so it might be a small task to find all of them..

artsiom-bukhautsou: Sure

artsiom-bukhautsou:
~~working on tests~~
✅

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -738,6 +738,29 @@ func (dc *CassandraDatacenter) GetSuperuserSecretNamespacedName() types.Namespac
...
+					if env.Name == "MGMT_API_DISABLE_MCAC" {
+						return env.Value != "true"
+					}
+				}
+			}
+		}
+	}
+
+	// MCAC is enabled by default
+	return true
```
burmanm, line 761: issue: This isn't true, MCAC is only enabled for Cassandra 4.1 or older (and DSE?). Cassandra 5.0.0 and newer (including HCD 2.0.x) do not have MCAC installed at all.

apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -784,20 +807,22 @@ func namedPort(name string, port int) corev1.ContainerPort {
...
+		nativePort = dc.GetNodePortNativePort()
+	}
 	// Note: Port Names cannot be more than 15 characters
 
 	ports := []corev1.ContainerPort{
 		namedPort("native", nativePort),
-		namedPort("tls-native", 9142),
 		namedPort("internode", internodePort),
-		namedPort("tls-internode", 7001),
-		namedPort("jmx", 7199),
```
burmanm, line 795: This could in theory be enabled in k8ssandra-operator still. We might need to change the k8ssandra-operator to add containerPort in that case.

pkg/reconciliation/construct_service.go
```diff
@@ -333,3 +289,35 @@ func makeGenericHeadlessService(dc *api.CassandraDatacenter) *corev1.Service {
 
 	return &service
 }
+
+func cassandraContainerPorts(dc *api.CassandraDatacenter) []corev1.ContainerPort {
```
burmanm, line 293: suggestion: This is actually port overrides / additional ones only, not the defaults (PodTemplateSpec is not mandatory). Might want to reflect that in the naming of the function perhaps?

pkg/reconciliation/construct_service.go
```diff
@@ -333,3 +289,35 @@ func makeGenericHeadlessService(dc *api.CassandraDatacenter) *corev1.Service {
...
+	portDefaults, err := dc.GetContainerPorts()
+	if err != nil {
+		return nil, err
+	}
+
+	userPorts := cassandraContainerPorts(dc)
+	combinedPorts := combinePortSlices(portDefaults, userPorts)
+
+	var servicePorts []corev1.ServicePort
+	for _, cp := range combinedPorts {
```
burmanm, line 315: issue: Kubernetes also allows one to add ports without names and that would cause a bug here. We should filter non-named ports here since they're not correctly configured (for us).

pkg/reconciliation/construct_service_test.go
```diff
@@ -60,8 +60,11 @@ func TestCassandraDatacenter_allPodsServiceLabels(t *testing.T) {
 		api.PromMetricsLabel:    "true",
 	}
 
-	service := newAllPodsServiceForCassandraDatacenter(dc)
+	service, err := newAllPodsServiceForCassandraDatacenter(dc)
 
+	if err != nil {
```
burmanm, line 65: nit: require.NoError(err)

Since we use require everywhere in the unit tests, even if the old test didn't.

pkg/reconciliation/reconcile_services.go
```diff
@@ -59,9 +59,17 @@ func (rc *ReconciliationContext) CheckHeadlessServices() result.ReconcileResult
...
 
-	cqlService := newServiceForCassandraDatacenter(dc)
+	cqlService, err := newServiceForCassandraDatacenter(dc)
+	if err != nil {
+		logger.Error(err, "Could not create headless service")
+		return result.Error(err)
+	}
 	seedService := newSeedServiceForCassandraDatacenter(dc)
-	allPodsService := newAllPodsServiceForCassandraDatacenter(dc)
+	allPodsService, err := newAllPodsServiceForCassandraDatacenter(dc)
```
burmanm, line 68: question: Are these terminalErrors (requeue won't fix it) or something that could be temporary? We could reflect that in the error type.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Changelog for Cass Operator, new PRs should update the `main / unreleased` secti
 * [CHANGE] [#969](https://github.com/k8ssandra/cass-operator/issues/969) Pull cass-config-builder from ghcr.io instead of docker.io where its no longer updated
 * [BUGFIX] [#971](https://github.com/k8ssandra/cass-operator/issues/971) Decommissioning last pod of a datacenter left its pod_status metric abandoned instead of being removed. And also we got an error logline which wasn't a real error
 * [BUGFIX] [#973](https://github.com/k8ssandra/cass-operator/issues/973) Prevent premature decommission cleanup when Cassandra metadata requests fail.
+* [ENHANCEMENT] [#959](https://github.com/k8ssandra/cass-operator/issues/959) Synchronize Cassandra Services with configured container ports and remove ports that are unavailable or intended only for pod-level communication
```
burmanm, line 17: issue: Enhancements go before bugfixes.

apis/cassandra/v1beta1/cassandradatacenter_types.go
artsiom-bukhautsou: Sure, PR for the k8ssandra-operator has been created: https://github.com/k8ssandra/k8ssandra-operator/pull/1797

apis/cassandra/v1beta1/cassandradatacenter_types.go
artsiom-bukhautsou: Thanks, fixed

pkg/reconciliation/construct_service.go
artsiom-bukhautsou: updated the function name to be more explicit

pkg/reconciliation/construct_service.go
artsiom-bukhautsou: good point, added an additional check

pkg/reconciliation/construct_service_test.go
artsiom-bukhautsou: fixed

pkg/reconciliation/reconcile_services.go
artsiom-bukhautsou: It appeared that these errors were redundant. Forgot to delete them after the refactoring

CHANGELOG.md
artsiom-bukhautsou: fixed

burmanm (approved):

PR #974, opened 2026-09-09
Prevent premature decommission cleanup when metadata requests fail

### LINKED ISSUE #973: Failed metadata requests can trigger premature decommission cleanup
### What happened?

During scale-down, failed Cassandra metadata requests can cause a node to be treated as finished decommissioning. The operator then reduces the StatefulSet's replica count and deletes the node's PVC without confirming that it has left the ring.

The [metadata lookup](https://github.com/k8ssandra/cass-operator/blob/031404df9b5732b2d952bab0fa6cc9f2bdd9d83e/pkg/reconciliation/reconcile_racks.go#L1398-L1416) ignores request errors and can return an empty endpoint list. [IsDoneDecommissioning](https://github.com/k8ssandra/cass-operator/blob/031404df9b5732b2d952bab0fa6cc9f2bdd9d83e/pkg/reconciliation/decommission_node.go#L262-L282) treats a node missing from that list as done, so an empty list allows cleanup.

This bug was found using **[Specula](https://github.com/specula-org)**, a tool I’m working on with the team.

Keep the node and its PVC until usable metadata confirms decommission has finished. Reconciliation should retry when metadata is unavailable.

Reproduction: [test_bugMC-2_premature_decommission_test.go](https://github.com/specula-org/specula-case-studies/blob/61c09e3178b25c26238fb061a724d643833cdfe0/systems/cass-operator/modules/core/runs/cass-operator-20260813/repro/test_bugMC-2_premature_decommission_test.go).

On a clean checkout of cass-operator's `master` branch, place that file in `pkg/reconciliation/` and run:

```sh
go test ./pkg/reconciliation/ -run TestMC2 -count=1 -v
```

These tests pass when the original bug is reproduced: a metadata HTTP 500 leads to PVC deletion, while the healthy-metadata control retains the PVC. They use the Kubernetes fake client and a local fake Management API server.

The bug was also reproduced on `master` with a regression test that exercises `ReconcileAllRacks()`:

1. Set up a two-replica StatefulSet with both Cassandra pods ready and their PVCs present. Set the datacenter's desired size to one, its ScalingDown condition to True, and the last pod's node-state label to Decommissioning.
2. Return HTTP 500 for `/api/v0/metadata/endpoints` on both pods.
3. Run `ReconcileAllRacks()`.

Reconciliation returns no error, reduces replicas from two to one, and deletes the decommissioning node's PVC.

### cass-operator version

master

### Kubernetes version

Not applicable to this reproduction. It uses the Kubernetes fake client without a running Kubernetes API server.

munimthahmid:
**What this PR does**:

When Cassandra metadata requests failed, the error was ignored, and reconciliation continued with an empty endpoint list. During decommission, that could make the node look absent from the ring, causing the operator to scale down its StatefulSet and delete its PVC before decommission finished.

This change returns the metadata error and stops reconciliation before starting or completing a decommission until usable metadata is available. Valid metadata where the node is LEFT or absent keeps the existing completion behavior.

Fixes #973

**Testing**

- `go test ./pkg/reconciliation -count=1` passed.
- Focused decommission tests passed with the race detector. The regression fails on the original code and passes with the fix.
- `make docker-build docker-logger-build` passed.
- Local Kind `scale_down` and `decommission_dc` passed with Cassandra 5.0.6.

**Checklist**

- [ ] Changes manually tested
- [ ] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

### DIFF
--- a/pkg/reconciliation/reconcile_racks.go
+++ b/pkg/reconciliation/reconcile_racks.go
@@ -1395,24 +1395,34 @@ func isNodeStuckAfterLosingReadiness(pod *corev1.Pod) bool {
 	return hasBeenXMinutesSinceReady(10, pod)
 }
 
-func (rc *ReconciliationContext) getCassMetadataEndpoints() httphelper.CassMetadataEndpoints {
-	var metadata httphelper.CassMetadataEndpoints
+func (rc *ReconciliationContext) getCassMetadataEndpoints() (httphelper.CassMetadataEndpoints, error) {
+	var lastErr error
 	for _, pod := range rc.clusterPods {
-		// Try to query the first ready pod we find.
-		// We won't get any endpoints back if no pods are ready yet.
+		// A cluster may have no ready pods while starting, so callers decide
+		// whether unavailable metadata should stop their reconciliation step.
 		if !isServerReady(pod) {
 			continue
 		}
 
-		metadata, _ = rc.NodeMgmtClient.CallMetadataEndpointsEndpoint(pod)
-
-		if len(metadata.Entity) == 0 {
+		metadata, err := rc.NodeMgmtClient.CallMetadataEndpointsEndpoint(pod)
+		if err != nil {
+			lastErr = fmt.Errorf("failed to get Cassandra metadata endpoints from pod %s: %w", pod.Name, err)
 			continue
 		}
-		break
+
+		if len(metadata.Entity) == 0 {
+			// A ready Cassandra node should report at least its own endpoint.
+			continue
+		}
+
+		return metadata, nil
 	}
 
-	return metadata
+	if lastErr != nil {
+		return httphelper.CassMetadataEndpoints{}, lastErr
+	}
+
+	return httphelper.CassMetadataEndpoints{}, fmt.Errorf("no usable Cassandra metadata endpoints available from ready pods")
 }
 
 // When a vmware taint occurs and we delete the pvc, pv, and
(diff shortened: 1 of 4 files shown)


pkg/reconciliation/reconcile_racks.go
```diff
@@ -2699,7 +2709,7 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
 	rc.clusterPods = pods
 	rc.dcPods = rc.datacenterPods()
 
-	endpointData := rc.getCassMetadataEndpoints()
+	endpointData, endpointErr := rc.getCassMetadataEndpoints()
```
burmanm: Hey, thanks for the report, but I don't think most of the code you modified is relevant. Is there a reason we couldn't just do 

```go
if endpointData, err := rc.getCassMetadataEndpoints(); err != nil {
  return result.Error(err)
}
```

And that would cause a retry? There seems to be no reason to drag this error to decommission.

munimthahmid: Thank you so much for reviewing this! You’re right that returning the error there would cause a retry, so there’s no need to pass it into the decommission functions. I’ve removed the extra error parameters.

I tried the early return locally. With `spec.stopped` set to `true` and no active decommission, metadata request failures caused reconciliation to return before `CheckRackStoppedState()`. The replicas stayed at two instead of going to zero.

So I’ve updated the PR to keep the existing helper and check for empty metadata only before starting or completing decommission. These checks still return an error and trigger a retry, while allowing the stop request in that test to proceed. The regression tests and the Kind `scale_down` and `decommission_dc` tests passed with this change.

Can you have a look?

burmanm: Yes, I forgot to mention the stopped state process. I was thinking more in line that if none of the pods are in readyState and stopped is set to true (not necessarily in this order though) we would not throw an error in that case. 

But your solution works to fix the underlying issue and it's not complex, so this is fine. I can later revisit the error handling on that metadata part, there are potentially some edge cases that the testing would not catch in which case any error catching could issues. I remember similar from the seeds handling where Kubernetes API server somehow managed to make a Pod stuck indefinitely in the Terminating state and we ended up looping forever to handle that issue. #887 

This solution of yours is safer in that sense.

burmanm (approved):

